PR #55, opened 2015-01-11
Add B9 proc wings.  Rearrange FASA launch apparatus

jwvanderbeck:
* Adds the new B9 Procedural Wings into START node like existing pWings.
* Moves FASA launch apparatus to START with the exception of the full fledged tower.  This allows use of clamps and umbilicals early on to avoid boil off issues.
* Sorted Stock Launch Clamp along with FASA clamps

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -117,6 +117,9 @@ start:
     winglet:
     winglet3:
     wingStrake:
+    B9_Aero_Wing_Procedural_TypeA:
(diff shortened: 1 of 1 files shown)


jwvanderbeck:
I did *not* test these because I didn't want to go through the hassle of setting up perl to compile it, but its pretty basic changes that should be sound.

Issue #57, opened 2015-01-13
Hiding/tagging unsupported parts


NathanKell:
When parts don't say "non-RO", people tend to think that means RP-0 supports them. We should at the very least tag unsupported parts, IMO, and consider hiding them.

OtherBarry:
I'm unsure how i feel about this. On the one hand, RP-0 is a very custom career ode with a very clear intention, hence, hiding unsupported parts prevents people from ruining RP-0, without losing the ability to play sandbox with whatever parts they want.

On the other hand, you have pjf's view, which is that RP-0 is very light weight, and as compatible and supportive of whatever users choose to do with it as possible, so hiding things is a bad idea. Tagging them may not be great either, as when I play, using a "non-RO" part makes me feel guilty, as if i've broken the rules of the realism mods. Which again, doesn't really fit with the (not that i like to call it this, but it is apt) pjf RP-0 way.

NathanKell:
The problem is RL10 available from node 1 for like 600 funds...

Sent by my thumbs, slowly.

OtherBarry:
Which RL10? From which part pack?

NathanKell:
KW, IIRC. Also it appears BobCat's engines unlock early too.

You're correct that hiding is bad for sandbox. Instead, we can just chuck them in the ORPHANS node, it's what it's there for. Since the problem seems worst for engines, we might only do it for engines, so that other parts we don't touch don't necessarily get made unavailable in career.

BevoLJ:
Would this affect the non-ro parts we've added.  I add Kos and IR.  I know they aren't supported, but feel they greatly enhance my style of play and like where they fall in the tech tree right out of the box so use them anyway.

NathanKell:
Create a mini-PR? :)

Felger:
Suggestion:
Create a RO Filter config for Filter Extensions ([Forum Link](http://forum.kerbalspaceprogram.com/threads/104231), [Github Link](https://github.com/Crzyrndm/FilterExtension/))

bax-:
Both of Felgers suggestions sound great to me (who is otherwise totally not involved in this).

cherrydev:
[...]

PR #61, opened 2015-01-22
Hide all non RO parts

bax-:
exception: Asteroids
maybe better optional?


NathanKell:
Um...this only deals with RO parts, not whether parts are supported or not by RP-0 itself. Also, per the discussion in #57 this might be too drastic. Thanks though!

bax-:
Totally understand!

PR #62, opened 2015-01-24
AIES Integration

jwvanderbeck:
Work in Progress
Added all AIES parts to ORPHANS until they are configured
Added LR-91 and LR-87 engines and derivatives


jwvanderbeck:
Added the satellite buses available in AIES.  Dates aren't 100% real because the real dates were driven more by commercial demand than technical ability.  Instead I placed them between the two main satellite nodes where they made the most sense.

However I have priced them lower than the existing FASA Agena probe core based on my estimation of prices given that the R&D program for NASA's common bus was about $10.7 million in 1965 dollars.  I think the FASA Agena probe core should be looked at to ensure its pricing is correct.

jwvanderbeck:
Some changes are dependent on pull #63

jwvanderbeck:
Ok so that covers all the AIES engines up to the end of the 1960s. Without any real tech clarity beyond that i'm holding off on the other engines.

jwvanderbeck:
Fixed merge conflict because git is stupid

PR #67, opened 2015-01-30
Update AIES engine VR1vulcan (RS-68A)

Arbos:
Remove engine from ORPHANS and add to giganticRocketry.

When I updated RP-0 to 0.21 I noticed RS-68A wasn't available. I have added it to giganticRocketry (TL7 present day) because it  was certified for use on Delta IV in December 2001.

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -888,13 +888,13 @@ fusionRockets:
 giganticRocketry:
 #TL7 present day
     liquidEngineconstelacion:
-    VR1vulcan:
 
 colossalRocketry:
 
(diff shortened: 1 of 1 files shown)


jwvanderbeck:
The tech tree out beyond 1965-1970 is really quite vague, so I had pushed this engine and the other more modern engines out until I had a chance to try and make sense of the future plans.

NathanKell:
This should be in exoticAlloys actually; we are using the Construction line as the hydrolox engines line.

PR #69, opened 2015-01-30
FASA engine configs

jwvanderbeck:
Initial FASA engine configs file
Added LR87 and LR91 to match their AIES counterparts


NathanKell:
I believe this has been superceded per our IRC discussion, @jwvanderbeck. Please let me know if it was not and reopen.

PR #70, opened 2015-01-31
add smaller sizes for Stack Chute and Cone Chute

bax-:
- they are now the biggest size  -> you have to press "previous size" to get them
- might also be integrated in RealismOverhaul but is probably RP-0 specific


NathanKell:
Yeah, I'd actually much rather you apply this directly to RO. I'm gradually trying to backport all the extra stuff RP-0 adds back into RO, such that RP-0 will *only* deal with tech and prices.

bax-:
sounds reasonable!

PR #72, opened 2015-02-03
Fix cost of Mark1-2Pod

BevoLJ:
1 Million to 10k


NathanKell:
@BevoLJ but you didn't change cost, you changed research cost (entrycost) which is supposed to be a billion bucks (1,000,000 funds). Cost remains 10,000. Questionmark.

PR #74, opened 2015-02-06
Add AJ10-104D

jwvanderbeck:
Yes I named my branch wrong.

This adds the AJ10-104D, the main engine used in Able Star launches.
ref: http://space.skyrocket.de/doc_lau/thor_ablestar.htm

Also a quick fix to the Isp of the -142

### DIFF
--- a/GameData/RP-0/VSR_Engines.cfg
+++ b/GameData/RP-0/VSR_Engines.cfg
@@ -931,6 +931,31 @@
 		}
 		CONFIG
 		{
+			name = AJ10-104D	// Main engine used on Ablestar
+			minThrust = 35.1
+			maxThrust = 35.1
+			heatProduction = 100
+			PROPELLANT
+			{
+				name = UDMH
+				ratio = 0.406
+				DrawGauge = True
+			}
+			PROPELLANT
+			{
+				name = IWFNA
+				ratio = 0.594
+			}
+			atmosphereCurve
+			{
+				key = 0 278
+				key = 1 248
+			}
+			massMult = 1.0
+			techRequired = advRocketry
+		}
+		CONFIG
+		{
 			name = AJ10-142-Delta
 			minThrust = 34.25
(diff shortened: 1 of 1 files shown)


NathanKell:
Needs a slightly higher massMult (1.1 or so?) for the larger nozzle, and it also seems to lack a cost? Other than that, very cool! :)

jwvanderbeck:
Do you have reference on the larger nozzle?  The mass is a bit up in the air because from what I can tell it was a very flexible engine that could be configured in many different ways based on mission requirements.  This included both an aluminum version that massed 30% less as well as a stainless steel version.

NathanKell:
I don't recall my reference offhand, but I did read the area ratio was higher (after all, gotta get that higher Isp somehow) from a nozzle extension. That extension is what adds the mass.

jwvanderbeck:
Can bump it up a little bit I guess, but my understanding is that it weighs LESS than its predecessors when used without the extension, so i'm not sure how much mass to add with it.
It goes from an expansion ratio of 20 without to 40 with if that helps determine extension size?

jwvanderbeck:
Actually this engine is supposed to have restart capability as well, but i'm not sure how to add that.

NathanKell:
@jwvanderbeck slight problem: if you add that engine ignitor module config to one of the CONFIGs you need to add them to all CONFIGs else when you swap to 104D you get the five restarts but when you swap away it doesn't change back to only 1 ignition (as with all things in CONFIG, if it's not set explicitly the existing value is maintained). See how the LV-909 config (also in VSR) handles it.

As for the 5-15 thing, I think you're thinking of Agena and the Gemini Agena Target Vehicle.

jwvanderbeck:
Arg well that's a pain.  Ok i'll look into that in a bit.  And yeah I couldn't find my source so quite possibly I was thinking of the wrong engine.

PR #75, opened 2015-02-07
Engin configs cleanup

jwvanderbeck:
** This will break save games **

Just like on the RO side, this is a pass to clean up all the engine configs to bring them into a neat orderly fashion with consistency across part packs, etc, etc.

**NOTE** At least one change in here requires the same change made to RO which is in this PR: https://github.com/KSP-RO/RealismOverhaul/pull/137

So the RO needs to be accepted before this one will work properly.

### DIFF
--- a/GameData/RP-0/FASA_Engines.cfg
+++ b/GameData/RP-0/FASA_Engines.cfg
@@ -97,3 +97,19 @@
 		}
 	}
 }
+// A4/A7
+@PART[FASA_Mercury_Redstone_Eng]:FOR[RP-0]
+{
+	@MODULE[ModuleEngineConfigs]
+	{
+		%configuration = A-6
+		@CONFIG[A-6]
+		{
+			%cost = 0
+		}
+		@CONFIG[A-7]
+		{
+			%cost = 50
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


pjf:
Good news: RealismOverhaul gives the antenna sensible stats. 0.05 EC/sec, and a range of 6Mm. While
that still makes it a not-so-attractive choice of antennae (4 x C16s in root mode costs the same, weighs
less, and gives more range), it's nowhere near as imbalanced as what I thought.

Bad news: Either RO isn't applying the change properly, or RemoteTech is overwriting it:

`RO_FASA_Agena.cfg`:

    @PART[FASAAgenaAnt]:AFTER[RemoteTech]
    {
            @MODULE[ModuleRTAntenna]
            {
                    @Mode1OmniRange = 6000000
                    EnergyCost = 0.05
                    @TRANSMITTER
                    {
                            @PacketInterval = 0.4
                            @PacketSize = 1
                    }
            }
    }

`RemoteTech_FASA_Antennas.cfg`:

    @PART[FASAAgenaAnt]:AFTER[FASA]:NEEDS[RemoteTech] // Agena Antenna
    {
        !MODULE[ModuleDataTransmitter] {}
        
        @MODULE[ModuleAnimateGeneric]
        {
            %allowManualControl = false
        }
        
        %MODULE[ModuleRTAntenna] {
            %Mode0OmniRange = 0
            %Mode1OmniRange = 3750000
            %MaxQ = 6000
            %EnergyCost = 0.36
            
            %DeployFxModules = 0
            
            %TRANSMITTER
            {
                %PacketInterval = 0.10
                %PacketSize = 2
                %PacketResourceCost = 8.0
            }
        }
        
        %MODULE[ModuleSPUPassive] {}
    }

I'm not exactly sure why RO's changes aren't taking. It may be we're running at the wrong time, especially since the RT config isn't `:FOR[RemoteTech`.

PR #84, opened 2015-02-16
New AJ10 fixes

jwvanderbeck:
Fixed name of shroud node


jwvanderbeck:
This seems to fix the AJ10 for me.  2 test flights with no issues.

PR #86, opened 2015-02-17
Cost/Place Skylab & fix Soyuz solar node

BevoLJ:
Solars from Electronics to Electrics nodes

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -844,6 +844,10 @@ flightControl:
 
 #now this is Hydrolox TL3
 advConstruction:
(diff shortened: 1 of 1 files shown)


BevoLJ:
Waiting on https://github.com/KSP-RO/RealismOverhaul/pull/146 for Skylab to be added to RO.

BevoLJ:
Looks like @Felger just merged the Skylab PR in RO, so if my costing is good enough it is ready to go.  :)

PR #89, opened 2015-02-26
Place even more parts

pjf:
A bunch of part placings and costings so things look better for anyone
with FASA or SXT installed.


NathanKell:
<3 commentary.

PR #92, opened 2015-03-01
FASA Apollo

BevoLJ:
One small step for a Bevo, one giant leap for RP-0

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -824,6 +824,10 @@ generalConstruction:
     ok_tft:
         entryCost: 25000
         cost: 800
+    # Apollo parts
+    FASAApollo_SM_Engine:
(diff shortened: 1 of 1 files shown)


BevoLJ:
https://docs.google.com/spreadsheets/d/1NxbJqqviDbGiVrrLG5V_Z-gmOBm72tFPn2WT995CgCY/edit#gid=617021307
This is the breakdown of the costing/placing. 

Much of it was guessing.  I aimed for a $2 billion for the CM + SM + Ascent + Decent parts.
However there were also parts of the LV that weren't costed/placed yet so I did those too.

Costing of CM/SM/AS/DS was $1,837,750
Overall costing of the parts I added was 3,277,750

PR #96, opened 2015-03-02
Placed and priced RTShortDish2

pjf:


NathanKell:
Great! @pjf FYI our dollar standard is 1965-year dollars, though I think that pricing works fine whatever (good points about balance :) ).

PR #105, opened 2015-03-07
Assorted part fixes from my train journey.

pjf:
See individual commit messages for details.

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -894,6 +894,11 @@ flightControl:
         entryCost: 3000
         cost: 60
 
+    # This is a really crappy reaction wheel. It means players have a hope
+    # of not using their entire RCS reserves if using the RT flight computer.
+    
+    sasModule:
+
     # Salyut parts
     almaz_sr_retro:
         entryCost: 1000
--- a/tree.yml
+++ b/tree.yml
@@ -192,6 +192,10 @@ start:
     B9_Aero_Wing_Procedural_TypeA: *stdwing
     B9_Aero_Wing_Procedural_TypeB: *stdwing
     B9_Aero_Wing_Procedural_TypeC: *stdwing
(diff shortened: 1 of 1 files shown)


NathanKell:
@pjf proceduralConeLiquid is a non-RF part that is hidden when RF is present. However, no Proc Parts has been released with that fix so it's in git only. I suggest not placing it, though, since it's not a supported part--though if placed, as with other parts, it should have cost 0.1.

Other than that, awesome! :)

pjf:
D'oh, how did my notes file get in here? :/

proceduralConeLiquid was already placed, I just popped a price on it. But let me fix that now.

pjf:
Amended. :)

NathanKell:
Sorry, didn't notice this earlier. The procedural wings, like other procedural parts, should have a cost of 0.1. Also we should probably make a FIXME or TODO note about giving wings and control surfaces the same price as a procedural one *of the same size*, not just "tenbux for all!" :)

pjf:
> The procedural wings, like other procedural parts, should have a cost of 0.1.

That's cool. This PR doesn't change the cost of any pWings.

> Also we should probably make a FIXME or TODO note about giving wings and control surfaces the same price as a procedural one of the same size, not just "tenbux for all!"

Done. This is now ticket #107.

Is there anything else blocking this from being merged? ;)

NathanKell:
The issues are:
B9_Aero_Wing_Procedural_TypeA: *stdwing
B9_Aero_Wing_Procedural_TypeB: *stdwing
B9_Aero_Wing_Procedural_TypeC: *stdwing

Although it looks like that's from one of your prior commits? Beyond that, no, looks fine to me. :) Sorry for the delay.

pjf:
Hooray! :) Merging.

PR #108, opened 2015-03-09
Sweep-move engineering parts out of hydrolox.

### LINKED ISSUE #94: Fixing weird parts in hydrolox
Raised in #57. Right now the hydrolox node contains all the IR parts and other things, because it used to be construction. Ideally we should place all these, but if the player installs new mods they may still end up with unplaced parts in odd spots.

This ticket is to track work fixing this.

The aerodynamics node doesn't have much value in RP-0, since they're all things we'd expect to be available early on, but I'm happy for this to be argued otherwise.

---

Possible solutions:

**Change the "Hydrolox" node to overwrite the Aerodynamics node, rather than the construction node.**

- This will break saves.
- The construction node was chosen because of its relations (nodes that it depends upon, and vice-versa).

**Leave hydrolox where it is, and repurpose aerodynamics.**

- I like this more.
- This leaves us with some odd node IDs for what they contain, which is unfortunate.

- [ ] Sweep all parts from aerodynamics to somewhere else (orphans?). Aerodynamics seems to be a bunch of stuff which isn't that hard.
- [ ] Rename the human-readable Aerodynamics label to be (Advanced) Construction.
- [ ] Sweep everything from Hydrolox to new the new construction node.

**Something else?**

- Is there another node we can sweep parts into? If so, we can just do a straight sweep and be done.

pjf:
Closes #94.

### DIFF
--- a/GameData/RP-0/CTTChanges.cfg
+++ b/GameData/RP-0/CTTChanges.cfg
@@ -122,4 +122,59 @@
 		@parents = node4_spaceExploration
                 @pos = -2000,150,-1
 	}
+
+        // Because construction nodes got repurposed for hydrolox,
+        // our unmanned node progression has been repurposed for
+        // construction. They have cheaper costs than you'd expect
+        // for nodes in their position, because we're keeping the
+        // cost of the nodes the tech was displaced from.
+
+        @NODE[node5_precisionEngineering]
+        {
+            @title = General Construction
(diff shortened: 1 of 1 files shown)


pjf:
Killing this in favour of actually making our own hydrolox nodes, otherwise this becomes a game of sweeping nodes to other nodes, and then sweeping *their* nodes somewhere else.

PR #125, opened 2015-03-18
Bobcat's Soviet Engines

BevoLJ:
Placing Only.  No clue on cost.  If someone wants to guess go for it!  :D


NathanKell:
Per IRC, node ids aren't really what they are anymore.
The engines should be placed as follows:
TL4: NK33/43
TL5: RD170
TL6: RD180, RD191
TL7: RD0124

Hydrolox TL5: RD0120
Hydrolox TL7: RD0146

PR #129, opened 2015-03-20
Place AJ 260in FL motor

mcirish3:


jwvanderbeck:
Same thing as your other one.  Spacing is key.  `cost:33081` is not valid. `cost: 33081` is.

pjf:
@mcirish3 : Oh my! 33,081 x 1000 1965 USDs = $33 million in 1965 per unit, not including development costs (which default to 20x the launch costs, but which we can set manually).

$33m in 1965 seems like quite a bit! I don't suppose you've got a citation for this?

Many thanks!

~ Paul

NathanKell:
@pjf, [cite](http://www.astronautix.com/stages/260lidfl.htm) says hello!
You may not be aware that this motor is 3/4 the mass of the Saturn V first stage, with about the same thrust as 5x F-1s.

NathanKell:
That said, @mcirish3, it needs to go [here](https://github.com/mcirish3/RP-0/blob/patch-1/tree.yml#L1285) under heavierRocketry -- it's a TL4 solid motor, not at all related to Advanced Landing. ;)

mcirish3:
Hi Paul,
This is one NathanKell found for me when he was showing me how to do these.  So as it happens I do have a source.  http://www.astronautix.com/stages/260lidfl.htm  I also keep a record of all my sources at the following link along with the list I am working from.(Excel file)  https://drive.google.com/open?id=0B27A074CVw4yVkVNdmtHVG1hbjg&authuser=0
I hope this helps let me know if you have further questions.
~Patrick

Date: Sat, 21 Mar 2015 00:54:32 -0700

mcirish3:
Yep, in my defence at the time I was not tracking on the different node categories.   For some reason at first I was thinking the parts cfg would have that info, which of course it doesn't.  I will work on getting better at it.

Date: Sat, 21 Mar 2015 01:00:36 -0700

NathanKell:
Don't worry about it. :)

pjf:
Also merged with thanks! :)

PR #133, opened 2015-03-21
Agena Avionics now has 300km omni.

### LINKED ISSUE #124: Agena Avionics have mismatched antennae.
FASAAgenaProbe has a 3km integrated antenna, the RP-0 version has 300km.

I'm guessing they should both be 300km.

pjf:
Closes #124.

Not actually tested because clearly I'm living dangerously today.


NathanKell:
Should be on :FOR[RP-0] I think. If RT isn't installed, it'll fail gracefully anyway.

pjf:
@NathanKell : Updated as you wish.

PR #134, opened 2015-03-21
Goo standardisation

### LINKED ISSUE #91: Let's talk goo!  Mass that is..
So I tried to put some goo into orbit today and found it to be a real challenge, which is good, though it might be a bit too hard?  That might be me though and the fact that i'm trying to do it at Early Orbital rather than later.

One thing I noticed though is that the Stack version of the goo masses twice that of the Radial version.  Now at first this might seem logical.  If someone uses the Radial Goo then they have to place two of them for W&B right? So let's make the Stack Goo, of which you would only place one, weigh the same as those two Radial Goos.

Problem with that reasoning, in my mind, is this.  What do you get for the mass?  If you carry *two* Radial Goos, you get *two* different science experiments.  If you carry one Stack Goo you only get a single experiment.  So for the same mass, you get less science.  Unless you absolutely positively **have** to have that Stack Goo, there really is no reason to use it due to the mass cost.  If you are like me, and considered using it just because of looks, and you don't want an oversized fairing, then it is a really hard choice to make because you are intentionally gimping yourself, which isn't fun.

I would argue that the Stack Goo should mass the same as a single Radial Goo, so that the Mass:Science ratio is the same.  If that isn't acceptable then at **most** 1.5 times the Radial Goo, though if a 1:1 ratio isn't acceptable I would prefer something like 1:1.25 maybe.

Thoughts?

pjf:
- Stack goo weighs the same as radial goo.
- Stack and wedge goo cost the same as radial goo.

Closes #91.

### DIFF
--- a/GameData/RP-0/PartHacks.cfg
+++ b/GameData/RP-0/PartHacks.cfg
@@ -74,3 +74,9 @@
 
     // The experiment itself is added in ExperimentAdder.cfg
 }
+
+// Goo mass standardisation. See KSP-RO/RP-0#91
+@PART[SXTProbeGooo]:FOR[RP-0]
+{
+    %mass = #$@PART[GooExperiment]/mass$
+}
--- a/tree.yml
+++ b/tree.yml
@@ -437,16 +437,14 @@ basicRocketry:
         cost: 40
 
     # Stock goo experiment
-    GooExperiment:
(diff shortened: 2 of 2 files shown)


pjf:
Tested with great joy.

PR #140, opened 2015-04-08
Removed unitCost for EC

aw94:


pjf:
What? Free electricity?

[](http://www.smbc-comics.com/?id=2305)

Issue #107, opened 2015-03-09
Price non-pWings to have equivalent cost to pWings.


pjf:
Do you *really* like making things and comparing them to other things? Gosh, do we have the job for you! In an ideal world, all our wings and control surfaces would have costs equal to their pWing equivalents, rather than being ten spacebucks regardless.

This ticket satisfies @NathanKell's request in #105.

aw94:
If no one else has worked on this yet, I'll take a shot

pjf:
Closed by #145.

PR #145, opened 2015-04-17
Non p wing pricing

aw94:
Takes care of #107, I think. May have some other wings to price, but those also need to be placed, so they'll go elsewhere.


pjf:
You are *awesome*.

aw94:
:)

PR #148, opened 2015-05-17
Updated Centaur costs

pjf:
Via @Dasms:

> NK said so.


NathanKell:
$15 million seems a bit low for Centaur development. Where was that from?

NathanKell:
@Dasms poke

NathanKell:
@Dasms poke again...

PR #158, opened 2015-07-16
Added sounding rocket altitude records

Shurikeeen:


NathanKell:
Awesome! :)

Quick question: Isn't 'name' not shown and only used internally? That's why I gave it a machine-readable name (like part names)...

Shurikeeen:
I think so. Does it matter what the title is?

NathanKell:
The title will be the thing actually shown in the GUI, so yes it matters what the title is. The name field, however...MM can't deal with things with spaces, and it's best to make names shorter (and more info-rich) so they can be easily tagged by other things. So I'd definitely suggest having a name that uses only alphanumeric characters and dashes and underscores, and leave the prettiness to title. :)

Shurikeeen:
Ah, I see.

Shurikeeen:
Things are kinda borked in the current commit. I'm releasing a patch what also should add speed records in maybe the next half a hour.

NathanKell:
Ah, sorry, should have asked before mergi...ah, screw it. 1min.

NathanKell:
Added you. No more PR jiggery. :)

PR #163, opened 2015-07-19
RN Proton + R-7 Avionics

Asgarmagol:
Updated the Proton Avionics to be able to control the craft.

### DIFF
--- a/GameData/RP-0/Avionics.cfg
+++ b/GameData/RP-0/Avionics.cfg
@@ -366,3 +366,32 @@
 		massLimit = 20.0
 	}
 }
+
+// Config for RN R-7 (Block A)
+@PART[r7_blok_a_sput|r7_blok_a|r7_blok_a_soyuz|r7_blok_a_molniya]:FOR[RP-0]
+{
+	MODULE
+	{
+		name = ModuleAvionics
+		massLimit = 280.0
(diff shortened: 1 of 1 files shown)


Asgarmagol:
Added R-7 configs too

GameData/RP-0/Avionics.cfg
```diff
@@ -347,3 +347,51 @@
...
+		massLimit = 700.0
+	}
+}
+// Config for RN Proton (Block D-M2)
+@PART[rn_proton_blockd]:FOR[RP-0]
+{
+	MODULE
+	{
+		name = ModuleAvionics
+		massLimit = 20.0
```
NathanKell: @Asgarmagol is 20t really enough for the Blok D as well as the L3 complex?

NathanKell:
Looks great, the exception being the line note I added--since Blok D was also used for TLI of the N1-L3 complex it probably  needs more avionics?

GameData/RP-0/Avionics.cfg
Asgarmagol: I was only looking for Block D in combination with Proton. 20t is the max payload of the Proton. Block D is part of the Payload. so it is correct for this usage.

PR #168, opened 2015-07-22
Add a high-level RP0 group

jrossignol:
Recommended for grouping in the debug menu (and a user-settings menu coming in 1.6.0).  Also allows you to put a min Contract Configurator version - use this post-release if you make a change that depends on a feature introduced in a new version of Contract Configurator, it'll save you from investigating bug reports from users that don't update the dependencies properly.


NathanKell:
Ooh, good call.

PR #190, opened 2015-08-03
TemperatureScan.cfg updated & BioSample.cfg started

jhathawaytn:
TemperatureScan is complete Mecury - Mars&Moons, some Saturn entries.


NathanKell:
Yikes! Totally missed this, sorry! Lost net access for a while and must have missed this on return.

PR #191, opened 2015-08-03
Place R-7 main engines from KWRocketry

SirKeplan:
not too sure about the upgrade costs for the later configs, but everything else should be ok i hope.


NathanKell:
Whoah I totally missed this! Sorry!

jhathawaytn:
No worries. .still needs lots more added.

Issue #177, opened 2015-07-25
Contracts moving forward


NathanKell:
We currently have:
- First Launch contract
- Speed/altitude records (crewed/uncrewed)
- Altitude-only sounding rocket contracts
- Pass the Karman line (crewed/uncrewed)
- First orbital-speed reentry (uncrewed)
- First orbit (crewed/uncrewed)
- First flyby, impact, orbit of Moon
- First GEO
- First EVA
- First docking
- Other planet flybies
- Landing (Moon, Mars/PhobosDeimos, Venus)
- Repeatable human orbital

We also currently use the stock satellite and collectScience contracts (pending replacement with CC contracts).

However, there's lots more to do.
Boldly stealing contract ideas from BARIS:
- [ ] Sounding rocket contracts generally, not just altitude records. For example, loft science payload X to Y km over biome Z ( @Felger aurora sounding rockets!); launch to MEO or HEO orbital altitude; do one of those on a certain day to catch some phenomenon.
- [ ] Earth Science Satellite contracts at all rep levels with sane inclinations (~30-60 degrees).
- [x] Comm network contracts (LEO, MEO, and GEO)
- [x] Film return: orbit over biome X at altitude Y, take pictures (i.e. activate material bay), use Ship Manifest to move science to reentry module, return film.
- [ ] Bio sample missions (LEO, HEO, translunar, etc). For flies/mice/etc this can be just the goo canisters, but for dogs/chimps we'd want to return and use a heavier payload.
- [ ] _Sane_ test contracts, i.e. test an uncrewed capsule/lander before humans fly in it, all-up lunar landing hardware test, test Mars mission in Earth-Moon space, etc.
- [ ] Probes: replace the existing Explore contracts. Flyby, impact, fly by the dark side of the moon and transmit, landing, sample return. Surveying missions (polar or high-inclination). Grand tour. Primary + satellites mission (bouncing between moons and Jupiter, say).
- [ ] Docking: automated, crewed: repeatable training contract that gives XP?
- [ ] HSF: better repeatable orbitals, duration (3-5 days, week, 10 days, two weeks, +), number of crew, perform EVA, docking (as above), translunar, lunar orbital, GEO, polar lunar, lunar landing, translunar missions.
- [ ] HSF records: duration, anything else?
- [ ] X-Planes: fly at x altitude at y speed for z time, reach X altitude record and return without staging, reach X speed record and return without staging.
[...]

PR #193, opened 2015-08-25
Add contract for three-person into lunar orbit

Arbos:
This contract resembles Apollo 8 mission, which was the first manned
flight to moon and made 10 lunar orbits in 20 hours.

For #177.


NathanKell:
Very cool! And I like how in the contract text it's all properly gender-neutral. :)

A request: could you make it a bit more generic? RP-0 tries to offer fairly generic contracts, so for example we just require *a* lunar orbit, with a fairly low periselene, for the uncrewed orbital, and fairly large Ap/Pe ranges for the HSF orbitals. So I'd suggest merely requiring a minimum of 1 crew (to allow for, say, Zond style shots, or Lunar Gemini), and having a larger Pe window.

Arbos:
Yes, I will do it.

NathanKell:
Awesome, thanks! :)

PR #195, opened 2015-08-27
Place the KVD-1/RD-56

SirKeplan:
It was meant to fly on later N1 designs, so placing it with NK33 makes sense to me.


NathanKell:
Hydrolox should go only in the hydrolox nodes. That'd be TL4 hydrolox, I think?

SirKeplan:
yeah, that probably makes sense..... forgot about the hydrolox nodes.

PR #198, opened 2015-09-05
Fix bug with toggling avionics

kendfrey:
The "Shutdown Avionics" event only updated the name to "Activate Avionics", and nothing else.


NathanKell:
Whoops. Thanks!

PR #202, opened 2015-09-11
Battery Costing

BevoLJ:
Costing the small batteries that no one uses since we all use Proc Parts batteries.  Just want to get rid of the non-ro / no costing title. =D

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -772,14 +772,11 @@ generalRocketry: # TL1 Late 1950s/1960 tech, first real orbital LVs. Atlas, R-7,
 
 basicConstruction:
     batteryPack:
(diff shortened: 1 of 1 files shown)


NathanKell:
Cool! :)

Uh...these seem a little low to me. EC costs 1 per 200 units in procedural parts, plus the cost of the part itself, so I'd suggest upping the bankMini to 110 (and proportionally the same for the others). Also, you don't need to specify an entryCost in these cases, it will be calculated from the cost. :)

NathanKell:
(that's 20500EC = 102.5 funds, plus 7.5% for the casing)

PR #208, opened 2015-09-17
Need my docking ports

shoe7ess:


NathanKell:
Uh, this patch failed to build. Also you're trying to place 2000s-era docking ports in a node from the mid 1960s...

PR #221, opened 2015-09-25
Docking Part Attempt 3

shoe7ess:


SirKeplan:
you are getting build errors......... that means you have done something wrong in tree.yml . try running 'make' or 'perl bin/yml2mm' before making a commit or pull. it should tell you the error.

shoe7ess:
I'm completely knew to editing these types of files. They aren't as simple as adding a part to a particular node client side, I'll look up a bit of a tutorial on what you're asking and try and do something with it

SirKeplan:
What do you mean by client side in this case? The procedure to place an item in the tree is just to edit the yaml file, then build it into a config with perl.

NathanKell:
We're currently reorganizing the tree a bit, but we'll make sure that docking ports get in the next update or so. :)

shoe7ess:
Yay! If you could add spots for the 1980's and beyond that would be helpful
as well. I use a number of mods at don't even appear in the tree. Cryogenic
freezing mod and so on. I have noticed parts from a number of my mods (like
non-androgynous docking ports, which come with male and female ends) and a
few I can't think of on the top of my head aren't anywhere in the tech tree
(checked through sandbox). My question is: is there a way there can be an
open/temporary node where these parts can appear, even if some are
construction and some are structural and so forth? I may be wrong in
thinking the parts don't show up because they aren't integrated into the
RP-0 Tech Tree, but I'll attempt a sandbox game after removing RP-0 and
check the part counts against eachother just to be sure.

shoe7ess:
Also, when I reach the point, I plan on adding a space-shuttle and some ISS
mods, and afterwards "futuristic" technology (like KSPinterstellar). Will
parts for mods like these (I'm going to be using the Space Shuttle SSP
because it's simple and takes up little room, as well as the same mod
author's ISS mod, for the same reason) be added towards the end of the
tree? I don't know enough to add my own nodes and make the parts appear
manually (and by client-sided, I meant editing the tree.cfg and making it
read-only, which didn't seem to change the tree whatsoever). Also
remembered that the ZZZ_Flags flags mod parts aren't appearing in the tree,
even if I create an MM to add the parts in last (using the :FINAL line now,
may change to :AFTER [Realism Overhaul] | [RP=0] and seeing if that works.

PR #227, opened 2015-09-29
Stock mk2 parts - Spaceplanes

winged7:


NathanKell:
A few things with this:
1. While the part namer ingame uses . that's only because KSP replaces all _ with . when compiling parts. So they should in fact be _ (you can verify by checking the part cfg that creates the part)
2. The crew cabin should probably be a bit cheaper than the cockpits, since it has no command module, even though it carries six people.
3. Speaking of prices--the X-15 parts (Mark1 1.25m) have 1800 skin max temp, the Mk1 parts (in #229 ) have 2200, and these have 2500. I'd suggest using these prices for the Mk1 parts (9000 and 8000, say), and then pricing these a bit higher yet.
4. Entries should be made in EntryCosts.cfg (like with the 1.25m X-15 parts) so that the heat shield research cost is shared between parts using it. That's why the entry cost is so high, because half of it you only pay once (it represents researching the shielding materials), the rest you pay per part.

That said, don't let that detract from my giving you a BIG THANKYOU for doing this! :)

NathanKell:
Oh, also, 5: not sure what to do about the docking port, it should probably go in avionicsTL7 since it's the NASA docking port.

NathanKell:
Ah, dangit, missed perhaps the most important bit: Shouldn't these go in advAerodynamics ?

winged7:
ok, I moved it to advAerodynamics, now crew cabin is cheaper, cockpits more expensive, docking port moved to avionicsTL7 

all . replaced with _

NathanKell:
These look good to me, although let me tag you on an issue re entry costs because the EntryCostModifiers need to be created. #227

PR #228, opened 2015-09-29
Stock parts - Cupola, LargeCrewedLab, CrewCabin

winged7:


NathanKell:
Seems reasonable to me!
As a future note, if no entryCost is specified, the default (20x the cost) is used, so if you're just setting it to 20x the cost you don't have to set it at all :)

PR #229, opened 2015-09-29
RO-Mk1cockpits - cost

winged7:


NathanKell:
Other than the notes on #227 I have a quick question here: why the massively lower price tag for the inline cockpit?

winged7:
in the latest RP-0 release cost of the mk1cockpit (1person) is 4x bigger than mk1 Inline cockpit (1person)

winged7:
Ok I see that now it was changed to 3000 for Mark1Cockpit and 2500 for Mark2Cockpit (inline)

winged7:
I updated costs: now it's 6000 for 2person Mk1 cockpit and 5000 for 2person Mk1 Inline cockpit.

NathanKell:
These look good to me, although let me tag you on an issue re entry costs because the EntryCostModifiers need to be created. #227

PR #230, opened 2015-09-29
Stock wheels, external seat, rover body

winged7:


NathanKell:
That seems very reasonable to me, although maybe the seat should cost a bunch more, maybe 10x that?

winged7:
Ok, now the seat is 10x more expensive :)

PR #231, opened 2015-09-29
Stock parts

winged7:
-decoupler
-radial attachment point
-stationhub
-RTG
-DroneCore
-Ion engine


NathanKell:
For now I'd suggest not placing the drone core and RTG, pending @Niemand303 's update of the electronics/probes subtree, but the structural bits look fine!

winged7:
I've deleted RTG and DroneCore from my update.

NathanKell:
Thanks! :)

PR #232, opened 2015-09-29
Stock engines - costs

winged7:
Aestus II
AJ10-190 OME
KVD-1
RD-855
RD-856

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -1740,10 +1740,10 @@ highAltitudeFlight: # Early Spaceplanes
 # TIER 4
 heavyRocketry: #TL3 Mid-late 60s tech. F-1, H-1 uprated, etc.
     radialLiquidEngine1-2: # R-36/Tsyklon vernier FIXME cost
-        cost: 600 #total guess (similar to the other 50-100kN hypergolic engines)
+        cost: 350 #total guess (similar to the other 50-100kN hypergolic engines)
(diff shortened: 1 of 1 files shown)


NathanKell:
Um, there should actually be numbers available for the CE 7.5 (aka KVD-1). Aestus II, and AJ10-190, so no need to guess. :)

on the verniers, I'd suggest only ~60 for the small ones, and maybe 350 for the large ones (they have a very low Isp for their era, so not much nozzle expense).

winged7:
I've updated cost of Tsyklon verniers. Now its 60 for the small one and 350 for the large one.

Regardng Aestus II, CE 7.5 and AJ10-190 - no idea where to find actual numbers, there is no info on astronautix.com.

NathanKell:
Yep, that's what makes placing and pricing parts nontrivial. :] AJ10-190 there should certainly be sources for, the costs of STS have been argued about so much there must be info somewhere. For Aestus, if Astrium doesn't have the data itself, I guess it's time to google? And for CE 7.5, same....
(Yes, most of these expeditions descend into googling, hence all the URLs in the tree :D )

winged7:
Sorry, I haven't time for googling :). I can delete costs for those engines if you want.

NathanKell:
Yeah, probably best for now. Make an issue, maybe someone else has time?

PR #236, opened 2015-09-30
AIES command pods

winged7:


NathanKell:
Pending @Niemand303 's rework of the probe/electrics subtree, and pending figuring out exactly what to model the AIES probe parts off (there's an open issue on that, either here or on the RO repo...) I'm going to ask you hold off on this one. Sorry! :(

PR #239, opened 2015-10-01
Added FASA Geminy Pod White

federicoaa:
Added FASA Geminy Pod White, an upgrade version of the Geminy Pod.
Since it is an upgrade, I placed it in the next node (commandModules)
Price adjustment is based on the FASA configs.

entry cost: 270m +16m -> 350m + 16m
cost: 5400 -> 6000

PS: this is my first time in github and modding in general, I'm still kinda lost in here. Apologies in advance if I messed up something :)

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -2035,7 +2035,7 @@ commandModules: # Third generation capsules, it would seem.
     Alnair_Engine_A: *ApolloLES
     # FASA Gemini Pod White
     FASAGeminiPod2White: &GeminiPodWhite
-        entryCost: 366000 # was 350m, +16m for shielding
+        entryCost: 316000 # was 300m, +16m for shielding
         cost: 6000
 
 spaceExploration: #crewed landing
--- a/tree.yml
+++ b/tree.yml
@@ -2033,6 +2033,10 @@ commandModules: # Third generation capsules, it would seem.
     Alanir_Crew_A:
     Alnair_Decoupler_A:
     Alnair_Engine_A: *ApolloLES
+    # FASA Gemini Pod White
+    FASAGeminiPod2White: &GeminiPodWhite
+        entryCost: 366000 # was 350m, +16m for shielding
+        cost: 6000
 
 spaceExploration: #crewed landing
     landerCabinSmall:


NathanKell:
You messed nothing up, you did it all right! :)

Quick question, though--how exactly is it an upgrade? I had thought our patches meant that the white and black versions were essentially identical...

federicoaa:
I called it upgrade because it adds some things, like reaction wheels and
more mono. Besides, it's lighter (2 tons vs. 1.5 tons)

NathanKell:
Ah, I was confused, our White Gemini is slightly different.
Although I think you're mostly thinking of the stock KSP + FASA pod, not the RO pod--the RO White Gemini is only a hundred kg lighter, and has some more batteries and life support supplies.
See: https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/RO_SuggestedMods/FASA/RO_FASA_Gemini_Pod.cfg

Given that, I'd make the entry cost only a bit (20,000? ) higher than Gemini (and use EntryCostModifiers to remove the entry cost of black Gemini parts from the White Gemini parts), though the actual cost of 6000 seems fine.

federicoaa:
Changed entry cost to 316000.

I'm not sure about what EntryCostModifiers is.

NathanKell:
Sorry, busy few days. EntryCostModifiers are the things that let us say "You unlocked an LR89? Now any other mods' LR89s are free to unlock." Or "You unlocked an LEO heatshield? You only have to pay the major resource cost once, you only pay a little bit now to unlock other sizes".

Here's an example. You can see that 5000 funds is removed from the entry cost of either part if you unlock the other. https://github.com/KSP-RO/RP-0/blob/master/GameData/RP-0/EntryCostModifiers.cfg#L25-L41
What's best is probably to add a entry for the white part, and add a subtractor of the black part's entry cost, linked to the black part's name.

federicoaa:
I see, something like this?

PART
{
name =  FASAGeminiPod2White
entryCostSubtractors
{
FASAGeminiPod2White = 286000
}
}

NathanKell:
I...guess you can tell when I hit crunch mode for 1.0.5. :D
Yeah, that sort of thing, although you'd want to have name = (this part) and in the subtractor, the _other_ part's name, you're saying right there that the part subtracts from...itself.

PR #240, opened 2015-10-01
RD-170/171, RD-191 - cost

winged7:


NathanKell:
tagging @Niemand303 in case he has better numbers, but putting these in for now. :)

Issue #237, opened 2015-09-30
Spaceplanes and entrycosts


NathanKell:
For the X-15 parts, we have a known R&D cost. That's fine. Same with STS (though it would probably break the bank).

However, for DynaSoar like birds, we're speculating.

Like the X-15, the best route IMO is to have some portion of each part's entry cost be for the shielding, and some part of it be for the part itself. Once you buy one of the parts at the full price, you don't have to pay the shielding R&D cost for the rest.

So...what sounds like a reasonable R&D cost for the Mk1 System (i.e. could be DynaSoar) shielding? I would guess something like 30000.
The Mk2 system's shielding would be about 50000 I think, and the Mk3's over 100k.

Then, as for parts, another hundred k or so for the cockpit of the Mk1 system (150k for the Mk2)?

@winged7

winged7:
as I said before i can increase development cost for the stock mk2  system 2x so it will be 260k for the cockpit standard and thus 86666 for substractors

NathanKell:
Following from #353
Yeah, 260K sounds good to me, or even an even 300K (with the MK1 system at 180K for the cockpit).
Apologies for being unclear, I did indeed mean 1/3 of dev cost, not unit cost.

So how does this sound: 
For Mk1 system, 180K for the cockpit, with 30K of that being the shielding (and a modifier as such). Tanks would have an entry cost of about 50K each (before 30K subtract). Crew can at 60K. The inline and regular cockpits should have multipliers of 0 for each other, too.
For Mk2 system, 300K for the cockpit (and 0 multipliers between inline and regular), of which 60K is shielding general cost. 80K for tanks, 120K for the crew compartment, 100K for cargo bays.
For Mk3 system, roughly STS-level numbers, of which, hmm, do we know how much TPS development cost? Something like 800K for the cockpit, maybe? 200K each tanks and cargo bays and 300K for the crew section? With a subtractor of 150K for shielding?

NathanKell:
*Note*: Those are all entry costs, not unit costs.

Unit costs should be probably on the order of 1/10th that, maybe less (1/15th? 1/20th?)

winged7:
ok so 
180k for mk1 system and 30k substractor for shielding
300k for mk2 system and 60k substractor for shielding
800k for mk3 system and 150k substractor for shielding

winged7:
and unit cost 1/20th

winged7:
[...]

PR #243, opened 2015-10-01
Mk3 system - pricing

winged7:
https://github.com/KSP-RO/RP-0/issues/237

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -2456,6 +2456,8 @@ hypersonicFlight:
 
     mk3Cockpit_Shuttle:
         cost: 26000 # 2x bigger than mk2 cockpit standard
+    mk3CrewCabin:
+        cost: 19000 # much higher than for Mk2crewCabin
     mk3CargoBayL:
         cost: 500
     mk3CargoBayM:
--- a/tree.yml
+++ b/tree.yml
@@ -2455,15 +2455,15 @@ metaMaterials:
 hypersonicFlight:
 
     mk3Cockpit_Shuttle:
-        cost: 26000 # 2x bigger than mk2 cockpit standard
+        cost: 40000
     mk3CrewCabin:
-        cost: 19000 # much higher than for Mk2crewCabin
+        cost: 15000 # much higher than for Mk2crewCabin
     mk3CargoBayL:
-        cost: 500
+        cost: 10000
     mk3CargoBayM:
-        cost: 250
+        cost: 5000
     mk3CargoBayS:
-        cost: 125
+        cost: 2500
     mk3FuselageLF_100:
     mk3FuselageLFO_100:
     mk3FuselageLF_50:
--- a/tree.yml
+++ b/tree.yml
@@ -2460,10 +2460,13 @@ hypersonicFlight:
         cost: 15000 # much higher than for Mk2crewCabin
     mk3CargoBayL:
         cost: 10000
+        entryCost: 200000
     mk3CargoBayM:
         cost: 5000
+        entryCost: 200000
     mk3CargoBayS:
         cost: 2500
+        entryCost: 200000
     mk3FuselageLF_100:
     mk3FuselageLFO_100:
     mk3FuselageLF_50:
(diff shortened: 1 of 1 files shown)


NathanKell:
Pending input on #237

winged7:
Not sure what I need to do here.

winged7:
So for example: Mk3 Cockpit should be cheaper to unlock than I unlock Mk2 Cockpit first? How much cheaper?

NathanKell:
I'd say maybe 1/3 the cost of the cockpit should be the shielding system in each case.

winged7:
something like that?

@PART,*:HAS[#tag[heatshieldedLEO]]
	{
		maxSubtraction = 16000
		entryCostSubtractors
		{
			0625-Heatshield = 16000
			Heatshield-1M = 16000
			1-25-Heatshield = 16000
			mk1pod = 16000
			FASAMercuryPod = 16000
			rn-vostok-sc = 16000
			FASAGeminiPod2 = 16000
			rn_voskhod_sc = 16000
			mk2CrewCabin = 16000
			mk2Cockpit_Standard = 16000
			mk2Cockpit_Inline = 16000
			mk3Cockpit_Shuttle = 16000
			mk3CrewCabin = 16000
			B9_Cockpit_MK2 = 16000
			B9_Cockpit_S3 = 16000
			B9_Cockpit_S2 = 16000
			B9_Cockpit_MK5 = 16000
			B9_Cockpit_M27 = 16000
			HL_Aero_Cockpit = 16000	
			B9_Cockpit_S2_Body_Crew_6m = 16000
			B9_Cockpit_S2_Body_Crew = 16000
			B9_Cockpit_MK2_Body_Crew_2m = 16000
		}
	}

but with much bigger number for mk3cockpit right? 1/3 of the develompment cost =  173300

NathanKell:
Yep, where you want to tag with say Mk3Heatshield and then have a subtractor (and max subtractor) of, say, 1/3 the cockpit price (which would probably be 4/5 of a fuel tank price and 2/3 of a cargo bay price?)

Uh...to be clear _not_ just adding the new parts to that old tag. The point is that there should be one of these set up for each "type" of shielding -- LEO ablative, lunar ablative, X-15, DynaSoar/Mk1, Mk2/S2, Mk3/S3/etc.

winged7:
Sorry but I'm giving up here. The whole concept of entryCostModifiers is not so very clear to me.

NathanKell:
Ok. Let me write a quick wiki entry on them, and see if that's better. If not, then yeah just leave them unpriced.

NathanKell:
Does this help?
https://github.com/KSP-RO/RP-0/wiki/Entry-Cost-Modifiers

winged7:
Examples in your tutorial are very easy to understand but I still don't know waht exactly I need to do.

https://github.com/KSP-RO/RP-0/pull/353

NathanKell:
Yep, exaclty!
Though per the note on #353 the entry costs probably need to come up--I'd expect something like 20000 for the shielding component of the Mk1, 40 to 50000 for the Mk2/S2, and maybe 100000 or more for the Mk3/S3 (STS is _expensive_).
(so, say, the Mk2 cockpit would be something like 100000 entryCost, with 40-50 of that being the shielding.)

winged7:
but the 4333 is not 1/3 x development cost, this is 1/3 x unit cost.

1/3 of  develompment cost for the mk2 standard cockpit  = 43333 (exactly like you want: between 40 and 50k)

NathanKell:
Ah, I thought the entryCostSubtractor you wrote was 4333. I missed a 3? Oops. :D

winged7:
I switched from 1/3 of the unit cost (4333) to 1/3 of the development cost (43333 - 10x bigger) however 130k for development cost is still a bit cheap compared to Mercury so maybe multiply increase it 2x to 260k?

winged7:
I can delete mk2, mk3, S2, S3 etc. development cost value from tech tree so it will be 20x more expensive than unit cost by default and thus 2x more expensive than now.

NathanKell:
Moving discussion to #237 til we get this figured out. :)

PR #258, opened 2015-10-05
KIS - Kerbal Inventory System

winged7:


stratochief66:
is KIS sufficiently supported in RO? if so,

@NathanKell looks good

winged7:
KIS doesn't require any RO support.

PR #260, opened 2015-10-05
B9 - jet engines

winged7:


winged7:
Also B9_Jet_Pod_Small is now GE TF-34, not JT8D

PR #269, opened 2015-10-13
Merlin engines -  cost

winged7:
http://space.stackexchange.com/questions/4388/what-are-the-costs-of-the-various-different-engines-in-current-use


stratochief66:
@NathanKell looks good

PR #271, opened 2015-10-13
RD270 cost

winged7:


stratochief66:
Hmm, it is much lighter than the F-1 and has better ISP. I would say as a guesstimate it should be a bit more pricey, may be starting at 20% more.

That said, I think this may be an acceptable starting point for the pricing. @NathanKell @winged7 ?

winged7:
Ok, updated. Now it's 18000 for RD270 and 21600 for RD270M.

PR #288, opened 2015-11-01
Soviet Engines

winged7:


SirKeplan:
It's not always appropriate to give equivalent engines the same price, i think the RD0146 was done in some kind of collaboration with Pratt and Whitney, but even so Russian engines are often significantly cheaper than their American counterparts.

stratochief66:
I agree SirKeplan. However, I think giving equivalent pricing until better pricing data can be found is a good temporary value.

Also, it would be handy to have TestFlight configs to help differentiate similar engines, since we know that many of those cheaper russian engines weren't as reliable.

PR #293, opened 2015-11-04
KW rocketry engines, solid boosters

winged7:


stratochief66:
@NathanKell looks decent as temporary pricing. Somes values might need some future finetuning.

winged7:
I spend about 2 hours googling and trying to find proper numbers. My numbers aren't based on real data but they're pretty well balanced.

NathanKell:
Huh, because I just went to the 4A listing on astronautix and it had a stage cost of 2.03m (in USD 1985, one presumes). So I added that during the merge. Didn't check the others, but do please check astronautix.com for the stage in question when pricing SRBs. :)

PR #297, opened 2015-11-05
RAPIER

winged7:


stratochief66:
@winged7 we might want a little better than just a total guess for this part. How does the price sit compared to other parts with a similar thrust/ISP?

winged7:
Rapier as jet engine gives you little better TWR than J-58 (cost 1200) but little less thrust. Also it will probably not explode at Mach 4 - Mach 5 speeds due to overheating (J58 temperature limit is Mach 4.1)

Rapier as rocket engne is something like half of the LR105-NA-6 (cost 700) - comparable vacuum isp but half of the thrust.

Maybe 1900. ?

NathanKell:
That seems a very low price for a combined-cycle methalox UBERENGINE, tbh.

winged7:
uberengine but with crappy isp (270-305) compared to other methalox engines (~350)

maybe 2x more (3800) ?

NathanKell:
Hmm, that sounds like RO doesn't actually support the RAPIER and it's using default KSP Isp (and thrust, probably) with AJE just changing the airbreathing mode and the propellants). Especially given https://github.com/camlost2/AJE/issues/44 we need to configure it ourselves, so...what seems reasonable?

winged7:
https://github.com/camlost2/AJE/issues/45

winged7:
hmm maybe 4800?

NathanKell:
Yes, as I tagged you in the comment directly above yours, it's now the responsibility of RO to give them proper propellants and Isp. :)

PR #298, opened 2015-11-05
Tu-95 engine

winged7:


NathanKell:
Tagging @Niemand303 in case he knows about NK-12 costs.

NathanKell:
Also doubled the cost; it's a beast of an engine.

PR #299, opened 2015-11-05
SXT MEM

winged7:


stratochief66:
@winged7 

Are these parts meant to peform as a LEM substitute/model?

If so, perhaps &/* tag another LEM part and this one so their prices are linked rather than independant?

winged7:
They are not exactly the same for example ascent stage is with built-in ascent engine and docking cone so the price is adequately bigger than for FASA part.

winged7:
I just saw that they are already placed and costed by Niemand303.

PR #300, opened 2015-11-05
Reaction wheels

winged7:


stratochief66:
What is your basis for grounding the guesses? Are the prices proportional to their power?

http://www.millennium-space.com/brochures/RWA1000Brochure.pdf

http://www.sst-us.com/shop/satellite-subsystems/aocs/10sp-m-small-satellite-microwheel-3-unit-package

These might prove useful for grounding estimates?

winged7:
Those reaction wheels are very small compared to ours.  
Also one question: 
@Pitch torque = 0.1 in the RO config is the same as 0.1 N.m here:  http://www.millennium-space.com/brochures/RWA1000Brochure.pdf (model 100SP-0)?
Or maybe 0.1 in RO config means 0.1kN.m ? 

Price for 100SP-0 is ~100.000 USD http://www.sst-us.com/shop/satellite-subsystems/attitude-and-orbit-control-systems/100sp-o-small-satellite-microwheel-3-unit-package

stratochief66:
I think 0.1 would be kN.m

My dislike of reaction wheels puts me in a conflict of interest. I think they should only be used (in RO/RP-0) for small satellites, as they are in reality. Any stock or mod parts that use reaction wheels to stabilize the craft are entirely unrealistic and without realistic precedence, so we can't really give them realistic prices, right?

Perhaps they should just be thrown in with these temp prices to see how the perform or are used, I can't really say.

@NathanKell ?

NathanKell:
Ya, IIRC the RO stats for them are incredibly OP as things stand, and desperately need a fix. Probably these prices will be correct once the RO stats are properly nerfed. :)

winged7:
so maybe 5x more than it was before?

NathanKell:
I wasn't talking about cost, I was talking about fixing our RO configs.

PR #305, opened 2015-11-05
Ion engine - pricing

winged7:
look at page 3:
http://esto.nasa.gov/conferences/nstc2007/papers/Manzella_David_D10P2_NSTC-07-0116.pdf

"In fact more than 40mln USD in cost overruns were directly related to the ion propulsion systems xenon tank and ion thruster power sources placing the cost of the Dawn ion propulsion system at more than 50mln USD."

So 50mln USD for the xenon tank and 3 NSTAR engines? How much for one engine?

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -2560,7 +2560,7 @@ fieldScience:
         entryCost: 500
 
 ionPropulsion:
-    ionEngine: #stock ion engine
+    ionEngine: #stock ion engine - NSTAR
 
 largeElectrics:
     # Remote Tech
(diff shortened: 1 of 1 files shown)


stratochief66:
That is an interesting number to learn, but the cost overrun isn't the whole cost. We would want to know the initial cost as well, right?

Also, this looks like useful information leading toward the Unlocking price, not necessarily the "per unit" cost

winged7:
Ok so maybe 50mln USD (2007 USD) for development cost and 20x less for unit cost?

PR #309, opened 2015-11-06
Orion

winged7:
According to new tech tree:https://camo.githubusercontent.com/7f7959e9a74bd0bd2e269a06d98a690770529d10/687474703a2f2f692e696d6775722e636f6d2f6b79746c704e502e706e67  Long Term habitation seems to be the best place for the Orion parts.

development cost: http://www.choicesarizona.org/dbe/sls.html
unit cost: http://www.thespacereview.com/article/2330/1

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -2842,6 +2842,28 @@ longTermHabitation:
     crewCabin:
         cost: 9000 # 2x bigger than mk2 lander can
         entryCost: 180000
+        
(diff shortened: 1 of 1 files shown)


stratochief66:
Yeah, I think that placement makes sense. Both orbital habitats like Bigallo & advanced Lunar/Mars capsules should only be available after Apollo level techs

PR #314, opened 2015-11-07
B9 intakes

winged7:


NathanKell:
Aren't the RMN and RMM basically the same as the ramAirIntake ? Perhaps even larger? Seems weird they'd be cheaper.

winged7:
Their intake area is 20% smaller than for ramAirIntake. I increased their price a little.

NathanKell:
Ah, ok.

PR #316, opened 2015-11-07
RL10 from Chaka  - configs

winged7:
https://github.com/KSP-RO/RP-0/pull/317

https://github.com/KSP-RO/RP-0/pull/327


NathanKell:
So this is the combination of #317 and #327 ?

NathanKell:
Or, if not...not sure what the references mean.

winged7:
No, this isn't combination of #317 and #327. It's a whole different thing:  there are 6 various configs for this engine so every other version needs its own pricing. RL10C-1 will cost you 6000 but for CECE-Base you will pay only 3800. The same as for FASA RL10 configs.

NathanKell:
Ok. Um, so why the links then? I'm confused.

winged7:
because config for Chaka RL10 engine is entwined with placing & costing Chaka RL10. Isn't?

 You can't just left RL10 with 5300 price because there are various config for it: some are cheaper, some are more expensive.

NathanKell:
Ah, ok. That's the sort of thing that should all be in one branch though (and there'd be fewer merge collisions :) ).

NathanKell:
Uh, also, this file isn't of extension .cfg and so KSP won't see it...

winged7:
updated to .cfg extension

PR #320, opened 2015-11-08
mk2CargoBay - pricing

winged7:


NathanKell:
See general discussion we've been having about spaceplane part entrycosts and thermal shielding systems.

PR #323, opened 2015-11-08
Gemini white

winged7:


stratochief66:
@NathanKell setup the Gemini White to be a bit more powerful than the standard part, should it be made more pricey NathanKell?

NathanKell:
Ya, see #239 which also works on this for more details?

winged7:
So Gemini White is more powerful? I can make it more pricey if you want.

winged7:
316k for entryCost and 6k for unit cost. exactly the same as what federicoaa proposed.
https://github.com/KSP-RO/RP-0/pull/239/files

NathanKell:
The entry cost modifiers need to also be made, since I presume it wouldn't be right to research this then have to pay full price for Gemini, and vice versa.

winged7:
entryCostModifiers for Gemini: https://github.com/KSP-RO/RP-0/pull/352

NathanKell:
Closes #239

PR #326, opened 2015-11-09
Stock radiators

winged7:
Could someone tell me where I should put them?


NathanKell:
Per #328 closing until RF etc makes them actually work for tanks. :)

PR #328, opened 2015-11-09
HeatPump

winged7:
No idea where to put this thing.
https://github.com/Starwaster/HeatPump/releases


NathanKell:
Heh. Probably best in thermal control, which should get reconnected to the tree...but it's all contingent on RF updates that take advantage of 1.0.5 to make radiators matter in analytic. For now I'll close this until we get an RF update to make them useful.

PR #331, opened 2015-11-10
Update Orbital 1 Repeat.cfg

ctiberious:

### DIFF
--- a/GameData/RP-0/Contracts/Human Spaceflight/Orbital 1 Repeat.cfg
+++ b/GameData/RP-0/Contracts/Human Spaceflight/Orbital 1 Repeat.cfg
@@ -6,7 +6,7 @@ CONTRACT_TYPE
 	
 	title = Human Orbital (1/LEO)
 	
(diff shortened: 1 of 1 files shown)


NathanKell:
@ctiberious  I'm confused. I merged your earlier PRs about these contracts--was that a mistake? Do these PRs supersede them? Please let me know and, if these recent PRs are obsolete, close and remake PRs based off what is currently in the repo?

ctiberious:
These are updates to the contracts.  I partly did them to take advantage of the capabilities of Contract Configurator 1.7.7 (like being able to natively round Duration and the fix to the "double()" command).  Partly I changed the way the contracts generate parameters.  In 1/LEO instead of selecting a random duration, it now generates a random number of orbits you need to complete (3 to 24) and converts that to a duration (it assumes 1 orbit is approx. 5400s).  It also uses the number of orbits to adjust the rewards (more orbits required = more reward).

NathanKell:
On the presumption that this is in fact safe to merge I am doing so--but please double-check all the contracts after merge, and please close and resubmit the one that is not able to be automatically merged. Cheers!

PR #332, opened 2015-11-10
Update Orbital 2 Repeat.cfg

ctiberious:


ctiberious:
When I first created this contract, Contract Configurator could not round a Duration variable.  So to get rounded durations I had to start off by generating a random duration.  Then recast that as a Double which I could then round.  Then finally recast that back to a Duration.  As of Contract Configurator 1.7.7 we can now natively round Duration so the main adjustment here is to use that capability.  Basically just cleaning up the process.

I also shortened the second duration so that the total flight is now 2-10 days.  And I set it up so that if you have a lower reputation you will be more likely to get shorter duration flights while players with higher reputation are more likely to get long duration flights.

PR #333, opened 2015-11-10
Update Orbital 3 Repeat.cfg

ctiberious:

### DIFF
--- a/GameData/RP-0/Contracts/Human Spaceflight/Orbital 3 Repeat.cfg
+++ b/GameData/RP-0/Contracts/Human Spaceflight/Orbital 3 Repeat.cfg
@@ -60,7 +60,7 @@ CONTRACT_TYPE
 	DATA
 	{
 		type = int
-		startApA = 300000 + Round(Random(0, 100000), 25000)
+		startApA = 300000 + Round(Random(0, 1000000), 25000)
 	}
 	DATA
 	{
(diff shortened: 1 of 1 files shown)


ctiberious:
When I first created this contract, Contract Configurator could not round a Duration variable. So to get rounded durations I had to start off by generating a random duration. Then recast that as a Double which I could then round. Then finally recast that back to a Duration. As of Contract Configurator 1.7.7 we can now natively round Duration so the main adjustment here is to use that capability. Basically just cleaning up the process.

I altered the duration so that it fits the capabilities of the Apollo CSM.  Now you get a random duration of 7-14 days.

Instead of a static minimum Pe, now you get a random Pe of 150km-200km.

The max Ap now randomly gives a range from 300km-1300km.

Finally, there is a maximum eccentricity that your orbit needs so even if your contract is minimum 150km by max 1300km, you can't just go for those extremes.  And the higher the players reputation, the lower the eccentricity needs to be.  So low rep players have plenty of leeway but high rep players need more precise orbits.

NathanKell:
As I mentioned on the other PR, I did merge your earlier changes, so now this conflicts. Can you close this and open a new one based off the current state of the repo? Unlike the tree changes, I don't feel comfortable attempting to do so myself as I don't know for sure what you do and don't want changed.

ctiberious:
Well I definitely spaced there.  I know I was going to submit changes but I didn't remember actually doing it because I was waiting for the contract configurator 1.7.7 update.  But I guess I actually did since all the changes are already in the main branch.

PR #334, opened 2015-11-10
Update Lunar Orbital 1 Repeat.cfg

ctiberious:
Just a minor update to change the title from "Moon" to "LLO".


ctiberious:
Should note, the main reason for this change is so I could use "1/Moon" for the moon landing contract.

NathanKell:
I would definitely suggest a floor, with some additional funds based on duration. The price difference between a 24hr and 36hr stay is miniscule compared to between no stay and 12hrs. :)

PR #335, opened 2015-11-10
Create Lunar Landing 1 Repeat.cfg

ctiberious:
New contract for manned landings on the moon.  I've setup this contract to randomly select a lunar biome so in addition to a time requirement, you also need to select an appropriate landing site rather than just being able to land anywhere.

NOTE: The rewards may be off.  I set them at a base 50k times the number of 12hr increments that you have to remain on the moon.  Everything looks right but on my first test the reward was significantly higher than anticipated.

### DIFF
--- a/GameData/RP-0/Contracts/Human Spaceflight/Lunar Landing 1 Repeat.cfg
+++ b/GameData/RP-0/Contracts/Human Spaceflight/Lunar Landing 1 Repeat.cfg
@@ -23,12 +23,12 @@ CONTRACT_TYPE
 	maxSimultaneous = 1
 	
 	// reward block
-	// (double(@/LandDur)/43200) should give a result of 1 - 6 based on number of 12hr increments
-	advanceFunds = 50000.0 * (double(@/LandDur)/43200)
-	rewardReputation = 30.0 * (double(@/LandDur)/43200)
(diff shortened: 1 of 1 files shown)


ctiberious:
So I'm running my 2nd Moon Landing contract in conjunction with my 2nd Low Lunar Orbit mission.  Both have the same 50k base reward but the landing contract I have setup to adjust that based on how many 12h increments you have to stay landed.  The LLO contract I received is offering 140k reward while the 2h landing contract (so 4x reward) is offering 562k.  So I think I need to adjust the multiplier down.  Maybe a 25% increase per 12hr increment?

NathanKell:
I would definitely suggest a floor, with some additional funds based on duration. The price difference between a 24hr and 36hr stay is miniscule compared to between no stay and 12hrs. :)

ctiberious:
I've adjusted the reward modifier.  Now we should get a multiplier of 1 - 2.25 resulting in a 12.5k boost in the base reward for each 12hr over the first.

NathanKell:
Sounds good!

Can you merge in from master and verify the contract is OK, so that it can automerge? Or make a fresh PR? Right now it's conflicting and unlike with trees I'm not sure which value to keep.

ctiberious:
Ok.  I've resubmitted this as #361.

PR #336, opened 2015-11-10
Update Avionics.cfg

ctiberious:
Added avionics for the FASA Atlas LV-3C and SLV-3C tanks.  I based the massLimit off of similarly sized Atlas-Agena tanks.

### DIFF
--- a/GameData/RP-0/Avionics.cfg
+++ b/GameData/RP-0/Avionics.cfg
@@ -416,6 +416,26 @@
 	}
 }
 
+// Atlas LV-3C (used as booster for early Atlas-Centaur)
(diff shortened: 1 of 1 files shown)


NathanKell:
Looks good to me.
@ctiberious as a general rule, can you give your commits (and your PRs) meaningful titles? "Update Avionics.cfg" doesn't tell us what it's doing, but "Add avionics for Atlas LV-3C/SLV-3C" does. Which you put in the description, which is great, but a useful title would be good too. :)

ctiberious:
I didn't even realize I could do that.  :)  I'll start doing it for all my future PRs.

PR #337, opened 2015-11-11
Space Shuttle - pricing

winged7:
With the numbers above and about 80% refund after landing you should get 1bln (USD 2011) per flight  = 130-140 mln (USD 1965)


NathanKell:
Heh. Those are big numbers. Merging, though we should remember to revisit when launch costs (as separate from part costs) get modeled.

PR #343, opened 2015-11-11
7K-LOK pricing

winged7:


NathanKell:
Tagging @Niemand303 in case he has some further sources to polish these, but merging. :)

PR #346, opened 2015-11-13
BE-4 pricing

winged7:


SirKeplan:
I think it's great to have the BE-4 in RO, but isn't it a bit premature to price something that hasn't even been built yet?

winged7:
it was placed (not by me) so this means that it should be priced also.

NathanKell:
Yeah, so long as it's documented in the tree that it's a guess, so we know to fix later, that seems fair enough.

PR #350, opened 2015-11-13
Manned Mars flyby - contract

winged7:


NathanKell:
Same two comments as the Venus contract #351

winged7:
Updated version with suggestion to use gravity assist.

NathanKell:
Looks good! Almost there, except please do also make the change from Manned to crewed in the title and filename as well as the description (indeed, any references whatsoever to manned should become crewed, and unmanned to uncrewed).

winged7:
Updated.

PR #351, opened 2015-11-13
Manned Venus flyby - contract

winged7:


SirKeplan:
Should this be Crewed, rather than Manned? applies to #350 too.

winged7:
Does it make any difference? (Sorry I'm not native english speaker)

NathanKell:
@SirKeplan is correct here. We very definitely use Crewed / Uncrewed rather than Manned/Unmanned to stay gender-neutral. :)

NathanKell:
Also, I note the description requires a gravity assist, but there's no parameter that actually requires it for completion. Should the description perhaps be altered to suggest, rather than demand, a gravity assist?

winged7:
Updated version with suggestion instead of demand.

NathanKell:
Looks good! Almost there, except please do also make the change from Manned to crewed in the title and filename as well as the description (indeed, any references whatsoever to manned should become crewed, and unmanned to uncrewed).

winged7:
Done :)

PR #352, opened 2015-11-14
Gemini - entryCostModifiers

winged7:
https://github.com/KSP-RO/RP-0/pull/323

### DIFF
--- a/GameData/RP-0/EntryCostModifiers.cfg
+++ b/GameData/RP-0/EntryCostModifiers.cfg
@@ -283,6 +283,19 @@ ENTRYCOSTMODS
 	{
 		name = FASAGeminiPod2
 		tag = heatshieldedLEO
+		entryCostMultipliers
+		{
+			FASAGeminiPod2White = 0.5
+		}
+	}
+	PART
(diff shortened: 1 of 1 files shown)


NathanKell:
The top PART { name = FASAGeminiPod2 } will collide with the already existing node of that name.

Please do actually test these things. ;)

winged7:
No idea how to make it better.

NathanKell:
Ok. So please then _ask_ how to do something if you don't know, rather than just submitting a PR that won't work. :] And we'll try to figure out how to make it work, together.

In this case there's a bit of a problem, because we want to remove only part of the entrycost, but there's two classes of subtractors as it were--the heatshields, and the pod. I think, sadly, it's probably best to use multipliers for the pod-pod interaction. i.e. add an entryCostMultiplers node to the original pod PART block in the cfg and set the White version to 0.5, and the same block to the white version, linking the FASAGeminiPod2 at 0.5.

winged7:
Something like that?

NathanKell:
Yep, exactly!

PR #353, opened 2015-11-15
Stock mk1, mk2, mk3, B9 shielding  - entryCostModifiers

winged7:


NathanKell:
Ah, yep! Just like that :)

Oooh, that makes me wonder however--should the cockpit cost (and thus this cost) be higher? And if so, maybe the shielding cost should be more like 1/2 not 1/3 ?

Maybe not change the per-item cost, but definitely the entry costs. I find it hard to imagine that the Mk2 system (i.e. a space shuttle system circa 1970, i.e. midway between DynaSoar and STS) would be so much cheaper than Mercury. For Mercury, the entryCost is ~100000 funds (100mil in 1965 dollars), with 16000 of those funds being for the shielding system. Hear, the shielding system is only 4333.

winged7:
I switched from 1/3 of the unit cost (4333) to 1/3 of the development cost (43333 - 10x bigger) however 130k for development cost is still a bit cheap compared to Mercury so maybe multiply increase it 2x to 260k?

winged7:
i can delete mk2, mk3, S2, S3 etc. development cost value from tech tree so it will be 20x more expensive than unit cost by default  and thus 2x more expensive than now.

NathanKell:
Moving discussion to #237 til we get this figured out. :)

NathanKell:
<3

PR #360, opened 2015-11-15
Avionics for many new parts

winged7:
-stock mk1, mk2, mk3
-B9
-FASA
-AIES
-DragonV2
-Orion
-Space Shuttle

### DIFF
--- a/GameData/RP-0/Avionics.cfg
+++ b/GameData/RP-0/Avionics.cfg
@@ -30,8 +30,17 @@
 	}
 }
 
+@PART[HL_Aero_Cockpit|B9_Cockpit_MK5|mk3Cockpit_Shuttle|STSOrbiter|CSSCrewCompartment]:FOR[RP-0]
+{
+	MODULE
+	{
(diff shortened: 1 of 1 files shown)


NathanKell:
I'd actually like to keep avionics off airplane cockpits, unless they are the sort of thing where there actually is an avionics limit. I.e. you could use the He 111 cockpit for anything sized between a light twin and a B-36 and as long as there's a person in it pulling the stick it would be fine. So the Lightning cockpit, an avionics limit makes sense (you can't really fly a C-5 with an F-22's avionics) but for old-skool stuff it does not.

The other stuff is fine though! :)

winged7:
ok, I delted avionics from Bonanza, He-111 and Ju-87. 

What about mk1, mk2, mk3 - shuttle parts? also delete?

NathanKell:
I think for the Mk1/2/3 avionics does make sense--it's controlling a spacecraft with rocket engines, in those cases, so an avionics limit makes sense.

PR #363, opened 2015-11-17
placement of some rn parts & fix typo

batuhantasli:
Only placements, no clue about prices. I thought salyut/soyuz left and right solar panels shouldn't have seperate development costs, so changed that. Also placed proton interstage adapters and soyuz astp.


NathanKell:
This is erroring out, that means there's a typo in the tree file now. It thinks line 1947? Please fix (and test) locally and update the PR?
Thanks!

batuhantasli:
Sorry about that. It should be ok now.

Issue #266, opened 2015-10-11
More science values


glilienthal:
Hi all, 
There are a lot of science values not yet implemented. What do you think of my proposal? - Thought I get some feedback before doing the pull request. 

G. 
# proposed Science values (not yet implemented)

| _Body_ | _landed_ | _splash_ | _flyLow_ | _flyHigh_ | _Spacelow_ | _SpaceHigh_ | _recovery_ |
| --- | --- | --- | --- | --- | --- | --- | --- |
| _Jupiter_ |  |  | 24 | 15 | 10.0 | 6.0 | 8.0 |
| Io | 30 |  |  |  | 15.0 | 8.0 | 10.0 |
| Europa | 30 |  |  |  | 15.0 | 8.0 | 10.0 |
| Ganymede | 30 |  |  |  | 15.0 | 8.0 | 10.0 |
| Callisto | 30 |  |  |  | 15.0 | 8.0 | 10.0 |
| _Saturn_ |  |  | 26 | 17 | 12.0 | 8.0 | 10.0 |
| Titan | 35 |  | 26 | 17 | 15.0 | 10.0 | 15.0 |
| _Uranus_ |  |  | 28 | 20 | 14.0 | 10.0 | 12.0 |
| _Neptun_ |  |  | 29 | 21 | 15.0 | 11.0 | 13.0 |
| Pluto | 40 |  |  |  | 18.0 | 14.0 | 20.0 |
| Mercury | 40 |  |  |  | 18.0 | 14.0 | 20.0 |

NathanKell:
That looks good to me!

pap1723:
RP-1 Science Overhaul

PR #382, opened 2015-11-21
Update BodyScienceParams.cfg

winged7:
Based on: 
https://github.com/KSP-RO/RP-0/issues/266
https://github.com/KSP-RO/RP-0/wiki/Science


glilienthal:
Oh. Have just seen that. Nice. :)

I think I might have overdone it a little bit. Might be too much Science.

PR #387, opened 2015-11-22
RN&Tantares R7 re-placement

batuhantasli:
Re-placed RN and Tantares R7 tanks as well as RN sputnik fairings etc to the Basic Construction node. Mercury - Atlas LV was already there so it should be ok now with the avionics restriction.

### DIFF
--- a/GameData/RP-0/Avionics.cfg
+++ b/GameData/RP-0/Avionics.cfg
@@ -407,6 +407,7 @@
 	{
 		name = ModuleAvionics
 		massLimit = 120.0
+		techRequired = basicAvionics
 	}
 }
 // Atlas E/F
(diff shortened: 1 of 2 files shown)


NathanKell:
Can you also tech-restrict the avionics being assigned?

batuhantasli:
@NathanKell sorry, how do I do that?

NathanKell:
Whoops, missed this.
When you add the avionics module, i.e.
```
@PART[foo]:FOR[RP-0]
{
	MODULE
	{
		name = ModuleAvionics
		massLimit = 5.0
		techRequired = stability // or whatever
	}
}
```

NathanKell:
(the avionics are presumably added in Avionics.cfg -- if the add is for multiple parts, break it out into single patches, one per part, so you can set individual tech requireds)

PR #391, opened 2015-11-27
new parts

winged7:
- probe parachute from FASA
- MB-3-3 engine from FASA
- spacecraft LM adapter from FASA
- airbrake (stock)
- avionicsnoseCone (stock)
- fuel line (stock)
- electric propeller engine from KAX
- heavy landing gear from KAX
- STAR 17 from KK_ATK pack
- GPHS-RTG from SXT
- GPHS-RTG shielded from SXT

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -1393,7 +1393,7 @@ aerodynamicSystems: # X-15 parts, Mach 2 engines
         cost: 10
     B9_Utility_Landing_Gear_HDG2AL:
         cost: 10
-    KAXheavygear
+    KAXheavygear:
         cost: 15
         
     #XLR99
--- a/tree.yml
+++ b/tree.yml
@@ -592,6 +592,8 @@ start: #TL-1 early tech. Sounding rockets, basic structural pieces and wings. La
         cost: 20 #guess, balanced against 2800.
     TwinWasp-R1830-94: # Prop Engine R-1830 had 173,618 made across all variants Date:1930-1951
         cost: 15 # guess, balanced against R1820.
+    KAXelectricprop:
+        cost: 10 # electric propeller engine, 300hp
     
     # source for some engine prices http://www.warbirdsandairshows.com/Aircraft%20manufacturing/enginemanufacturewwii.htm
     #V1710 - 33.3 before obsolete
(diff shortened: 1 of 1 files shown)


stratochief66:
Star 17 looks good, as does the FASA LM adapter.

Some of the parts might warrant some discussion about the role we want them to play in RP-0 (ie. the fuel line, I feel like that part could be easily abused. I'm unsure how it could be changed to perform well as an RP-0 part)

Some parts might need a bit of description about form and function since I am unfamiliar with them. For instance, is the KAX landing gear some gear for planes?

Perhaps in your PR notes, if you could describe the part a bit, and what tech era and node you are proposing placing it. It might also be worth stating whether you are proposing initial placement of the part, of if it is being moved from one to another.

I know much of that information can be gleaned from careful analysis of the PR, but if you could put some of it upfront in the PR notes, it would make PRs a lot easier to review, and it would be very much appreciated :)

The other existing parachutes (other than specific craft FASA ones) appear to open on Start. Currently, only heatshields appear in Survivability and the FASA Mini Parachute shows in 'Miniaturization' (I think due to it's non-RP-0 tech placement?) and that might not be a bad place to stick it officially. Thoughts? Should it be in Start, or Survivability, or Miniaturization since it might be the smallest chute? Is there a way for us to differentiate the abilities/limits of various chutes as we can with heat shields?

@NathanKell or someone should look at the plane parts, as I have no basis for thinking where they belong and how they should be judged.

NathanKell:
MB-3-3 is exactly the LR79-NA-13. It should be placed in the same node that unlocks in, and cost (LR79 base cost + CONFIG cost of LR79-NA-13). Its entrycost should be, likewise, the entrycost of the LR79 + entrycost of the LR79-NA-13 CONFIG.

All parachutes should go in Start, IMO.

The fuel line has unlimited flow rate, and thus allows better-than-present-technology crossfeeding. For that reason it should not unlock so early, IMO. On the other hand, now that decouplers can be set to allow crossfeeding from the start, that ship has sailed, so I guess that placement is workable for now.

NathanKell:
Airplane related parts look fine.

winged7:
I've added also GPHS-RTG from SXT.

PR #392, opened 2015-11-29
Tech tree to suit the new antenna stats.

Schnobs:
Generally, antennas will be available as soon as one has the means to power them; in most cases that's a bit earlier than in real-life history.

BEWARE: this makes the AsteroidDay mod mandatory for RP-0, as I used it's antenna for the mariner model.

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -1302,6 +1302,8 @@ stability:
     FASASolarMini:
         cost: 75
 
+    RTShortDish2: # early small dish, CommGEO
+        cost: 400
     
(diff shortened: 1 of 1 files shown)


NathanKell:
Is the RTG meant to be costed?

Schnobs:
Don't know what you think, but the rtg is in the same node as before, at the same price.
I only moved it up a bit so the "Radio and plasma wave antennae" comment is near the part(s) it refers to. The rtg seems to have slipped in there during an earlier edit.

PR #397, opened 2015-12-04
Update tech placement for Atlas and Centaur

ctiberious:
Atlas G/H/I didn't start being used until the 80s so I've moved the Atlas H tank from Specialized Construction over to Advanced Metal Works.  Though it may actually be more accurate to wait until Nanolathing.

The Centaur-T was effectively designed specifically for the Titan IV so it makes sense to have it be introduced in the same tier as the 11A engines used on the TItan IV.  Those engines become available with Advanced Rocketry so I've put the tank in Nanolathing.

Centaur D2 doesn't get introduced until early 90s when the Atlas II starts being used.  Since the engines for the Atlas II don't become available until Experimental Rocketry, I've moved the Centaur D2 tank to Exotic Alloys to keep it in the same Tier and in a construction node.

I've moved the Centaur D3 (used on the Atlas III) over to Orbital Assembly to keep it in a construction node.  Also because the Atlas III didn't start to enter service until early 2000.  I did notice that the RD-180 engine is currently available at Refined Staged Combustion which I'd guess is closer to the 90s but that may be because the engine was used in Russia long before the first Atlas III was put into active service.

Finally, I've moved the Centaur D5 (used on Atlas V) over to Orbital Megastructures, again to keep it in a construction node.

I know we have the initial Centaur tank becomes available with the engine in Early Hydrolox Engines.  When I initially placed the other Centaur tanks, I kept them in Hydrolox tech nodes because of how the first was setup.  It makes a certain kind of sense that the first would be introduced with the engines but I figured the subsequent tanks were better placed in "tank" nodes.  Since it looks like most tanks are in construction nodes, I've tried to place the later era Centaur tanks into later tier nodes that seem to follow the Construction path.

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -2708,7 +2708,6 @@ hydroloxTL4:
     RO_KVD1: #engine was tested in early 1970s as the RD-56, never flown as later N1 designs where cancelled
         cost: 650
         entryCost: 27750
-    FASAGeminiLFTCentarCSM_T: # Centaur D5 for Atlas V
         
 precisionPropulsion:
     # Bobcat's Soviet Engines -Need Costing-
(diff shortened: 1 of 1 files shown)


stratochief66:
Agree with the tentative placing of the Atlas H, since we haven't refined the tech tree out to that era yet.

Same for the Centaur T and Centaur D3, etc.

This seems to make sense for starting to get 'future' (80s+) hardware lined up right.

PR #398, opened 2015-12-04
Added Bell Configs

stratochief66:
A few more configs were added in RO without changes here. Set estimated price increases, as without these changes the new configs unlock at the same price and tech as the base one.

If someone wants to research these engine configs, they might also belong at a higher tech node unlock, although they quickly become obsolete (in my opinion) as early hydrolox is shortly unlocked afterward.

### DIFF
--- a/GameData/RP-0/FASA_Engines.cfg
+++ b/GameData/RP-0/FASA_Engines.cfg
@@ -115,13 +115,13 @@
 		}
 		@CONFIG[Bell-8096]
 		{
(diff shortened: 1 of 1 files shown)


NathanKell:
Thanks!

I _think_ these should be requiring advRocketry for the 96 and, um, TL3 or TL4 rocketry (heavy or veryHeavy Rocketry, IIRC?) for the -39. They're later configs than the B model that unlocks in basicConstruction

PR #405, opened 2015-12-09
Added AsteroidDay as dependency

Schnobs:
The recent changes to antenna stats use that mod's antenna as *the* mariner item.


NathanKell:
Sorry, that file is confusing and shouldn't even exist. Fixing.

PR #408, opened 2015-12-10
small FASA adds

stratochief66:
Added FASAGeminiLFTLongIF tank (for Titan 3A, 3B), priced a couple percent more than Titan II tank. 200


stratochief66:
Don't close, pending decision on whether Transtage (FASAGeminiLFECentarTwin) should be moved to MatureOrbital.

stratochief66:
Ready to go.

PR #410, opened 2015-12-12
Place and price Cryoengines mod

SirKeplan:
Something seems wrong with the RL10B-2 pricing, either it's me or it's the pricing of XXxAres1J2-XHIGH, where does the 3500 figure come from?

### DIFF
--- a/GameData/RP-0/CryoEngines_Engines.cfg
+++ b/GameData/RP-0/CryoEngines_Engines.cfg
@@ -1,7 +1,7 @@
 //RS-68
 @PART[cryoengine-375-1]:AFTER[RealismOverhaul]:NEEDS[!RftS,!RealFuels_StockEngines]
 {
-	@MODULE[ModuleEngineConfigs]
+	@MODULE[ModuleEngineConfigs],1
 	{
 		%configuration = RS-68
 		@CONFIG[RS-68A]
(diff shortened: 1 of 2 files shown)


NathanKell:
Should be the base cost of the RL10, plus the cost of the RL10B-2 CONFIG. 3500 doesn't sound that unreasonable.

SirKeplan:
sorry, should have been clearer, it looks like there is a previously priced RL10B-2 at cost: 5300  (1800 + 3500) 
i was wondering why the figure 3500 for the config, the RL10B-2 config costs 2000. so i was going to have the standalone RL10B-2 at 3800 cost.

SirKeplan:
Ah that explains it! some config files seem to have different prices for the RL10, in RO the prices are lower than when priced in RP-0, and i was only looking at the FASA RL10 as a guide.

Should all the pricing be moved into RP-0 so there isn't confusion?

NathanKell:
Yeah, I think that probably makes sense. I had them in RO to lower work, but they don't make much sense in RO, I agree

SirKeplan:
i should be able to do that, if i can figure out which pricing is *correct* :)

NathanKell:
Lemme know when this is good to go! :)

SirKeplan:
done i hope. now depends in pull #759 from RO though
https://github.com/KSP-RO/RealismOverhaul/pull/759

SirKeplan:
Well, now it's done.

NathanKell:
Um, you're running on AFTER[RealismOverhaul] not FOR[RP-0] -- as was discussed in IRC no _particular_ advantage in this case, except uniformity, but that seems worth it. :)

PR #412, opened 2015-12-12
Migrate RL10 Configs to RP-0

SirKeplan:
there will be a RO PR to go with this soon

Don't Merge yet, needs checking/testing.

### DIFF
--- a/GameData/RP-0/FASA_Engines.cfg
+++ b/GameData/RP-0/FASA_Engines.cfg
@@ -156,7 +156,7 @@
 }
 
 // RL10
-@PART[FASAApolloLFERL10]:FOR[RealismOverhaul]
+@PART[FASAApolloLFERL10]:FOR[RP-0]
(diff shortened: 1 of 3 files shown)


SirKeplan:
Shouldn't we start with configuration = RL10A-1 instead of the RL10A-3-3 ?

SirKeplan:
should be good to go.

NathanKell:
@SirKeplan  Oh, missed the question. Yes, we should start with A-1, and I thought I had a patch to do just that. >.>
RO starts with A3-3 because that's the most famous/generally-used model, but RP-0 should start with A-1. Though as you say it'll automatically bounce back to the first available.

SirKeplan:
when I tested it actually did not bounce back to the first available, you can actually use a locked config if it is the default one.

NathanKell:
That sounds like an RF bug...

Issue #417, opened 2015-12-17
Later American placement, ctiberious


stratochief66:
@ctiberious 

You can post questions or notes on your progress here if you like. It might also be good to post a statement of your intended goal :)

ctiberious:
I am going to be working up a timeline for the American Atlas, Thor/Delta, and Titan rockets.  Basically from inception to their modern versions.  I plan on including the more well know (to me at least) 2nd/3rd stages for these rockets as well (i.e., Able, Agena, Centaur, Transtage, etc).  I know a lot of the earlier versions of these rockets have already been correctly placed, but I'd like to work on getting the later and modern era versions placed correctly.  Once I have the initial time line created, I'll see about getting it posted here for comments.  After that, I'll work on actually moving engines, configs and tanks into the correct nodes.

ctiberious:
Still working up the timeline but should have something available tonight or tomorrow.  In the mean time I've come across a couple things.  Much of this is just notes I wanted to stick up here but some are questions I hope someone might be able to comment about.

1) We don't really have the GRC X-242 SRM that was used on the Vanguard.  Closest option we have is the Altair which isn't available until a later tech node.
2) We've got an MB-3-3 engine which is just the other designation for the LR-79-13 that was used on early Thor/Delta rockets.  That's cool but the MB-3-3 part we have has a few issues.  For one, it's too large to fit on the 2.44m dia Thor tank.  It also doesn't look like it's got the right thrust or ISP.  And finally, it doesn't have any other configurations.  Since the RS-27 and RS-27A are basically just upgrades of the MB-3-3, I'd think we should include those configurations.
3) Looks like I didn't realize there was a non-FASA version of the LR-105 and LR-89 atlas engines.  I think those need to have the configs updated like the FASA versions have.
4) There appears to be two sets of engines on the LR91 series Titan engine.  One set has configurations and a single ignition available.  The other had 5 ignitions and an "Engine" button but when you press it, nothing comes up.  I'm not sure what this second engine is for but it seems to sometimes interfere with the configurable engine.
5) The transtage we currently have has 490L of available volume remaining by default.  Was there multiple versions of this stage?
6) We have an "AJ10 (Early)" that has the -37, -42, -142 and -118D configs.  And we have an "AJ10 (Mid)" that has the -104D and 118E configs.  Based on what I'm finding, the -37, -42, -142 and -104D should probably all be part of the Early version and all the -118 configs should probably be part of the Mid part.
[...]

PR #422, opened 2016-01-06
Update Avionics for issue #417

ctiberious:
These are the avionics adjustments discussed in issue #417.  The changes include:
1- Modify the avionics for the Centaur D3 & D5 to reflect the integrated avionics on those two vehicles.
2- Added avionics for the new FASA Atlas II tank.
3- Added avionics for all five of the new FASA Thor/Delta avionics parts.


ctiberious:
I've added tech requirements to the avionics module for the Titan 2nd stages, Atlas tanks and Centaur tanks.

PR #424, opened 2016-01-06
Update tree.yml for issue #417

ctiberious:
These are the bulk of the tech node changes discussed in issue #417.  In a few places I've also altered Cost and/or EntryCost so that duplicate parts match (i.e. SRBs from ATK have the same costs as their KW Rocketry equivalent).

Also included are tech node placement for the new parts discussed in issue #417.  This includes the AJ10Adv (118F/118K), the Atlas II tank, and the Thor/Delta avionics parts.

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -990,6 +990,8 @@ generalRocketry: # TL1 Late 1950s/1960 tech, first real orbital LVs. Atlas, R-7,
 
     liquidEngine: *LR105
 
+    FASAGeminiLR91: *LR91
+
 # Black Arrow engines (derivatives of Black Knight engine, so can be early,
 # despite Black Arrow being quite late).
 # see Mature Solids for Waxwing.
@@ -1104,6 +1106,10 @@ basicConstruction:
         
     FASATitanLR91Dec: # Aerojet LR-91 Decoupler
         cost: 75
+    FASAGeminiLFTLong:
+        cost: 187 # same as proc
+    FASAGeminiLFTMed:
+        cost: 250
     # FIXME other Titan II parts go here, with costing.
     FASAGeminiDecDark25_Titan3m:
         cost: 75
@@ -1172,6 +1178,10 @@ basicAvionics:
         entryCost: 12000
         cost: 400
 
+    FASADeltaAv0:
+        entryCost: 2500
+        cost: 100
+
     # Improved Sputniks
     sputnik2:
         cost: 100
@@ -1197,16 +1207,14 @@ basicAvionics:
 
 
 basicSolids:
-    # Castor I (Sergeant / M33) solid motor
-    solidBooster1-1Small: &CastorI
-        cost: 300 # nautix 1.01m assume 1985 dollars
-    FASADeltaCastorSrb: *CastorI
-
     # Add Polaris and Scout parts here
     
     RO-X-248: #late-Vanguard/Able/Delta kick motor
         cost: 170 # nautix says .6mil, but doesn't say what-year dollars, assume 1985.
     
+    RO-X-258: # Altair II
+        cost: 200 # 700k in 1985, nautix
+        
     # moved Proc SRB to here
     proceduralSRBRealFuels:
         cost: 0.1
@@ -1550,12 +1558,12 @@ advRocketry:
         cost: 1800
     FASA_SIB:
         cost: 1800
-    FASAAgenaLFT: # Agena B/D tankage
-        cost: 50 # TODO compare to proc part
+    FASAGeminiLFECentarTwin: # Transtage
+        cost: 4300
+        
     # Titan II upper stage
     liquidEngineorbit2: &LR91
         cost: 600
-    FASAGeminiLR91: *LR91
     #Europa launcher
     NLV_Engine_A:
 
@@ -1630,12 +1638,6 @@ matureSolids:
         cost: 2000 # TODO we *should* have references on this from BellComm studies
     # on Saturn IB-based INTs (as to how much the M55 really did cost in serial production.)
     
-    RO-X-258: # Altair II
-        cost: 200 # 700k in 1985, nautix
-        
-    RT2: # Castor II
-        cost: 320 # unstated, should be about Castor I's cost.
-        
     SXTCastor30: &Star17 # Star 17, used on Atlas Agena D
         cost: 172 # nautix 580k assume 1985 dollars
     KK_ATK_Star17A: *Star17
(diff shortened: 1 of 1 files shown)


SirKeplan:
Using the YAML node marker syntax(&partid, *partid etc) allows you to avoid duplicating the cost and entryCost fields. i think you are getting build errors because you are misusing that somewhere.

ctiberious:
I'll figure out what is causing this to fail.

ctiberious:
I think you're right.  I'm moving a lot of things around and since I'm not completely familiar with the YAML format, I've just been cutting&pasting.  I'll go through the changes I'm making and look for &partid, *partid syntax to make sure I have things is the correct order.

SirKeplan:
what's the rationale for moving the RS-68 into giganticRocketry?
i think we've established it's TL6 not TL7(the RS-68A variant is TL7 though) and AFAIK all hydrolox engines belong in hydrolox nodes.

SirKeplan:
>YAML::XS::Load Error: The problem:
>
>    did not find expected key
>
>was found at document: 1, line: 1654, column: 9`

clearly the `CastorI` that's causing your problem.

ctiberious:
I missed the fact that the RS-68 is a Hydrolox engine which is why I put it in the node I did.  That's my mistake and I'll correct it.
I proposed putting the RS-68 into giganticRocketry (TL7) to keep the Delta IV from being available prior to it's first in 2002.  Unless I missed it, no one commented about the placement in the #417 discussion (other than @NathanKell pointing out the best mod for that engine) so I assumed there were no complaints about it's placement.  With the addition of the DCSS4 and DCSS5 avionics parts, it's less of a concern having the RS-68 turn up in TL6 (even though 1998 is pretty much the end of that TL) so if that's the preference, I can move it accordingly.

As for the CastorI thing, it looks like I had an extra space that I wasn't seeing.  So it's finally passed that check.

SirKeplan:
the RS-68 might be getting near the edge, but i'm pretty sure @NathanKell said early 2000s is TL6. that lets the RS-68A upgrade become TL7.

ctiberious:
I'm just doing a quick search here but the wiki article I'm reading about the RS-68 states a couple things:
1) The RS-68 was specifically designed for the Delta IV
2) The first successful test firing happened on September 11, 1998.
3) The engine wasn't certified for use on the Delta IV until December 2001
4) The engine's first official flight wasn't until November 20, 2002

In the #417 discussion we agreed that the Atlas V became available in TL7.  The Atlas V and Delta IV both made their first flights in 2002 (August 21 and November 20, respectively).  So if the Atlas V is TL7, Delta IV should also be TL7.  To my mind this means TL7 is at least 2002 on (until we have a TL8 anyway).  Since the RS-68 was specifically designed for the Delta IV and didn't make it's first flight until 2002, to me that implies it should be in TL7.

As for the RS-68A configuration, it's first flight wasn't until June 29, 2012.  For the time being that puts it in TL7 but I'd say once we have a set of TL8 nodes, that's where it should ultimately go.

NathanKell:
In this case I would suggest erring on the side of gameplay and using TL6 and TL7. We can reexamine when we deal with TL8. (There certainly wasn't anything preventing a somewhat earlier RS-68; it grew out of STME tech anyway....)

ctiberious:
I'll move the RS-68 to TL6 if you think it's really required, but considering it didn't fly until TL7, it really makes no sense to me to have it show up early.  Yes, it is based on the SSME but that doesn't change the fact that it didn't fly until the end of 2002.

SirKeplan:
@ctiberious the techtree isn't exactly a timeline, it's meant to be a progression tree. fitting it to real life engines make it look like a timeline, and maybe it's become rather timeline like, but i don't think that's the intent of the tree.

ALSO, NathanKell said STME, not  SSME. STME was an engine for the Nation Launch system rocket, a family of lifters that were cancelled in the early 90s. the STME  tech was recycled into the RS-68.

ctiberious:
Progression or timeline, the point is the same.  We've already agreed that the Atlas V is a TL7 rocket.  It flew for the first time two-three months before the Delta IV.  So, logically, if the Atlas V is a TL7 rocket, then the Delta IV is also a TL7 rocket.  We've also agreed that the RD-180 used on the Atlas V is a TL6 engine because it is based on the RD-170 which is a TL5 engine.  But unlike the RD-180, the RS-68 was purpose built specifically for the Delta IV.  It wasn't simply a cut down version of an engine that was already flying.  To me, logic says that if the RS-68 was designed specifically for a TL7 rocket, and if it was never used until that TL7 rocket first flew, then it should be a TL7 engine.

I'm all for making game play over realism decisions where they are appropriate.  After all, I am the one who suggested that the Titan IV should come a bit earlier then it really did because it would have if not for the space shuttle program.  But development of the RS-68 (at least from what I can piece together) was not held up because of the space shuttle.  Therefore I can't see any reason the engine should appear early.  But I don't make the final decision here and if Nathan definitely feels the RS-68 should show up earlier then it did, I'll make the correction so that this update can be merged.  But I really don't think it should show up before TL7.

NathanKell:
Thing is, it does not matter when the rocket was introduced. Example: Black Arrow. First flew in the 1970s. The engines were 1950s HTP technology, and thus they sure as heck aren't in a TL4 engine node. Those engines weren't tested until the late 60s, it's true, but they were reworks from Black Knight, and its own Gamma 301 was just a small Stentor.

RS-68 itself didn't fly until Delta IV, as you say, but it's very much STME reworked (in the same way we have a 200klbf J-2 available early, and the LR87-LH2, though J-2 itself wasn't put to use until TL3 in reality). That's the comparison I'm seeing here.

Another way of looking at this was raised on the thread, actually, one we've kinda been ignoring (I'm very guilty of this myself of course, I'm not pointing fingers). And that is that the historical record can't be so guiding for us, because unless we want RP-0 to model the Vietnam War, 70s retrenchment and the Brezhnev stagnation, horrific NASA mis-steps, and so forth, the historical record post-1970 simply does not make for reasonable progression. This is why, for example, I felt fine placing the Merlin earlier, and why I think the RS-68 fits fine as a TL6 engine, because *even with our own history's reduced emphasis on spaceflight* it was built with that level of technology.

(This is also, of course, an argument for bringing RD-180 and RD-191 down to TL5, maybe...)

Does that make sense?

ctiberious:
I'm not a rocket scientist so I can't tell you how complicated it is to take one engine and advance it.  SirKeplan says the STME engine was shelved in the 90s.  If it hadn't been, would we have had the RS-68 sooner?  Maybe.  Maybe not.  But where do you do draw the line?  We've got about a dozen different configs for the AJ10 and they show up in something like five different TLs.  But they are all pretty much based on the original Vanguard engine.  So would we be justified in having all of them show up in TL1 or TL2 even though some variations didn't really turn up until TL5?  And what changes should determine when something progresses to a more advanced TL?  Right now we have a pretty straight forward justification system for explaining why a particular part shows up in a particular node.  If it was first used in year X, it goes into TL Y.  Every time you change that, you have to constantly explain the justification for the change.  Sure, RS-68 could have shown up sooner than when it actually did but the same could be said of most of the engines.  My opinion is that we keep it simple and have parts show up in TLs that correspond to particular eras except were there is a compelling reason to do things differently.

That said, this is all just my opinion.  Since it seems like you guys really want the RS-68 to show up in TL6, I'll update this PR accordingly once I'm back at my main computer.

PR #425, opened 2016-01-06
Fixed "Have Part on Vessel" bug

ntimpe:
I changed the "Title" section under PART_REQUEST to "Article/PartDescription/VesselDescription" which is the new syntax for 1.0.5. This should fix the bug that had contracts asking for "Part on Vessel." This is my first edit, so I hope I didn't screw up :)


NathanKell:
You rock! Thanks! :)

PR #426, opened 2016-01-07
Added more results for CrewReport.cfg

ntimpe:
I added some more answers for the crew reports and fixed a typo or two.


NathanKell:
Thanks! :)

PR #431, opened 2016-01-19
Delta LTTAT avionics upgrade

ctiberious:
Slight increase to allow the avionics part designed for the LTTAT to function on the Delta 0900 rocket.

### DIFF
--- a/GameData/RP-0/Avionics.cfg
+++ b/GameData/RP-0/Avionics.cfg
@@ -640,7 +640,7 @@
 	MODULE
 	{
 		name = ModuleAvionics
-		massLimit = 110.0
+		massLimit = 115.0
 	}
 }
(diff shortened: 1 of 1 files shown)


NathanKell:
I'd suggest just bringing this up to 130 (was 120), i.e. in line with the standard 2m unit.

ctiberious:
Isn't the early 2m unit a 120?

NathanKell:
9423a221864ff5429d346dfe5e1c6e5b35137859 :)

ctiberious:
Ah.  It's about to be updated.  Gotcha.  130t is probably a bit overkill for the Thor since it had a much lower payload than the TItan and Atlas, but I've updated that avionics package all the same.  :)

PR #441, opened 2016-02-15
Standardizing RL-10 config unlock nodes

ntimpe:
The config unlock nodes for the RL-10s from FASA and KW were different, so I changed these to mirror the ones KW has.


NathanKell:
Sadly the other way around is correct. Or rather, the nodes in the Squad one are correct (they may be set in RO not RP-0 at the moment). With the current changeover to the new engine config system in RO, this kind of duplication can finally die (you can see the problems it's caused!)

PR #449, opened 2016-02-26
probeCoreCube - cost revision

winged7:
Not sure why it was more expensive than even Apollo CM. Now 15x cheaper - similar to AIES satelite buses


NathanKell:
This makes it cheaper than the Surveyor core. Instead, I am fairly sure the price for the Apollo CM is wrong.

PR #452, opened 2016-02-26
another RD-108 from VSR

winged7:


NathanKell:
This is from Ven's ;)

PR #457, opened 2016-02-27
adapterEngines (stock Mk3 part)

winged7:


NathanKell:
It may be structural, but its price will depend on the prices of other Mk3 parts due to the heat shielding. For now please leave it unpriced like the other Mk3 parts, or work on the Mk3 pricing issue. :)

PR #460, opened 2016-02-27
VSR cargobays

winged7:
https://github.com/KSP-RO/RealismOverhaul/pull/989/files


NathanKell:
Those prices seem a bit low. I'm going to merge in, but then push them up some--I doubt a cargo bay on a 2m-diameter spaceplane would only cost, say, less than a WAC Corporal (and a 1.25m one cost 1/4 a single WAC round)

PR #464, opened 2016-03-01
Nuclear engines - minor changes

winged7:


winged7:
Price for the NERVA from Chaka Monkey was shrinked down from 25000 to 8200 because the engine was nerfed in one of the latest RO releases: from 333kN to only 67kN, Also BNTR now provides 111kN instead of 67kN of thrust so the price was adjusted.

NathanKell:
Ok. And does this have any backup from the costs in the NERVA reports?

winged7:
Sorry, I don't understand phrase "backup from the costs" :/ Could you explain this with simpler words?

SirKeplan:
@winged7  i think NathanKell is asking if this matches any data in the Offical NERVA documents/reports.

winged7:
That's exactly what I was thinking :) The problem is that currently the price for the stock NERVA (333kN) is just guessed or at least there is no source mentioned.

winged7:
i've found only one number: 226 mln USD (1970 -1980? USD) - this is unit cost for the space tug with 867kN nuclear engine.
http://www.astronautix.com/craft/nerva.htm

Let's assume that engines are worth half of that - 113mln. 333/867 * 113 = 43,4 mln 

43,4 mln (1980 USD) = 16,6 mln (1965 USD)
43,4 mln (1970 USD) = 35,23 mln (1965 USD) 

I think it will be good to keep them without any changes so 25 mln USD per one.

winged7:
I've also added PR to entry cost modifiers so you will pay only once for the development cost.

NathanKell:
Awesome!

NathanKell:
Ah, one thing--I would expect the majority of the cost of a NTR to be in the N part, not the TR part, so I wouldn't think lower-thrust models would be all that much cheaper. Not so?

winged7:
I adjusted prices for the lower-thrust models.

NathanKell:
...I asked a question. I don't know the answer either. I had hoped that you, having read those source documents, might have an answer...

stratochief66:
Delivering more thrust calls for a larger reactor space with more nuclear fuel, etc. So it is sensible to scale the price linearly with the thrust. 

An engine designed for more reliable relights, or to also generate electricity would have a whole different cost function, but just straight NTRs would scale fairly linearly between cost and thrust.

We will have a better idea of the function for extrapolating once we have a few more data points, but I think this is a good starting point with some researched backing.

NathanKell:
Ok.

PR #465, opened 2016-03-01
New parts

winged7:
- GEM 63
- GEM 63 XL
- Soyuz orbital module from Habitat pack
- Soyuz heatshield (stock)
- stock APASport


NathanKell:
There's already a Soyuz heat shield, the new one should be reference to that. Same with the APAS port. Rest looks reasonable.

winged7:
Price for the APASport is the same as for the other APAS portas (there are two of them).

As to the Soyuz heatshield - I can't find other heatshield so deleted for now.

NathanKell:
Ya, it wasn't about the price, it was about using the reference operators. :)

winged7:
Done.

PR #466, opened 2016-03-01
AIES command pods

winged7:
Most of these parts should be placed in much later nodes because they are from 1980+ era.


NathanKell:
Agree. Manually merging.

PR #473, opened 2016-03-02
RO H-1/RS-27 P&P

PhineasFreak:
* Place and price the RO copy of H-1. Cost values taken from the "entryCostModifiers" of this engine but i do not know if this is the correct thing to do.


NathanKell:
Uh...this is already done. ;)
https://github.com/KSP-RO/RP-0/blob/master/tree.yml#L1567

PhineasFreak:
Forgive me, i had not updated my RP-0 repo.

PR #475, opened 2016-03-03
SXT habitats

winged7:


NathanKell:
The DLK thing has a Science Lab in it, and should probably be more like 50000. I'll change the existing lab costs, which are also hilariously low.

winged7:
Price adjusted.

PR #479, opened 2016-03-06
Placed and Priced STX NTR

Zarbizaure:
See [RO configs](https://github.com/KSP-RO/RealismOverhaul/pull/1026) for these two NTR.

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -4049,9 +4049,13 @@ advNuclearPower:
 nuclearSupport:
 
 improvedNuclearPropulsion:
-
+    SXTNERVAB:
+        cost: 75000
+        entrycost: 750000 //lower than original Nerva since it's just an evolution
 advNuclearPropulsion:
-
+    SXTNERVA:
+        cost: 28000 //still lower thrust than the NERVA but higher ISP
+        entrycost: 1500000 //expensive bi-modal developpement, no sources
 fusionRockets:
 
 colossalRocketry:


NathanKell:
Ah, RP-0 note: when you are adding an item that is an evolution of an earlier item, you can add an entryCost modifier to EntryCostModifiers.cfg. [For example, the mid-period AJ10 is cheaper to unlock if you unlocked the early AJ10.](https://github.com/KSP-RO/RP-0/blob/master/GameData/RP-0/EntryCostModifiers.cfg#L64-L71)
I'm going to go ahead and make that change here, but just FYI going forward. :)

NathanKell:
Also, missed this. When you place/price something, make sure you delete any reference to it down at the bottom of the file in the ORPHANS node. Shouldn't be in there twice.

NathanKell:
Ah! And one more, I'm so used to C# / cfg that I missed this too until make complained. # makes comments in yml, not //

Zarbizaure:
Well it was my first RP0 config, but now i now! Thanks!
It make sense to make the Nerva II cheaper to unlock than the Nerva I (just more thrust and eficiency), however the bimodal NTR should still be expensive because it's an advanced engine (so over a million funds).

NathanKell:
Yep, not blaming you at all :)

The bimodal is still quite expensive even after you subtract the NERVA I entry cost, isn't it?

Zarbizaure:
You're right, over a million is still huge in RP0 (I though to i during the day). I'm currently in a midly-late carreer (about to perform my first Lars unmanned landing) and i'm totally out of fund.
It should more be something between ~200 000- 500 000 no? And the same for the nerva II...

NathanKell:
Yes, once you subtract you get something like 500-1000k for each, so it looks good to me.

PR #481, opened 2016-03-08
Add exp to Manned Landing contract

ctiberious:
Since RO severely restricts the experience Kerbals can get (mainly because manned missions beyond Earth SOI are extremely difficult) this should grant 1 exp to each Kerbal that lands on the Moon.  The experience gain is repeatable so if Jeb makes 10 landings, he'll earn an exp for each contract.


ctiberious:
Added same code to manned Mars landing contract so it will also grant an experience point for each completion.

NathanKell:
I would prefer to solve this by http://forum.kerbalspaceprogram.com/index.php?/topic/131381-requestidea-kerbal-training-and-other-time-based-kerbal-mechanics/ - the existing XP system doesn't really suit us anyway.

PR #483, opened 2016-03-09
RD270 cost revision

winged7:
Look at my previous discussion with @stratochief66, he said clearly that weaker variant RD270 should be around 20% more expensive than F1 (15000).
https://github.com/KSP-RO/RP-0/pull/271


NathanKell:
ah, ok.

PR #485, opened 2016-03-09
SXT Airbags

Zarbizaure:
[RO Reference](https://github.com/KSP-RO/RealismOverhaul/pull/1048)
@NathanKell it is ok this time with the `entryCostSubstractors` ?


NathanKell:
It seems odd to me that you can make money on 'em--the subtraction is higher than the entry cost for the part that is doing the subtracting! :]
But yes, the syntax looks correct to me! :)

Zarbizaure:
You're right, that's a bit too much. Something like ~13000 should be better :)

PR #488, opened 2016-03-11
RO-H1-RS27 NOCOST

stratochief66:
I believe a 1 character error here was causing this not to copy the H1 template made at FASAApolloLFEH1


NathanKell:
Good catch. I made some other changes though so I'll do it in my next push. Thanks!

PR #501, opened 2016-04-03
FRE Viking Engines support

Zarbizaure:
Support for my Forgotten Real Engines pack.


Zarbizaure:
Laos, they are RO viking engines placed later in the tech tree (Tier 4 instead of Tier 3). Which one is the best? Viking engines were basics engines (no regenerative cooling, low ISP) so I don't think they should be that late in the tech tree.

PR #505, opened 2016-04-07
Swap around Tantares_Habitation_A and Tantares_Orbital_A

SirKeplan:
A pruned part was being placed in the tree.


stratochief66:
Was it just a complete mistake that happened? I don't know what those parts are, are they pretty much the same part, but one was pruned and the other was the one we were using?

Just want a little more context before I merge this.

SirKeplan:
The two parts are a similar name, The one that is pruned by RO is placed in the tree in RP-0, and the proper one was put in orphans, so i assume what happened originally is someone accidentally switched them up.

PR #507, opened 2016-04-11
FRE Vega engines support

Zarbizaure:
I don't know if they are in the best node, move them around if needed. Also one KSP fund = how much euros/dollars? I have something like one KSP fund = $4000, I don't know if it's accurate (calculated from F-1 price). Since they are advanced SRBs, maybe they should be later in the tech tree. @stratochief66  or @NathanKell any idea about this?


SirKeplan:
The ksp costs are units of one thousand 1965 US dollars

Zarbizaure:
According to the inflation rates, $1 USD (1965) =$7,58 USD (2016) = 6,59 euros. So the Vega would have costed $4,86M (32M euros). It will change the price on this base.

PR #524, opened 2016-05-11
Support LE-5 (FRE)

Zarbizaure:
is the price/placing ok?

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -3138,10 +3138,6 @@ hydroloxTL5:
     
     KW3mengineWildcarXR: #HM7 engine for Ariane family
         cost: 1200 # quite rough, going of 11.6 million for the stage in 1985 dollars
-
-    FRELE5: #From FRE
-        cost: 3500 #assuming 30M stage cost in 1992 so about 15M engine cost
-        entryCost: 75000
         
 experimentalAerodynamics: #TL5 jets
     SXTKe90TurboJet: # Progress D-18T
@@ -3289,6 +3285,10 @@ hydroloxTL7:
     
     XKosmos_TKS_RD-0225_EngineLANDERS: # Kestrel 1B Retro
         cost: 40
+        
+    FRELE5: #From FRE
+        cost: 3500 #assuming 30M stage cost in 1992 so about 15M engine cost
+        entryCost: 75000
 
 stagedTL7:
     # AJ26-62 config here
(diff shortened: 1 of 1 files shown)


NathanKell:
The engine should probably be about half to two thirds the total stage cost. In terms of placement, it should be in hydroloxTL5, it was designed and put into service in the 1980s. You placed it alongside the RL10, designed in the 1950s. ;)

Zarbizaure:
I saw stage cost = 30M so engine cost ~15M, in 1992 USD.and I converted it in 1965 USD.
About the placing, why not, but it use the same technology as the RL-10...it was just developped later...
(also, I will add pricing & placing by engine config in a later update of FRE, so probably LE-5B and B2 will  end in hydroloxTL5, with a slightly modified cost)

NathanKell:
The Merlin 1D Full Thrust uses the same "technology" as the LR79 first flown in 1957. That does not at all mean that the Merlin 1D is at the same _level of technology_ as the first LR79. We place parts based on when they were designed and flown (with a few odd exceptions, but this is not it).

Let's compare: The LE-5 uses a gas generator cycle (same as the J-2 and LR87LH2, _not_ the same as the RL10's expander cycle). With that cycle, it gets ~450s in vacuum. The J-2, a hydroloxTL3 engine, gets only about 420.

Zarbizaure:
Yes, you're right. And the LE-5 has an advanced spark ignition system (rated for 16 restart). So I should move it to TL7? If I do this, I will probably cost the other configs in the same node, since they were'nt really developped later (LE5-B2 was developped in 2003, but it's mainly an improvement in term iof reliability).

NathanKell:
The base LE-5 was designed and flown in the mid eighties, and therefore should be in hydroloxTL5 as I mentioned above. The later variants...well, IIRC the LE-5A was a near-complete redesign (even switched to expander cycle!), so maybe TL6 makes sense?

Zarbizaure:
well I'm so stupid. Moved everything to TL7 and you said TL5 (facepalm)
I will move this back to TL5, and cost my other configs for TL6.

NathanKell:
No worries. :)

I gotta sleep though, I'll catch up in the morning!

Zarbizaure:
No problem!
The engine configs placing (for the LE-5 as well as for the Vikings) will probably come in the next release along the LE-7 engine.

NathanKell:
Thanks! :)

PR #525, opened 2016-05-13
added SSTU engines to tech tree

Temeter:
What it says on the tin! Adds all configured SSTU engines into the tech tree, following the same pattern and pricing as existing engines. F-1B went into the SLS-tech slot.

I've perled my own tech tree out of this and it seems to work just fine.

Only strange issue I encountered is the RD-107x (a 107 without verniers) costing 680 instead of the RD-107a's 700. While it makes sense in gameplay to have a non-gimballing engine to be cheaper, I did point them both to the 107's price to avoid using arbitrary pricing. So this is most likely an issue not connected with the tech tree, should you encounter the same problem.


jwvanderbeck:
<3 !

SirKeplan:
Cool! one note. instead of copying the cost from the RD-0110, you should probably put an `&RD-0110` on the first one, and an  `*RD-0110` on the SSTU one

stratochief66:
SSTU may need a special operator for altering the cost; although it might
already have one if it interacts well with stock career. for example, is
the price being doubled by SSTU when it changes from a 1 to 2 engine mount?

Temeter:
Good guess! Might be it. Should be somewhere in the RO configs then i guess?

Scaling is a bit broken as it seems. Not that the engines don't work, but it's really kinda weird. I just doubled my RD108, and the effect was a static price and tripple the weight. Doubling the RD107a increases price by 50%, but tripples weight.
Worth having the mod in there just for the great engine models, but fixing that probably is a bit more work.

Zarbizaure:
nice work @Temeter ! As I said I'm a bit busy with FRE right now but I will be happy to help writing configs for SSTU :-)

SirKeplan:
Another thing is the Merlin1A should be in TL5(veryHeavyRocketry)

Temeter:
^Done. Historically its a 2006 engine, but if we can put it ealier, i'm all for it for gameplay reasons.

Temeter:
Since good guy shadowmage just updated his Pods, I'll added them as well. Used generic prices, mind this doesn't neccessarily account for included equipment (put a disclaimer in there).

(this is based on his pull request, but names shouldn't change)

SirKeplan:
Yeah, historically it may be a 2006 engine, but there's a big performance difference between it and  the Merlin 1D(a 2013 engine) so it makes sense to place it further back in the tree (as the stock Merlin part already is). I like to think of the tech tree as a tree mapping the progression of tech, that just happens to resemble a timeline most of the time, as opposed to a strict timeline.

On that note, the Merlin 1C should be TL6, probably along with the F-1B ([Pyrios boosters are already in TL6](https://github.com/KSP-RO/RP-0/blob/master/tree.yml#L3192) , so it should make sense).

SirKeplan:
Oh, and if you can fix the merge conflicts too, that would be great :). seems some nearby changes happened in the file in the meantime

Temeter:
Yeah, I've seen how that's been done with nuclear propulsion as well, where you basically can play like research on Nerva never stopped. Liking that approach a lot! Also thanks for the input, gonna shift some stuff. ~

Now... How exactly do I deal with merge conflicts? Kinda expected them to show up, but gotta figure out first how to do it...

SirKeplan:
@Temeter ok, so i don't know your setup but you want to 'pull' the latest changes from this repo.
 if you have it set up with the command line client it will be `git pull upstream master`
https://help.github.com/articles/syncing-a-fork/

after you pull from upstream it will prompt you to fix the conflicts, then you commit that.

Temeter:
I'm not actually using a local repo, but only worked on this over the browser.

Got the desktop app, but damn, github is a fucking chore to work with. >_>

Temeter:
Got it. While the functionality is there, github gotta be one of the least intuitive UIs in human history.

Temeter:
So, I think the branch is ready for merging for now. There are more SSTU updates to come, e.g. for the multi-engines, but i'd personally rather get this out now and start a new pull request when necessary. 

You guys think that's good?

Temeter:
Added the AJ10s i've included in RO #1244.

Gotta test a last time, but then everything should be fine for now!

Temeter:
So, now everything is ready. If nobody sees anymore issue, stuff is ready to merge.

PR #528, opened 2016-05-15
Support for FRE LE-7

Zarbizaure:
The LE-7 engine is almost ready and will come with the latest FRE update, probably later today.
I will also include pricing by config.

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -3144,7 +3144,7 @@ hydroloxTL5:
         cost: 1200 # quite rough, going of 11.6 million for the stage in 1985 dollars
         
     FRELE5: #From FRE
-        cost: 3500 #assuming 30M stage cost in 1992 so about 15M engine cost
+        cost: 2650 #assuming 30M stage cost in 1992 so about 10M engine cost
         entryCost: 75000
         
 experimentalAerodynamics: #TL5 jets
@@ -3223,16 +3223,14 @@ hydroloxTL6:
     cryoengine-375-1: *RS-68
 
     KW2mengineVestaVR9D: # 2X japanese LE-7A
-        cost: 3840 # 50x less than development cost
+        cost: 12450
         entryCost: 192000 # 781M (1991 USD) http://www.lr.tudelft.nl/en/organisation/departments/space-engineering/space-systems-engineering/expertise-areas/space-propulsion/design-of-elements/cost/
 
-    cryoengine-25-1: # LE-7/LE-7A
-        cost: 1920  # Half the cost of dual model above
+    cryoengine-25-1: *LE7 # LE-7/LE-7A
+        cost: 6210  # Half the cost of dual model above
         entryCost: 192000 
         
-    FRELE7:
-        cost: 1920  #same cost as cryoengine-25-1
-        entryCost: 192000 
+    FRELE7: &LE7
 
 avionicsTL6:
     RLA_tiny_vac: # Marquardt/Aerojet R4D-11.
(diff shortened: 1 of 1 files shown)


Zarbizaure:
And...is my pricing for the LE-5 incoherent? Here I see that people placed LE-7 at 1900 funds...and my LE-5 cost 3500. What's wrong?

NathanKell:
When adding the second copy of an engine, please use *name to make the first one an anchor, and just reference it by &name in the second. I.e.

cryoengine-25-1: *LE7
cost: ...
entryCost...

FRELE7: &LE7

As to the pricing, I'm afraid I have no idea as to the cost of either, except what I saw on nautix regarding the LE-5's stage.  @winged7 what did you base that cost on? (I think that was your commit?)

Zarbizaure:
Oh yes I saw that. I will correct it soon.
Maybe the LE-7 is underpriced? Or the LE-5 is overpriced, but we need LE-7>LE-5.

Zarbizaure:
$30M for LE-5 stage http://astronautix.com/stages/le5.htm
$80M for LE-7 stage http://astronautix.com/stages/h21.htm (one engine)
So the LE-7 is about thrice the LE-5 price. How could we fix that?
Let's assume:

- LE-5 price: $15M by currency convertor: $3500k 1965 USD so **3500 funds**

- LE-7 price: $40M by currency convertor: $10000k 1965 USD so **10000 funds**
But these prices are way too high compared to other engines in KSP (example: Vulcain costs 6400 funds and is a bit more powerful than the LE-7; RL-10 (a bit similar to LE-5) also costs way less (1300 funds). I know there could be a difference depending on the tehcnology used (gas-cycle,...) but it doesnt explain a 2-3x factor.

NathanKell:
You'd want to compare with the period-specific RL10, which would cost quite a bit more (check the CONFIG additional costs). But yeah, I think probably what it comes down to is different costing for Japanese use, and also flight rate. As http://www.astronautix.com/articles/costhing.htm points out, in some ways this endeavor is a fool's errand. :]

I would therefore suggest going with 1/3 of the stage cost instead of 1/2. That places the LE-5 at about the same price/effectiveness point as the RL10A-4, and the LE-7 at about the same point as Vulcain.

Zarbizaure:
So about **6210 funds** for the LE-7 and **2650 funds** for the LE-5. This seems better than the previous prices.
Also, I know that price for an engine means absolutely nothing; as you said it really depends on the production number, which could explain why the RL-10 is much cheaper.

Should I change the price for all LE-7 in RO? If so you could accept my PR.

Zarbizaure:
apparently Travis failed, I will let you fix that.
Also thinking on my next engine I thought RO should really integrate next-gen avionics, like this one: https://rocketlabusa.com/about-us/electronic-systems/avionics/
This will reduce rocket's weight by a lot and therefore increase payload capacity. This avionic module is indeed on a small rocket, but it's a bit painfull to be stuck with the early and heavy saturn avionic in RO since there are much more performant ones...

Temeter:
I've encountered the same issue when updating SSTU tech tree. It's important to set the anchor first, and you got no references above. Seems like Nathan confused symbols, should be this way:

> cryoengine-25-1: &LE7
> cost: ...
> entryCost...
> 
> FRELE7: *LE7

Which is funny cause Nate was the one explaining that stuff to me. :D

Zarbizaure:
thanks @Temeter ! it should be fine now ;)

PR #529, opened 2016-05-15
Lower LE5 base price

Zarbizaure:
It has to be lower than the LE-7 first stage, right (which costs 1900)? If I'm wrong ignore this.


Zarbizaure:
closed because of #528

PR #530, opened 2016-05-15
Costing fuel tanks in the RP0-WIP category

dxdy-name:


NathanKell:
Has a merge conflict now that I merged your other one...

Also, what sourcing did you use for those costs? They seem somewhat high for "just tanks." If no cost is specified, the Proc Costs file should autocost the tank based on its volume...

PR #534, opened 2016-05-18
Tweak first launch

aw1621107:
Changes first launch contract to require the vessel to be moving upward at 1 m/s at some point. I originally set it to 0.1, but that showed as 0.0 in-game.


aw1621107:
Er... That first commit shouldn't be there. Hold on...

PR #535, opened 2016-05-18
Satellite Contracts - Meteorological Contracts

dxdy-name:
A small progression of contracts vaguely following along history. Starts with small contracts (Vanguard-2, TIROS, Micrometeorite measurements), ramping up difficulty with contracts for polar orbits (Nimbus) and then ending in two basic network contracts (GOES-SMS (inclined geosynchronous orbit) -> GOES First Gen (geostationary orbit)).

They're not exactly cash-cows, but give a bit of science in exchange.


dxdy-name:
Fixed some numbers that had the wrong units and organized the new contract group a bit differently to ensure there are only so many of the custmo Satellite contracts at once and that they're reasonable mixed.

NathanKell:
Apologies for the obscene delay on this!
Some notes:
1. Deadlines etc need to be given as 4x the desired days (since Contract Configurator assumes Kerbal days when converting to seconds).
2. The rewards look a bit high to me.
3. There's no inclination specified for the early satellites. Shouldn't they be in at least 51-degree inclination to get decent world coverage?
4. Could some of the parameters be randomized a bit? If we go with linear progression, there should be at least _something_ to differentiate different playthroughs. :)

Other than that, again, thanks so much!

dxdy-name:
-> 1: Are you sure? Putting 365 as the contract deadline and using earth time in the ingame display gives me a displayed contract length of one year, which seems absolutely correct. You might want to peruse IRC backlog on that, because for example the Sounding Rocket Contracts, which I gather were intended for a 90 day period (multiplied by four -> 360) actually afford you 360 days of time instead of 90 too. So it would seem that the setting does work in Earth time.
-> 2: Could you give an example? I tried to make the payouts similar to the existing contracts. For example, while the geostationary weather network one pays 50,000, it requires you to deliver two satellites to that orbit, which will either clog up your build pipeline significantly if you're doing it in two launches or require a relatively large launch vehicle if you're doing it in one, which justifies its payout in my mind.
-> 3: I'd like to keep the first one free from inclination requirements, since it's hardly fair to ask that of RP-0 players when the real-life equivalent was launched into about thirty degrees. Not to mention that it's less of a weather satellite than about measurements in the upper athmosphere. As for the second one, you're probably right that there ought to be some requirement on inclination in there.
-> 4: I'm a bit torn on that, to be honest, because it's such a lip service to variability. The differentiation between get a satellite into a 600x400km orbit, and get one into a 620x390km orbit is, I would propose, totally meaningless in differentiating different playthroughs, especially compared to the much more significant player-made differences created by their attachment to different launch vehicle designs and such.

jwvanderbeck:
@NathanKell Yeah on contract deadlines...several people are reporting all the deadlines in RP0 being 4 times longer than they used to be.  Following the code, it appears to me that CC is simply passing the given days directly to KSP it isn't doing any conversion (https://github.com/jrossignol/ContractConfigurator/blob/a36ebb9ed7df44f5e02157bf5b22c94e3555f304/source/ContractConfigurator/ConfiguredContract.cs#L151), so i'm wondering if maybe KSP is now actually aware if the setting for Kerbin/Earth time and setting it accordingly.  If that is true then we would need to chop everything down by 4.

dxdy-name:
@NathanKell Any thoughts on points 2 and 4?

NathanKell:
1. I have been conclusively proven wrong :D
2. I would suggest no more than 5000 for the advance for the simplest contracts (even an R-7 style LV is only going to be about 5000, a simpler one is 1000-2500 funds) and another 5000 for the reward. The failure penalty should be the sum of the advance and the reward, generally. Polar versions in LEO/MEO should only go up a mil or two each I'd say. A GEO launch should be about double that tops, I think? Depends on what kind of LVs people make, I guess.
3. Cool, makes sense.
4. It's meaningless in the grand scheme of things, yeah, it just...it feels cookie-cutter if it's the same every time. It's more about player feel than actually forcing different play, if that makes sense?

dxdy-name:
Changed the pricing & added randomness.

NathanKell:
Cool! :)

PR #539, opened 2016-05-22
Support for FRE Rutherford engines and Viking price tweak

Zarbizaure:
Explanation for the pricing:
Complete rocket price is $4.9M (2015). Given that engine price should not be too high (they are 3D-printed) and that the rockets uses more expensive components than usual (carbon tanks, next-gen avionics...) I though that the total engine price should be one quarter of the full price, so $1.225M. So $122500 per engine. Which means, after conversion to 1965 units, about 21 funds. And I suppose the vacuum version is a bit more expensive.

However entry prices are guessed.

About the node: I placed them along merlin engines, maybe they should be later?


Zarbizaure:
I still have no idea why the PR conflicts with RP0 master branch. Perhaps I should push from a new, clean branch; but It would mean doing a new PR.

PR #541, opened 2016-05-23
RN engines adds

raidernick:
-First pass engine adds for RN engines based on existing values for identical parts

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -438,11 +438,11 @@ start: #TL-1 early tech. Sounding rockets, basic structural pieces and wings. La
     rn_aerobee150_eng:
         entryCost: 0
         cost: 10
-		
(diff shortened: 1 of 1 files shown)


SirKeplan:
@raidernick YAML syntax is `&` to define an anchor, and `*` to make a reference to it, so you want to change round the `&BabySergeant` and `&LR101` bits. just have one `&BabySergeant` with cost, and the rest using `*` etc.

SirKeplan:
@raidernick look at the error message, it gives an error on line 866. In this case you are giving a cost when it's already defined once in the `&` bit. make sure there are no costs etc under any `*` references

NathanKell:
Is this done, or should I leave open?

Issue #490, opened 2016-03-14
Engine prices need fixing


Schnobs:
Reason 1) gutsy-feely. Soviet launchers are so much cheaper than US hardware, this just can't be right.

Reason 2) Gameplay: Things just don't work out. The RL-10 only becomes worthwhile at payloads in excess of 5t to the Moon / Mars; the Saturn5 is by no means big enough to justify J-2s on the second stage; and SRMs are like 10x as expensive as similar performance from liquid-fueled boosters. I guess there's good reasons why real-life rockets tend towards solids at lauch and hydrolox as early as possible, but Realism Overhaul doesn't reflect this.

Reason 3) sources. Knowing full well that the above two wouldn't convince anybody, I tried to find actual prices. http://www.coldwarweaponsystemcosts.com/resources/EnginesPDF6.pdf gives a few tidbits like
- E44 -- "NASA has ordered 30 F-1 rocket engines and varied support services from Rocketdyne at a cost of $141 million.  Source – Technology Week, November 28, 1966, p 14."  -- $4.7m per engine
- E19 -- "NASA has increased the value of North American Aviation, Inc., Rocketdyne Div., contract for the J-2 engine by $75.8 million for a new total of $206 million.
  The add-on provides for 48 more engines to be produced, raising the total purchase to 103 engines.Source – Missiles & Rockets, December 6, 1965, p 13. " --  $2m (all 103 engines) / $1.6m (last 48) / $2.36m (first 55)
- E56 -- "$19.7 million for manufacture of 114 [H-1] engines. Source – Aviation Week, November 25, 1963, p 30" -- $173k apiece
## 

Also of interest, http://www.coldwarweaponsystemcosts.com/resources/LaunchvehiclesPDF5.pdf
I especially like item  LV10: "$13m based on limited production of 8 flight vehicles [...] $9 million could be achieved with a production rate of 25 boosters per year for 5 years."

This, together with the give-or-take $800k for J-2s, suggests that we should treat any numbers as no more than a ballpark figure.
## 

coldwarweaponsystemcosts.com is a ternary source, but the best I could find; I've been browsing NASA press releases from 1965/66 and came up empty (at least when it comes to engines: there's a lot of construction and wiring contracts in the press releases). It's a shame, I'm pretty sure these things are public -- possibly not only the press releases but the actual contracts.
## 

[...]

PR #543, opened 2016-05-31
SSTU Merlin engines

Zarbizaure:
Merlin 1B/1Bvac (slighly less expensive than 1C) and XLR81 (part in developpment)

### DIFF
--- a/GameData/RP-0/EntryCostModifiers.cfg
+++ b/GameData/RP-0/EntryCostModifiers.cfg
@@ -2080,4 +2080,129 @@ ENTRYCOSTMODS
 			SXTAirbagSmall = 13000 #price = 7000 instead of 20000
 		}
 	}
+	
+	//Rutherford engines
+	PART
+	{
+		name = FRERUTHERFORD
+		entryCostSubtractors:
+		{
+			FRERUTHERFORDVAC = 50000 #price = 5000 instead of 55000
+		}
+	}
+	
+	PART
+	{
+		name = FRERUTHERFORDVAC
+		entryCostSubtractors:
+		{
+			FRERUTHERFORD = 50000 #price = 10000 instead of 60000
+		}
+	}
+	
+	//SSTU Merlin Engines
+	PART
+	{
+		name = SSTU-SC-ENG-Merlin-1A:
+		maxSubtraction = 16000 #down to 1050
+		entryCostSubtractors:
+		{
+			SSTU-SC-ENG-Merlin-1B = 16000
+			SSTU-SC-ENG-Merlin-1BV = 14000
+			SSTU-SC-ENG-Merlin-1C = 16000
+			SSTU-SC-ENG-Merlin-1CV = 14000
+			SSTU-SC-ENG-Merlin-1D = 16000
+			SSTU-SC-ENG-Merlin-1DV = 14000
+		}
+	}
+	
+	PART
+	{
+		name = SSTU-SC-ENG-Merlin-1B:
+		maxSubtraction = 18500 #down to 1500
+		entryCostSubtractors:
+		{
+			SSTU-SC-ENG-Merlin-1A = 10000
+			SSTU-SC-ENG-Merlin-1BV = 15000
+			SSTU-SC-ENG-Merlin-1C = 18500
+			SSTU-SC-ENG-Merlin-1CV = 1600
+			SSTU-SC-ENG-Merlin-1D = 18500
+			SSTU-SC-ENG-Merlin-1DV = 1600
+		}
+	}
+	
+	PART
+	{
+		name = SSTU-SC-ENG-Merlin-1BV:
+		maxSubtraction = 20000 #down to 2000
+		entryCostSubtractors:
+		{
+			SSTU-SC-ENG-Merlin-1A = 8000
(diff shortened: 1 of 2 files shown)


Temeter:
Wouldn't price the merlin alone. I did just copy the Merlin-1D(+1Dvac) prices for the other merlin engines.

1D is the only creddited price from an interview (which might be already outdated), while the Dvac price is an assumption.
Guess you could make the A/B/C progressively cheaper, tho, that way we'd have more consistent pricing.

Zarbizaure:
Make them cheaper? It make sense but then the entryCost needs to be more and more expensive. Do we have any entryCost source right now? (And, it means we'll have to use entryCostSubstractors).
I'm going to sleep now, so if anybody else want to work on that...

Zarbizaure:
I changed SL versions to 205/195/185/175 and vacuum versions to 220/210/200. Maybe we could do better, but I honestly don't know withouth sources.

Temeter:
Sorry, my post was a bit messy i guess, I blame being distracted and bad english.^^
 
The Merlin-1D is 175/200, following real sources, and you brought up the idea of making the 1B cheaper. So I was saying currently Merlin-1A/1C still cost 175/200, so they maybe should be cheaper as well!

Entry cost is a bit of a mess anyway. Merlin engines are developments based of each other.

Zarbizaure:
That's why we should use entryCost multplier (or substractor). That way, if we unlock one of them, all of the other merlin engines will have their prices reduced. Maybe I'll do that tonigh.
(And I've also a bad english, I'm french)
I don't think earlier should be cheaper (after some though) because the 1A/1B were just prototypes so they were more expensive to produce (but less expenseive to developp).

Temeter:
German here. ;)
I don't think we should make engines more expensive on the base of them being prototypes. A lot of Realism Overhaul/RP-0 is about making your own alternative history. I think NathanKell made that point himself: You basically can play like research on nuclear engines never stopped after the NERVA (which only had a ground based prototype).

So if you buy rockets out of Merlin-1B's, then they aren't prototypes but actual mass production in your history.

Zarbizaure:
I can speak german but I don't think RO guys will like that x)
Yes, but an engine that is produced only in a little number won't have its cost lowered; if you produce a large amount of engines, their price will decrease over time (or for any other stuff).
That's why I think a decreasing curve would be good. Maybe @NathanKell could help us?

Temeter:
Good point, that's exactly what came up in another issue: https://github.com/KSP-RO/RP-0/issues/490

Russian engines like the RD-253 are ridiculous cheap, because russians/soviets were better at mass producing, while a lot of the american material was custom made.

Zarbizaure:
I though exactly to this Github issue.

Anyway, since the Merlin 1B and 1C are in the same node, players will probably never play with the 1B. But it make sense since they were never used on a rocket (for others reasons like reliability, and the lack of regenerative cooling).

Zarbizaure:
I'm closing this and I'll open a new PR from my master branch.

Zarbizaure:
It should be fine now. Sorry if I closed/reopened, I'm still pushing from my pach-6 branch.
The placement is probably far from perfect, but it's better than nothing! Plus every player will at least purchase 2 merlin engines, which lowers to the minimun price mots of the time.

Temeter:
Github is a mistery of our times.

If nobody does it in the meantime, i'll check those files (+RO) ~tomorrow. Atm busy fixing my broken, half updated and incomplete installs, as well as checking SSTU pricing fixes.

Zarbizaure:
I don't have time to test it too; but I think it should be ok.
We should also definitely set entryCostModifiers for the RL10s too.

NathanKell:
Whoops, this conflicts now. Also, are the prices corrected for inflation?

Zarbizaure:
For the prices I'm based on what Temeter found for the Merlin D. I don't know how accurate it is, however it seems to be a reasonable price.

Why the hell is there conflicts? :P anyway you can just copy/paste these few lines into the designed files; I'll still look at it later and try to find the error.

Zarbizaure:
so, 2.17 millions for one merlin 1D (https://www.quora.com/How-much-does-it-cost-to-produce-a-single-Merlin-engine-in-SpaceX-rockets)
Which is 369 funds (1965) so the engines are indeed slightly underpriced. I'll higher the cost a bit.

jwvanderbeck:
There are probably conflicts because a previous PR that was merged changed the same file as you.

Sadly this is what happens when the project accumulates a ton of PRs then accepts them all at once rather than over time.

Zarbizaure:
Yes I understood that...
Obviously that's a small PR so maybe people with merging rights can copy/paste this manually? Otherwise I would have to start a new PR, I guess?

jwvanderbeck:
strangely your branch is trying to merge files it shouldn't be, so let me look at it

jwvanderbeck:
ok looking at this now.. let me see if I can clean it up.

jwvanderbeck:
So basically your changes on #539 are conflicting with this now.  I'm fixing it now.

jwvanderbeck:
Cleaned and merged

Zarbizaure:
thanks! :-)

PR #544, opened 2016-06-03
Adjusted all contract deadlines to use Earth days

jwvanderbeck:
Deadlines appear to now be passed directly from CC to KSP and KSP appears to properly use Earth days when in Earth mode, so adjusted all contract deadlines to be in Earth days.

Earth.

### DIFF
--- a/GameData/RP-0/Contracts/Fly-by Contracts/DeimosFlyByContract.cfg
+++ b/GameData/RP-0/Contracts/Fly-by Contracts/DeimosFlyByContract.cfg
@@ -19,7 +19,7 @@ CONTRACT_TYPE
 	targetBody = Deimos
 	maxCompletions = 1
 	maxSimultaneous = 1
-	deadline = 2920 // 2 years
+	deadline = 730 // 2 years
 	prestige = Significant
 	
 	// rewards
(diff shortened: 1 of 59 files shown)


Zarbizaure:
2 year for some contracts is kind of short (Mars and Venus in this case). Indeed, there is a transfer window every two year for mars, and you then have a transfer time of 6-8 month...which could exceed the contract time. I suggest setting it to 3 year (for Mars at least).

leudaimon:
@Zarbizaure Aren't those contracts supposed to be accepted during the transfer window?

jwvanderbeck:
To be clear, all I did was adjust the dates from Kerbal to Earth days.  This puts them back to what they were pre 1.1.  I didn't actually change any lengths.

Zarbizaure:
@leudaimon You can do so, but honestly I just usually accept them and then check when I'll have a new transfer window. I think the correct length would be : "max time between transfer window + average transfer time + 10% margin or so". Which would be arount 1070 days for mars, so around 2 year and 11 month.
And of course, include an eventual gravity assist margon for farther bodies (like it's already for somes)

NathanKell:
Yep, probably worth a re-examine of contract windows. Certainly HSF contracts will need longer deadlines once training becomes a thing.
That said, that's a different undertaking than this pull.

PR #546, opened 2016-06-07
SSTU Tanks and procedural decouplers

Zarbizaure:
The tanks pricing should be fine with medium-sized tanks (aligned with PP); indeed, SSTU does not use the same pricing system and a flat and wide tanks will be less expensive than a long and thin one, for the same volume.

Decouplers are just here not to appear in the "non-RP0" section; I do not have a way to change their price (radial one is less expensive, stack one is more expensive).

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -1458,6 +1458,10 @@ survivability: # 3-Axis Stability and early probes ##
     Heatshield-5M:
         entryCost: 30000
         cost: 1700
+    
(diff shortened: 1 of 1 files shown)


Zarbizaure:
I'll try to add interstage and fpayload fayrings btw (In my opinion, SSTU fairings are better than procedural fairings - they have more textures, simpler system...)

PR #550, opened 2016-06-20
BobCat SovietEngines Update 1.1.2 Tree

Theysen:
renamed the few placed parts to correct config names, cloned parts are
unaffected from this


NathanKell:
Uh, doesn't loook like that's all of them though?

Theysen:
I don't know how this happened, I'm very sorry. Must have screwed up the search function in my editor.

PR #551, opened 2016-06-20
Add Rocketdyne E1 engine

jwvanderbeck:


NathanKell:
Probably should get the same entrycost modifier as the H-1 gets (i.e. you get some of what you spend on LR89/79 back).

PR #557, opened 2016-07-02
change the highest tier runway to unlimited size

Zarbizaure:
at least it should be more than 40: the settings are : "width, height, lenght", and a lot of planes are longer than 40m.


NathanKell:
Then it should be 500,40,500 - I specifically do not want unlimited height, that's what the pad is for. You shouldn't be able to launch rockets from the runway.

NathanKell:
Thanks for the find, though, to be clear! :) It was a stupid typo!

PR #558, opened 2016-07-04
Rss expansion contracts

hargn:
This is my first contribution to RP-0 or Realism Overhaul suit of mods, so be indulgent if I did something wrong

This PR adds Landing and Flyby Contracts for bodies added by the RSSExpansion mod.
RP-0.ckan is modified to make RSSExpansion as recommended mod.

The rewards maybe need some balancing.


NathanKell:
Right now RSS Expanded seriously messes with the inclinations of everything. I will not support it while it does that, in any of my mods. We have the inclinations we do for good reason.

hargn:
OK, I just saw the RSS Expansion thread.
I keep the branch just in case if the author fixes the inclinations of planets.

If I want to continue to bring some new contracts to RP-0, do you have any ideas?
I read the issue #177 and I think to make some "Return Sample" contracts that will work with mystery goo container for the probes and/or samples taken by kerbals.
What do you think about it?

SirKeplan:
Slightly going off on a tangent here, but i think a dedicated surface sampling part is necessary, for RO probably. Either a robotic scoop thing, or maybe repurpose a stock drill as a core sampler. This would  be an extra scuebce experiment part i suppose

NathanKell:
@hargn I have lots of ideas, and it's be _AWESOME_ if you made more contracts! :) @dxdy-name has made a set of earth science contracts, so maybe you could do some lunar science contracts? Stuff like "get science from the far side of the moon and transmit it", then repeatables "orbit with x characteristics", "land uncrewed at x biome", etc., or building up a comms network.
Same with the other planets--the goal here is to replace the stock Satellite Contracts system and the stock Science From X contract system

I suggest making a new issue to track/discuss all this :)

Issue #566, opened 2016-07-12
Uncrewed planetary exploration contracts


leudaimon:
Following up discussion on #562, I intend to build a TODO to have interesting satellite contracts for the exploration of other bodies. After putting some thought into this, I think a nice progression would be something like this:

- [ ] First Orbit contracts - similar to the already implemented flybys. 
  * requirements: any of: flyby of the body and orbit of a body in the previous "tier"
  * parameters: in orbit and PeA below flyby limit/2-3?

- [ ] Polar orbit contracts - could be a repeatable, demanding different transmittable scientific payloads
  * requirements: First orbit of the body
  * parameters: have some specified experiment and inclination 90+-5°
  * additional idea: bonus credits if the vessel transmits science from all biomes

- [ ] Low polar orbit contracts - can also be repeatable.
  * requirements: Not sure, First orbit or Polar orbit?
  * parameters: circular orbit, below SpaceLow threshold, inclination 90+-5°
  * also bonus (or maybe even required) transmit science from all biomes

- [ ] Impactor - similar to moon impactor, I'm not sure how to deal with this for gas giants, even tough it has already been done.
  * requirements: any of: first orbit of the body or impactor on a body in the previous "tier"
  * parameters: similar to the moon impactor for rocky bodies, don't know about gas giants
  * I think it makes sense to have this with limited completions for each body
  
- [ ] Lander - similar to moon landing, can be made as a repeatable right away for the other (rocky) bodies.
  * requirements: any of: impactor of the body or lander on a body in the previous "tier"
  * parameters: similar to the repeatable moon impactor
  * both for the moon lander and for the other bodies, I think it would be cool to restrict contracts for the different biomes, selecting an unlanded one each time. Thus, maxcompletions would be the number of biomes of each body.

- [ ] High orbit communication satellite for relay purposes - can be an extension from the Comsats currently in development. *Not necessarily last - actually parallel to the others and not exactly scientific*

Any additional ideas? Anything worth changing? Any problems?

NathanKell:
These look good to me. In addition
* Rover contracts to survey an area using waypoints (steal from CC's science contract pack?)
* Sample return

pap1723:
@leudaimon not all of these were implemented with the new contracts, there are still some good ideas left on here!

leudaimon:
[...]

PR #568, opened 2016-07-13
Flyby Contracts for the gas giant moons

leudaimon:
Creates Fly-by contracts for the gas giant moons. 
- Rewards (loosely) based on the difference between Mars and Phobos fly-bys
- Minimum distance kept the same 20Mm, given all flyby missions (except the moon) have this
- Deadline is the same as the one for parent body flyby
- Requirement is parent body flyby

### DIFF
--- a/GameData/RP-0/Contracts/Fly-by Contracts/CallistoFlyByContract.cfg
+++ b/GameData/RP-0/Contracts/Fly-by Contracts/CallistoFlyByContract.cfg
@@ -5,9 +5,9 @@ CONTRACT_TYPE
 	group = Fly-ByContracts
 	
 	title = Callisto Flyby
-	description = Design and successfully launch a probe on a flyby of Callisto with a closest approach of 20000km or closer, then record observations and transmit.
+	description = Design and successfully launch a probe on a flyby of Callisto with a closest approach of 2000km or closer, then record observations and transmit.
 	
-	synopsis = Fly by Callisto closer than 20000km and transmit science.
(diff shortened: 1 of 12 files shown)


NathanKell:
Good stuff! Will have a look asap! :)

leudaimon:
@NathanKell, I'm doing the stuff from #566 in this branch: https://github.com/leudaimon/RP-0/tree/UncrewedExploration given it's big and may have bugs. Is this the "correct" way?

NathanKell:
Coming up for a bit of air, gonna try to get back to this. Apologies.

leudaimon:
Good point. I was in doubt about the reason for the planet flybys to have the same altitude for all planets. That's why I kept the same for the moons. I guess 2,000km might be enough? 

Btw, these contracts will need the corrections for the CC update also.

leudaimon:
Sorry for the spam, fixed this in the online editor, don't know how to edit multiple files in a single commit here.

rsparkyc:
Was looking at merging this, what exactly do these need as far as CC updates?

stratochief66:
I looked through them, but I don't know enough CC to answer that myself. We'd either need to leave these be for now, or else someone would need to learn enough CC.

rsparkyc:
I remember doing changes for the contracts we had already to update them to the new version.  The main thing I remember doing was removing the weight parameter, which this doesn't have.  I'm trying to load them into my setup now to see what happens.

rsparkyc:
these seem to load fine in my install

stratochief66:
Nope!

PR #572, opened 2016-07-25
Support new SSTU XLR81 engines

Zarbizaure:


jwvanderbeck:
Should additionally add cost modifiers from one to the other I think so that the pricing remains as it is now.

Zarbizaure:
of course I'll add them.

Zarbizaure:
so where is this Agena entryCost? I did not found it, maybe it's not properly configured?

SirKeplan:
Closing this as i think it's covered in #577 , and it looks to me the entrycostmodifiers aren't necessary to add, these are different variants of the engine.

correct me if i'm wrong.

PR #580, opened 2016-08-29
Add Tantares_Engine_C, same as A and B.

ahmedcharles:


SirKeplan:
interesting, this has brought to light an error in the tree, the `Tantares_Engine_A` is for Soyuz T onwards, which should be in a later node, it first flew in 1979.

ahmedcharles:
Which node represents 1979?

SirKeplan:
`specializedControl` is probably the place for this.

Though I must say that nodes aren't really designed to match specific years, as sometimes new tech comes along based on fairly old designs.

ahmedcharles:
That's probably slightly better. :)

SirKeplan:
Looks good to merge, unless anyone has an idea for the price of these service modules, they should be different in real life.

PR #584, opened 2016-08-29
Add LLV_* engines and decouplers.

ahmedcharles:
I'm interested in suggestions for the costs on the C and D engines, since I don't have a good reference point.

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -2730,6 +2730,10 @@ propulsionSystems: #staged combustion TL3 -- NK-15 here.
         cost: 20100
     LLV_Engine_B: # in 8x NK-15V configuration
         cost: 6240
+    LLV_Engine_C: # in 4x NK-31 configuration
+        cost: 2000
+    LLV_Engine_D: # in 1x NK-19 configuration
+        cost: 450
 
     ALV_Engine_C: #RD-213/214
     Libra_Engine_B: #Block D RD-58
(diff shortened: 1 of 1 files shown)


SirKeplan:
hmm, looks like there is already a bit of a prototypical inconsistency here, the N1 3rd stage used 4 X NK-21s (later upgraded to NK-39s but unflown). this might be worth fixing too.

ahmedcharles:
I was just looking at the configs for the N1 Block A and they all have different values for stuff. Should they be consistent (and perhaps listed in one place)? I know that's not stored in this repo, but in RealismOverhaul.

SirKeplan:
Huh, just looked at them myself, there must have been a great deal of confusion when they where made. some of the numbers are way off or from the wrong engines.

For now i'd say you should leave out the `GameData/RP-0/Tantares_Engines.cfg` file, then as you say some stuff needs doing in the RealismOverhaul repo.

ahmedcharles:
What do you suggest in terms of pricing, or should I just leave that out for now?

SirKeplan:
probably base it on multiples of these this https://github.com/ahmedcharles/RP-0/blob/057bbd066c5094f03f6bee06967a41151077fa07/tree.yml#L1978-L1982 , though these are kinda provisional as just based on the other engines prices.

ahmedcharles:
Done, I think :)

PR #586, opened 2016-08-29
Add support (tech node) for the other Capella parts.

ahmedcharles:


SirKeplan:
Seems `Capella_Solar_A` isn't even configured in RO yet, not sure of it makes sense to merge this yet.

ahmedcharles:
The benefit here is that at least it won't show up in a random node somewhere, even if it's not configured yet.

PR #587, opened 2016-08-29
Add Libra_Engine_A.

ahmedcharles:


SirKeplan:
There isn't a Libra_Engine_C or D?

ahmedcharles:
There are only two stages in Tantares LV for the Libra. Why is that surprising?

SirKeplan:
because you talked about costs for the C and D engines.

ahmedcharles:
Yeah, it's an issue where github/my browser puts messages from previous PR's into future PR's... it's annoying.

SirKeplan:
OK, seems the Tantares part is effectively an integrated RD-858 and RD-859, so this should be double the price.

ahmedcharles:
Updated

PR #589, opened 2016-09-13
Price and place the Gemini Lander SM (and typo fix).

ahmedcharles:


stratochief66:
@ahmedcharles 

I think that the entry cost has an extra zero in it, or am I mistaken?

rsparkyc:
look like he added 20% to the FASAGeminiUtilityPack, which has an entry cost of 178000, so it looks like the number is right (unless what it's based off of is wrong)

ahmedcharles:
Yeah, I don't remember, but that's what the comment says. The values look correct for being 20% more.

stratochief66:
The original is wrong then. Generally, the entry cost is 20x the part cost, this is 200x

stratochief66:
@ahmedcharles could you adjust your addition & the original to be 20x?

ahmedcharles:
I accidentally deleted the repo, so I had to create a new pr. #618

Issue #591, opened 2016-09-21
Weather Satellite Contract Completes After One Launch


gruneisen:
The Weather Satellite Contract to insert three geosynchronous satellites in orbit completed out after only inserting a single satellite into orbit.

hargn:
I see what is happening : 
`PARAMETER
   {
   name = IsNotVessel
   type = IsNotVessel
   vessel = weatherSatellite
}`
are missing in contract parameter group.
I'll probably release the fix this weekend

PR #593, opened 2016-09-30
Fix for issue #591 - Weather Satellite Contract Completes After One Launch

hargn:
Fix for multiple weather satellite contracts completed after first satellite insertion #591 

Polish required ElectricCharge Amount

Increase rewards to match current (stock) satellite rewards (30k per satellite approximatively)
Can be reverted if that's too much

### DIFF
--- a/GameData/RP-0/Contracts/Satellite/WeatherContracts.cfg
+++ b/GameData/RP-0/Contracts/Satellite/WeatherContracts.cfg
@@ -355,9 +355,9 @@ CONTRACT_TYPE
 	maxSimultaneous = 1
 	
 	// reward block
-	advanceFunds = 30000.0
-	rewardFunds = 20000.0
-	failureFunds = 20000.0
+	advanceFunds = 40000.0
+	rewardFunds = 60000.0
+	failureFunds = 60000.0
 	rewardReputation = 15.0
 	failureReputation = 50.0
 	
@@ -381,6 +381,18 @@ CONTRACT_TYPE
 		}
 		PARAMETER
 		{
+			name = IsNotVessel
+			type = IsNotVessel
+			vessel = weatherSatellite2
+		}
+		PARAMETER
+		{
+			name = IsNotVessel
+			type = IsNotVessel
+			vessel = weatherSatellite3
+		}
+		PARAMETER
+		{
 			name = HasCrew
 			type = HasCrew
 			minCrew = 0
@@ -398,7 +410,7 @@ CONTRACT_TYPE
 			name = HasResource
 			type = HasResource
 			resource = ElectricCharge
-			minQuantity = 30000.0
+			minQuantity = 30000.0+Round(Random(-10000,10000),500)
 		}
 		PARAMETER
 		{
@@ -425,6 +437,18 @@ CONTRACT_TYPE
 			name = NewVessel
 			type = NewVessel
 		}
+                PARAMETER
+                {
+                        name = IsNotVessel
+                        type = IsNotVessel
+                        vessel = weatherSatellite
+                }
+                PARAMETER
+                {
+                        name = IsNotVessel
+                        type = IsNotVessel
+                        vessel = weatherSatellite3
+                }
 		PARAMETER
 		{
 			name = HasCrew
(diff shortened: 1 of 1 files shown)


ctiberious:
I was going to submit a PR for some changes to the "WeatherContracts" but since they would interfere with this PR, I figured I'd just make the suggestion for the changes here.
1) For "Atmospheric Analysis", I'd suggest that the minPeA parameter be altered to to "Random(200000, 580000).  This allows the contract to simulate not only Vanguard 2 (as indicated in the contract notes) but also Vanguard 3 (515km perigee) as well as all Sputnik 1-3 (approx 215km perigee).
2) Also for "Atmospheric Analysis", either completely remove the "maxCompletions = 1" or increase the maximum number of times you can run the contract.  If you remove the maxCompletions altogether, I'd add an additional requirement that the contract is no longer available once you complete either "WeatherSecondA" or "WeatherSecondB".  As it currently stands, once you complete "Atmospheric Analysis", there are no more RP-0 satellite contracts that can be run using Early tech so you basically have to go back to running nothing but "sounding rocket" contracts while you wait for Basic tech nodes to open up.
3) For "Meteorogical Satellite", the name should be corrected to be "Meteorological Satellite".
4) For both "WeatherSecondA" and "WeatherSecondB" contracts, I'm suggesting the range between minApA and maxApA be increased.  By default, "WeatherSecondA" only has a 10km split while "WeatherSecondB" has a 50km split between minApA and maxApA.  That's a pretty tight margin for Early and Basic avionics, especially since a lot of satellites at this era will have no ability to make even minor corrections after their final stage is spent.
5) Also, for both "WeatherSecondA" and "WeatherSecondB" contracts I'd remove the "maxCompletions = 1" for the same reason as I stated in #2 above.  These either need to be runnable up to 3 times each or until the player completes "WeatherNimbusOne".  Otherwise you have no contracts while you wait for more tech nodes to complete research.

hargn:
Added some polish to periapsis and apoapsis of randomly generated orbit.
Point 1 and 3 are fixed as required by @ctiberious.

I think the other point don't need to be fixed, I'm currently (and occasionally) on ComSat contracts that will provide some progression via milestone on kind of orbits).

I suggest you, to make the improvement you want in submit them via a new PR.
There is also a numerous issues describing the new contracts that would be nice to have, you can inspire yourself to add more contracts.

rsparkyc:
With the current state of things, I haven't found that I need to complete these contracts more than once, and have no issues with funding, so I think having them capped is ok.  Also, I'm going to try to complete some of these with some early tech to see if they seem too hard.

ctiberious:
@rsparkyc On my current play through I can't say I had issues with funding.  Only that the primary way I got funding was by repeating the sounding rocket contracts over and over.  Currently I'm up to something like 253 launches (I'm up to mid Apollo era) and close to 100 of those were simple sounding rockets.  Primarily because in the early stages of a career they are the only contracts that make any funds.  I'm not saying that the game is unplayable with the current contracts.  Only that I think RP-0 needs more early repeatable contracts so you aren't limited to launching the same few sounding rockets over and over again.

Sounding rockets are great in the earliest stages of the game.  And they are also very useful for testing parts for Test Flight.  But once you can start doing satellites, I think we need more than the 7 contracts (only one of which is currently repeatable) that currently exist.

Finally, I'm not saying that the existing satellite contracts are impossible with very early rocket tech.  Just that they are extremely difficult.  I found completing some of the early satellite contracts with very early historic rockets (Sputnik, Vanguard, Juno-I) to be harder to complete than the moon landing contracts I'm doing now, mostly because the orbital margins are extremely tight considering you have extremely limited attitude control, no restart and (often) no engine shut-off capability.  Considering the contracts I commented on are supposed to be "Trivial" difficulty, I just feel the orbit margins should be loosened a bit.

rsparkyc:
Ok, that makes sense, especially since these are "Trivial".  Since this PR is really aimed against fixing https://github.com/KSP-RO/RP-0/issues/591, it sounds like these changes (and the one I mentioned above) would probably be worth doing in a separate PR.  I think I'll merge this (because this does not make anything _worse_), and create an issue to track that work.

rsparkyc:
Ok, created an issue here to track these improvements: https://github.com/KSP-RO/RP-0/issues/614

PR #595, opened 2016-10-09
Rn dnepr place

infinitybit:


raidernick:
@stratochief66 can this be merged also. It has configs for my parts, they appear to be good and I don't want them to end up conflicting with configs Probus wants to start doing for this.

PR #597, opened 2016-11-20
Experiment cooldown & shielding

PhineasFreak:
* Add cooldowns for the experiments (cannot be rerun immediately) and
"is shielded" checks.
* Remove some trailing spaces.

[Alternate diff (ignoring whitespace changes).](https://github.com/KSP-RO/RP-0/compare/master...PhineasFreak:RP0-Experiments-Cooldown-Shielding?w=1)

### DIFF
--- a/GameData/RP-0/Science/Experiments/AtmosphereAnalysis.cfg
+++ b/GameData/RP-0/Science/Experiments/AtmosphereAnalysis.cfg
@@ -23,6 +23,8 @@
 {
 	@MODULE[ModuleScienceExperiment]:HAS[#experimentID[atmosphereAnalysis]]
 	{
+		%cooldownTimer = 15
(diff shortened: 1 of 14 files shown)


PhineasFreak:
Note to self: find what the is problem with the experiments stating that "Cannot perform experiment, must wait 00:00:00".

PhineasFreak:
@stratochief66 did you have the same problem with the experiments as i have? Because i was unable to perform the experiments due to the message stated above.

stratochief66:
Odd, no I didn't. It looks like the way you squared #612 will remove these changes, so that will give us a chance to try again?

PhineasFreak:
I tried at least three times to debug it with no success (new installation, only the minimum mods, making sure that everything is updates) . Every time i got the same message.

PR #598, opened 2016-11-20
RO PF flat interstage adapter P&P

PhineasFreak:
* Place and price the Procedural Fairings flat interstage adapter from RO.


PhineasFreak:
Working on an old branch is always a bad thing to do...

PR #599, opened 2016-11-20
RO Castor IV & Star 37 P&P

PhineasFreak:
* Place and price the RO Castor 4 and Star 37 SRMs.


rsparkyc:
Looks like an extra RT1 node at the end of the file is causing errors when writing out tree.cfg

PR #601, opened 2016-12-07
SSTU MUS Tanks

ctiberious:
Right now, if you use one of the SSTU MUS tanks, you have unlimited avionics control.  Difficulty is, these tanks are procedural so the larger the tank, the more avionics you'd need for it.  What I've done here isn't pretty since the parts will now display 9 separate avionics modules, but in VAB/SPH and in flight, they do combine correctly.  So if, for example,  you have Flight Control, the MUS tanks can handle up to 40t for avionics purposes.  This doesn't have anything to do with the actual size of the part you create, but it does allow these tanks to function better throughout the course of a career game.  If there is a cleaner way to setup something like this, that would be a better option, but this at least works.

### DIFF
--- a/GameData/RP-0/Avionics.cfg
+++ b/GameData/RP-0/Avionics.cfg
@@ -675,3 +675,60 @@
 		massLimit = 150.0	// Just a guess but since the TRS was supposed to reposition Skylab, I figured it should be set high
 	}
 }
+@PART[SSTU-SC-TANK-MUS-*]:FOR[RP-0]
+{
+	MODULE
+	{
+		name = ModuleAvionics
+		massLimit = 5.0
+		TechRequired = earlyAvionics
+	}
+	MODULE
+	{
+		name = ModuleAvionics
+		massLimit = 5.0	// should add up to 10t
+		techRequired = basicAvionics
+	}
+	MODULE
+	{
+		name = ModuleAvionics
+		massLimit = 10.0	// should add up to 20
+		techRequired = stability
+	}
+	MODULE
+	{
+		name = ModuleAvionics
+		massLimit = 20.0	// should add up to 40t
+		techRequired = flightControl
+	}
+	MODULE
+	{
+		name = ModuleAvionics
+		massLimit = 60.0	// should add up to 100t
+		techRequired = advFlightControl
+	}
+	MODULE
+	{
+		name = ModuleAvionics
+		massLimit = 150.0	// should add up to 250t
+		techRequired = specializedControl
+	}
+	MODULE
+	{
+		name = ModuleAvionics
+		massLimit = 250.0	// should add up to 500t
+		techRequired = avionicsTL5
+	}
+	MODULE
+	{
+		name = ModuleAvionics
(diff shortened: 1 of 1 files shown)


stratochief66:
Avionics add weight and cost as well. We do have a few tank parts with avionics built in, but IIRC we've adjusted the price of those parts to reflect it, and also they were tank/stages that actually had avionics built into the stage.

Doing that (ie, some of the FASA tanks) was done to enable 'recreation overhaul' behaviour, so when someone builts an Atlas-Mercury, for example, they didn't have to clip an RP-0 avionics unit of appropriate size into the tank.

IMO, the better way to handle this would be to simply make these tanks not have avionics powers at all, and act as plain tanks?

@rsparkyc @SirKeplan @NathanKell @jwvanderbeck ?

SirKeplan:
I have hardly used SSTU tanks, but it seems to me best to treat them as just plain tanks. leave avionics as a separate part. I sometimes find myself offsetting avionics rings into proc tanks which is not ideal either, but i tend to agree with @stratochief66 on this.

ctiberious:
Fair enough but in that case, we really need a procedurally sizable avionics component.  Something we can adjust the diameter of so it can be fitted to a procedural tank we create regardless of size.  I specified the SSTU MUS tanks here because I use them a lot (I like the built in rcs thrusters) because I can use one part to recreate a lot of different 2nd stages.  But you can do the same thing with procedural parts and you run into the same basic problem.  You either get no avionics, you're limited to creating stages of specific sizes that match the static avionics parts we have, or you have to get really creative to fit "oddly" sized parts.

I know a procedurally sized avionics part wouldn't be overly complicated, but that giving that part different avionics capability based on size and tech would be.  Maybe a viable alternative would be to create a small radial part with a little avionics, battery, mass and electrical use.  Then you could just fit multiple parts as you need them.

rsparkyc:
Ultimately I'd like to figure out what the _right_ way is to handle avionics.  IRL, what are "avionics units" actually doing?  Do they have hydraulic pumps to drive the gimbals in engines?  What makes it so a bigger unit can control a heavier rocket?

I also don't use SSTU part, and generally try to put my avionics units either above my tanks (if their profile lines up flush with my rocket design), or under a payload fairing base.

Temeter:
I always wondered what exactly constitutes the relationship between rocket size and avionics module mass. It's certainly only electrical equipment. The main Saturn 5 instrument units was in the third stage.

Wikipedia has a quite extensive article on it:
https://en.wikipedia.org/wiki/Saturn_V_Instrument_Unit

Of course, it gets more complicated because there are computers on every single stage on an S5, including the electrical equipment to control the stages. Well visible on this adorable picture:
http://i.imgur.com/NAok3jp.jpg

S5 is of course more complex by nature, since it has a complex structure/mission and multiple redundant equipment because of potential danger to the crew. That alone might explain why the avionics are also much heavier.

stratochief66:
I think we'd just want the SSTU tanks to have no avionics. Although procAvionics may be on the way.

PR #603, opened 2016-12-15
Reduce cost of H-1 family (issue #490)

ec429:
Price 750, comparable to the cheapest LR79 (S-3D) configs, since it's
largely a simplification of that engine.  H-1B pushes up to 1050 to pay
for additional reliability / man-rating, then RS-27 are cheaper again
(at 800, 825) because they're mature tech and are the inexpensive option.

No actual sources for these numbers.

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -1863,7 +1863,7 @@ advRocketry:
 
     # Saturn I/IB H-1 engine.
     FASAApolloLFEH1: &H1
-        cost: 1050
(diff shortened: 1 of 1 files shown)


stratochief66:
I like sourcing, but I also like cheaper H-1 engines...

This is the challenge of deriving part prices from what data we can find on overall launcher or stage costs. Leaving the entry cost high feels like a decent compromise for now.

PR #604, opened 2017-01-05
Heat Shield ECM fixes

ec429:


stratochief66:
This seems like step forward, I think. But there is no cross-costing for the parts, correct? That is always something I find odd in RP-0, IIRC, that learning to make a 2m heathshield doesn't make learning how to make even larger heatshields easier.

ec429:
Yeah, effectively there's a "learn to make lunar heatshields" cost that's shared, then separate costs for each size.  It's not perfect, but it's simple, and the per-size costs are small enough that I think further sophistication is unnecessary.

Issue #571, opened 2016-07-24
Todo: Update CC for 0.15


NathanKell:
http://forum.kerbalspaceprogram.com/index.php?/topic/143244-nightingale-softworks-development-thread/&do=findComment&comment=2688588

leudaimon:
Looks really cool! 
@NathanKell, if you provide a more detailed todo list, I may try to do some of this, given the efforts I was doing in uncrewed exploration contracts (#566) will probably be simpler after the contracts are converted with the dependency structure in place.
Things that came to mind after reading the updates:
- The DATA_EXPAND nodes are very promising in providing contracts for all planets/bodies, instead of the long lists currently implemented. With the flyby contracts as a "backbone", all other contracts can be implemented with the dependencies increasing from the respective flyby.
- How does this system deal with randomness in the contracts? Especially in cases that some random factor creates very different contracts.

NathanKell:
Awesome, thanks! <3

Per the post, looks like the best bet is to load up with the new CC build, and see what looks odd. Then we can take it from there?

leudaimon:
I had a look at the logs from CC after changing minVersion to 1.15.3 as suggested in the post. 
The warnings and errors abound, but most are minor:
- Almost all contracts have a weight attributed which is deprecated, but this is minor and doesn't affect anything. 
- Many of the other issues are also minor, related to genericTitle and genericDescription fields. Should be an easy fix.
- Some contracts have other issues that will require further investigation.

I'm posting the logs for contracts with issues (except the ones with only weight present as issue) in the following link:
https://www.dropbox.com/sh/pvqlkvvykpdy4s8/AAAzcqD5IrnDnrQJ1nwQXVOWa?dl=0 That's unfortunate that log files are not provided for group warnings. There is some useful stuff there in the debug logs ingame, so I'm providing an example screenshot along with the contract logs.

Besides, some stuff to take into account to use the new system better:
- Nightingale apparently strongly suggests using agents for different groups
- contract groups are now explicit to users, should they be made more meaningful?
- I don't think MaxSimultaneous for groups makes sense anymore, one more reason to rework groups.

NathanKell:
* We can remove weight, yep! Hallelujah!
* Great!

I'll look through the archive tonight (I hope).

* We kinda only want one agent for many different groups, though. FAI should be the agent for milestones, records, and first flybies (which are milestones but not in the milestones group), landings ditto)...maybe even more. That causes problems.
[...]

PR #605, opened 2017-01-16
1.2 update

rsparkyc:
I removed a bunch of the electric charge usage code, because 1) I couldn't figure out the new code that would be needed, and 2) I think this may be supported in stock.

I'm sure there's more work to do here, but wanted to at least get something started.


rsparkyc:
Also created this PR for kerbal renamer (a RP-0 dependency) https://github.com/KSP-RO/KerbalRenamer/pull/2/files

leudaimon:
I believe the changes to make use of the major update of CC must be finished before RP-0 release. I started looking into it with help from @NathanKell at #571 and #573, but didn't get the time to work on this properly. Given I found the stuff for configuring contracts more complicated than I expected, and probably beyond my abilities, if someone is up to take this over, I would be very thankful.

rsparkyc:
Ok, awesome, that gives me some stuff to look into.

rsparkyc:
Ok @leudaimon, those CC updates have been put here: https://github.com/KSP-RO/RP-0/pull/606/commits

PR #606, opened 2017-01-17
1.22.2 CC update

rsparkyc:
This PR is in response to https://github.com/KSP-RO/RP-0/issues/571.  It should be merged only after https://github.com/KSP-RO/RP-0/pull/605 has been reviewed and merged, as it is based off that codebase (since this version of CC is meant for KSP 1.2.2).  It replaces https://github.com/KSP-RO/RP-0/pull/573, as it has that code merged in, as well as the other changes required to get the contracts to load.


rsparkyc:
There are potential changes once there is a response to this: http://forum.kerbalspaceprogram.com/index.php?/topic/91625-12x-contract-configurator-v1222-2016-12-16/&do=findComment&comment=2924303 as I don't like having the minVersion referenced in a bunch of places (but leaving it off produces warnings)

PR #611, opened 2017-02-07
K2 Command Pod RP-0 support

PhineasFreak:
Change log:

* Add RP-0 support for the [K2 Command Pod](http://forum.kerbalspaceprogram.com/index.php?showtopic=155013) (Gemini style).

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -2400,6 +2400,11 @@ simpleCommandModules: # Second gen capsules
     
     rn_zond_les: #zond launch escape system
 
+    # K2 Command Pod (Gemini - alike).
+    K2Pod:
+        entryCost: 322000
+        cost: 6120
+
 electrics:
     # Solar: Level 3 static, Level 2 moving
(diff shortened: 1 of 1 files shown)


leudaimon:
Cool! Any reason for this to have a different pricing than the FASA and Tantares Gemini?

PhineasFreak:
The K2 includes the Reentry Control System in the pod structure. In FASA and Trails these parts are separate. So:

* RCS (cost/EntryCost): 720/36000
* Pod (cost/EntryCost): 5400/286000

with the sums of these two to be 6120/322000. These values should probably be tweaked a bit since it is not a Gemini CM per se.

leudaimon:
Oh, ok! I tried to find stuff that could be included in the K2, but was looking for parachutes and this kind of stuff.

stratochief66:
Nice! A non-FASA, non-Russian 2 crew part.

PR #612, opened 2017-02-09
Add RP-0 support for the RealEngines pack

PhineasFreak:
Change log:

* Add support for the RealEngines pack.

**Note:** this PR will conflict at least with PR #600. Do not be alarmed, i am aware of it and it will be fixed in due time.

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -1908,11 +1908,18 @@ advRocketry:
     RD0110: &RD-0110
         cost: 400 # guess based on other engines of the era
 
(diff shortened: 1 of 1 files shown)


stratochief66:
What do you mean, handling everything related to a giant mod as a single YAML file has downsides? Whatchutalkinbout?

PhineasFreak:
@stratochief66 I think that i fixed it. The extra added files make me a bit nervous but whatever...

PR #622, opened 2017-03-02
Uncrewed (but with capacity) modules can now be shut down

rsparkyc:
line endings got weird, so you can look at this: https://github.com/KSP-RO/RP-0/pull/622/files?w=1


SirKeplan:
Ahhh, the troubles of multi platform line ending chars :P. seems like CR LF is getting replaced with simply LF, which is odd because most files just use simply LF already.

only possible problem is losing the ability to `git blame`

PR #624, opened 2017-03-10
Procedural avionics

rsparkyc:
Needs plenty of testing.  Checkout the changes to the README.md, and let me know if we should move that somewhere else.  Tech Nodes have been built out to the Ranger Block I Core, instructions are in the readme for creating new ones.  Also, this spreadsheet helps:
https://docs.google.com/spreadsheets/d/1lpuszH_nUbWlTXZzfOMzsy-uhUNh016eNu0WL8x6qvU/edit#gid=1101913381


README.md
```diff
@@ -76,3 +76,152 @@ We recommend installing via the CKAN, `ckan.exe install RP-0` on the command lin
...
+
+###Configuration
+Procedural Avionics can have different configurations.  For instance, booster avionics might not be toggleable, use a lot of power, but be fairly light (for the amount of mass they can control), while probeCore avionics would be more efficient power wise, but a bit heavier (due to their robustness).  Upper stages might lie somewhere in the middle.
+
+###Optimal Utilization Percentages
+
+As part utilization goes up, so do part cost and part weight.  However, these do not rise at the same rates.  Consider this example: as you try to fit more avionics controls into the same amount of space, you should probably invest some effort into making those controls smaller and more efficient.  The smaller and more efficient you try to make those parts, the more they cost (and the higher the cost per controllable ton).  However, since these parts are more efficient, while your overall part mass continues to rise, your mass per controllable ton actually decreases.  Because of this, we actually get two different "curves", a mass curve, and a cost curve:
+
+ - mass curve: -x^2 + 2x  (as utilization goes up, additional mass needed per controllable mass goes down)
```
rsparkyc: Is there any way to make these formulas show up better in markdown?

leudaimon:
Great job!
About the probe cores... I think we may have a couple of issues with the early tiers. 
1) Up to stability, all probe cores provide no avionics. Is this going to make into the procedural avionics in any way? I don't think we should have avionics in cores at these tiers, but then a probe core procedural part would have a minimum mass and energy consumption just to have a core similar to the sounding rocket core from start node and the RP0probeVanguardXray from early avionics. Or can we just live with the already implemented parts, and introduce probe cores in the stability node?
2) In stability, the probeCoreOcto2 is actually more efficient than the ranger1, which comes in a later tier (50kg to control 200kg vs 80kg to control 300kg). I think the efficiency at the ranger1 tier could be buffed, to be an actual improvement over the Octo.

In a side note, I think we should think about some way to prevent upper stage avionics from being used as a probe core. It is very easy to abuse it already, and being able to make a small upper stage avionics breaks it completely. Would it be possible to make their batteries non-rechargeable?

rsparkyc:
Thanks for the feedback @leudaimon 
As for 1) I only put probeCore tech nodes in at those upper tech levels, though I haven't tried to use a "scaled down procedural booster" as an upper stage early probe core.  One solution may be to have a minimum controllable mass on the parts, which would force you to lug a heavy probe core up there in the early game.
2) I must have missed the probeCoreOcto2 node when building up the tech tree, but whatever we can do to make sure this stays balanced sounds good to me.

SirKeplan:
Apart from the very earliest parts for sounding rockets, proc avionics should have  a `ModuleSAS` how do we deal with that?

leudaimon:
Cool @rsparkyc, I believe just leaving the non-procedural parts for the early cores should work very fine.

About 2), thats at line 1632 of tree.yml. Ingame, I think it's named early controllable core. I always found it very underwhelming to arrive at the ranger core tier and see it was worse than the one I already had.

rsparkyc:
@SirKeplan I'm thinking that we should not use proc avionics for anything that you would not expect to be controllable, so all proc avionics should have a `ModuleSAS`

SirKeplan:
@rsparkyc that makes sense, only thing is setting an `SASServiceLevel`, that is less of an issue though.

rsparkyc:
I'm thinking we should add an optional minimum/maximum controllable tonnage to the config nodes to prevent people from using scaled down booster avionics as upper/probe core avionics units.

leudaimon:
Limiting tonnage goes a little against the procedural stuff spirit, but I don't see a different solution to keep people from using reduced upper stages as probe cores. One solution I think would be more elegant, I don't think is possible: making batteries on upper stages and boosters non-rechargeable. That would limit the applicability of those avionics as long-term probe cores, but I don't think it would be feasible.
One more suggestion: I don't think a limit is necessary for probe avionics, as those should be the least efficient in a given tech level, and thus not adequate for other applications.

rsparkyc:
@leudaimon Good call on the probe avionics, I've removed the limitation there.  Hopefully having those limitations on the other nodes don't dampen the procedural spirit too much, and mainly just limit abuse of them.

BevoLJ:
No idea if this would complicate or simplify, but what about two different types of avionics.  Rad/non-Rad hardened avionics.  Non-rad hardened avionics could be limited by MET.  Like disable non-rad hardened avionics after MET > 604,800 (7 days).

This would allow more freedom in tier balancing of avionics for gameplay.

rsparkyc:
At this point I think I'm ready to have this merged.  I was considering adding in the "Show GUI" option that RealFuels uses, but at this point that seems to be a bit of an undertaking (though I'm not opposed to letting someone else take than on).

@SirKeplan, per your suggestion, I added a ModuleSAS, and a configurable SASServiceLevel to all the TECHNODEs

@stratochief66 @leudaimon @SirKeplan @NathanKell, thoughts on merging this?

BevoLJ:
Might it be possible to have the power requirement show on the tool tip scale between watts and kilowatts?  For small upperStage ones it shows as 0 kw / 0 kw, although I know it is drawing power.  Just not sure how much.

PR #627, opened 2017-03-26
Fix Mercury Flyby Contract

BevoLJ:
Fixed the requirements as they weren't preventing the contract from being available prior to Venus (or even lunar) flyby completion.

Also increased rewards because small bodies next to huge Sun's are far harder than say Venus.  I set rewards equal to Saturn, but should probably be more inline with Uranus or Neptune.


rsparkyc:
This sounds reasonable to me.

leudaimon:
In theory you could build a probe for a "tour" to Venus and then Mercury using a gravity assist for the first flyby of both planets... However, I agree that it is at least surprising to have such a mission before a regular Venus flyby. Anyway, just food for thought... I think the change is indeed reasonable.

PR #632, opened 2017-03-28
Removed the NewVessel requirement from the flyby contracts

rsparkyc:
@stratochief66 @NathanKell see any problems with this?  By doing this, I could setup a Jupiter flyby, and then if I see that I could also swing by Saturn, I could accept that contract later and use my Jupiter probe.

### DIFF
--- a/GameData/RP-0/Contracts/Fly-by Contracts/CallistoFlyByContract.cfg
+++ b/GameData/RP-0/Contracts/Fly-by Contracts/CallistoFlyByContract.cfg
@@ -44,12 +44,6 @@ CONTRACT_TYPE
 		
 		PARAMETER
 		{
-			name = NewVessel
-			type = NewVessel
-		}
-	
-		PARAMETER
-		{
 			name = ReachState
 			type = ReachState
 			targetBody = Callisto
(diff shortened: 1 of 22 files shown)


leudaimon:
afaik the newvessel requirement is in place to prevent people from taking the contract when it is already set to complete, just before the encounter for example. However, in this specific case, I think the tour-like nature of most flyby missions is more important than preventing this exploit.

hargn:
Maybe the requirement should be based on these two criterias :
- New Vessel
OR
- Vessel belongs to the group of vessels that have completed a flyby mission (I think that it is not too hard to implement into CC)

alimoncul:
Maybe 2 type of contracts 1 for reusing a satellite with gravity assists (lower reward) 1 for new launch for satellite (higher reward)

rsparkyc:
I feel like the first "flyby contracts" probably wouldn't even need that limitation.  I think they're meant to reward the first time you visit a planet/moon.  It would be one thing if it was a contract that was meant to establish a permanent orbit (where you could "cheat" by using the same craft to orbit a bunch of moons), but I don't think that's the case here.

PR #648, opened 2017-05-01
Atlas only depends on basicAvionics

Schnobs:
Unlocking Basic Avionics will enable the integrated avionics on all Atlas tanks.

### DIFF
--- a/GameData/RP-0/Avionics.cfg
+++ b/GameData/RP-0/Avionics.cfg
@@ -449,7 +449,7 @@
 	{
 		name = ModuleAvionics
 		massLimit = 125.0
-		techRequired = flightControl
+		techRequired = basicAvionics
 	}
 }
 // Atlas F-mod (appears identical to E/F)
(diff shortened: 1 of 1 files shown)


stratochief66:
Strange; I searched some of these parts in the RP-0 tree.yaml and expected to find that the tech node that unlocks them is/was also the node that gave them avionics, but it wasn't.

I'm not sure what the logic was for some of these original placements. Since the parts aren't accessible until they are unlocked in the tech tree anyway, I think simply having the avionics connected to an avionics tech as you've suggested here makes sense.

Any second opinion to merge? @rsparkyc or @SirKeplan ?

Schnobs:
The first Atlas comes relatively early on the tech tree, "Construction" for 10 points. It used to be possible that one could just research that node and immediately start launching Atlas rockets without researching any avionics at all.
As I recall it, the option to make integrated avionics dependent on tech unlocks was supposed to close this loophole.

Basic Avionics provides the generic guidance systems for R-7 and Atlas class rockets; it really think that should suffice to unlock the builtin avionics of all the relevant tanks, whether they hold 98, 100 or even 125m³ of fuel, and even if their historic first flight was in 1970.

Issue #523, opened 2016-05-10
RF Tech Levels


batuhantasli:
It is possible to choose any type of tank type from the start.

NathanKell:
Nope. Construction nodes unlock new tank types.

batuhantasli:
I only unlocked first construction node in my RP-0 career and even cryogenic was on the list. Isn't it unlocked in a late node?

Zarbizaure:
the cryogenic tank type is unlocked later, in the hydrolox TL2 node I think (but i'm not sure).
Also, that's something PP should show in the tech tree, a bit like Mechjeb: just display the standard procedural tank and explain the upgrade in the description.

leudaimon:
Per RP-0/ResourceTechs.cfg:
Default -> Early construction
Service Module/Balloon -> Basic construction
Cryogenic/Balloon Cryo -> Adv construction
If you have anything different, you may have a problem in your install. 
It would be nice if this was shown along with the diameter upgrades in the tech tree - is it possible?

Temeter:
I had at some point the issue where RealFuels would show me all tanks, but I only could use cryogenic. Maybe that's your problem?

pap1723:
I can confidently say that the proper tanks are unlocking in the proper nodes, however, I think we can use the new PARTUPGRADE feature to create the icons for these so they are shown in the proper nodes.

PR #657, opened 2017-05-24
Created PARTUPGRADES for the RF Fuel Tank Types

pap1723:
In reference to #523 I decided to create very simple PARTUPGRADES to show up in the Tech Tree for each of the different types of fuel tanks made available by RealFuels.

I have used stock Icons so that they will work on any install and given them no costs to unlock (because they don't actually do anything, they are just informative).

Here they are in action:
http://imgur.com/a/6yfvB

### DIFF
--- a/GameData/RP-0/ResourceTechs.cfg
+++ b/GameData/RP-0/ResourceTechs.cfg
@@ -52,3 +52,94 @@
 {
 	%techRequired = survivability
 }
+
+PARTUPGRADE
+{
+	name = RF-cryogenic-tank-unlock
+	partIcon = fuelTank3-2
+	techRequired = advConstruction
+	entryCost = 0
+	cost = 0	
+	title = Cryogenic Tank Type Unlock
+	basicInfo = You can now choose the tank type Cryogenic
+	manufacturer = Generic
+	description = You can now choose the tank type Cryogenic
+}
+
+PARTUPGRADE
+{
+	name = RF-balloon-cryogenic-tank-unlock
+	partIcon = fuelTank3-2
+	techRequired = advConstruction
+	entryCost = 0
+	cost = 0	
+	title = Balloon Cryogenic Tank Type Unlock
+	basicInfo = You can now choose the tank type Balloon Cryogenic
+	manufacturer = Generic
+	description = You can now choose the tank type Balloon Cryogenic
+}
+
+PARTUPGRADE
+{
+	name = RF-default-tank-unlock
+	partIcon = fuelTank_long
+	techRequired = engineering101
+	entryCost = 0
+	cost = 0	
+	title = Default Tank Type Unlock
+	basicInfo = You can now choose the tank type Default
+	manufacturer = Generic
+	description = You can now choose the tank type Default
+}
+
+PARTUPGRADE
+{
+	name = RF-service-module-tank-unlock
+	partIcon = Size3MediumTank
+	techRequired = basicConstruction
+	entryCost = 0
+	cost = 0	
+	title = Service Module Tank Type Unlock
+	basicInfo = You can now choose the tank type Service Module
+	manufacturer = Generic
+	description = You can now choose the tank type Service Module
+}
+
+PARTUPGRADE
+{
+	name = RF-balloon-tank-unlock
+	partIcon = Size3LargeTank
+	techRequired = basicConstruction
+	entryCost = 0
+	cost = 0	
+	title = Balloon Tank Type Unlock
+	basicInfo = You can now choose the tank type Balloon
+	manufacturer = Generic
+	description = You can now choose the tank type Balloon
+}
+
+PARTUPGRADE
+{
+	name = RF-electric-propulsion-tank-unlock
+	partIcon = xenonTank
+	techRequired = electronics
+	entryCost = 0
+	cost = 0	
+	title = Electric Propulsion Tank Type Unlock
+	basicInfo = You can now choose the tank type Electric Propulsion
+	manufacturer = Generic
+	description = You can now choose the tank type Electric Propulsion
+}
+
+PARTUPGRADE
+{
+	name = RF-life-support-tank-unlock
+	partIcon = miniFuelTank
+	techRequired = survivability
+	entryCost = 0
+	cost = 0	
+	title = Life Support Tank Type Unlock
+	basicInfo = You can now choose the tank type Life Support
+	manufacturer = Generic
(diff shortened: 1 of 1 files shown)


PhineasFreak:
I imagine that we can replace the stock tank part names with the ones from procedural parts, right? Asking only out of curiosity.

pap1723:
Yes, basically any valid part name will work. In the future, it will be
VERY useful to include the diameter increases for Procedurals as well.

PhineasFreak:
Isn't that possible with the above system? I mean, these are just dummy parts in the tech tree, i imagine that you can add more with different information, right?

NathanKell:
It would be good to get rid of all dummy parts, that's what these are meant
to replace

pap1723:
@PhineasFreak Yes, I am sorry, I wasn't clear in what I meant. This is absolutely the easiest way to show when the Diameter increases will happen in the tech tree. The SSTU parts are currently setup this way.

I will be creating them for the Procedural Parts as well as the Tech Level increases for RCS as well.

pap1723:
OK, in reference to the above, there is this bit of code and I want to make sure I understand what the description / title should be to make sense to the player...

    @RF_TECHLEVELS
    {
        @ENGINETYPE,*:HAS[~name[N]]
        {
            %TLTECH0 = basicRocketry
            %TLTECH1 = generalRocketry
            %TLTECH2 = advRocketry
            %TLTECH3 = heavyRocketry
            %TLTECH4 = heavierRocketry
            %TLTECH5 = veryHeavyRocketry
            %TLTECH6 = experimentalRocketry
            %TLTECH7 = giganticRocketry
        }
        @ENGINETYPE,*:HAS[#name[S*]]
        {
            %TLTECH0 = basicRocketry
            %TLTECH1 = basicSolids
            %TLTECH2 = matureSolids
            %TLTECH3 = advFuelSystems
            %TLTECH4 = largeVolumeContainment
            %TLTECH5 = highPerformanceFuelSystems
            %TLTECH6 = specializedFuelStorage
            %TLTECH7 = exoticFuelStorage
        }
    }

NathanKell:
That sets where each TL for generic engines (RCS, x kN Thruster, proc
solids, etc) unlocks.

The top one is saying all non-nuclear generics, the bottom for solids.

leudaimon:
Given the increase in thrust and Isp will depend on the specific engine and fuel, I guess you can just state that Isp and thrust of RCS (or solids in the second case) were uprated, without further detail.

As we are mentioning things hidden in the tech tree... isn't it possible to show alternate configs of engines (including here the different fuels that get available for RCS) in the tech tree somehow?

pap1723:
The original thing that made me think of this is with the different Engine
Configs that get unlocked, but here is where it gets tricky...

You need to have a part to point to in order to create an icon. As we all
know, almost all of the engines are implemented by mods. Trying to create
the icons required will have to be done at a mod level for each of the
engine configs you want to show. That is the more complicated part of it.

NathanKell:
Nope just have a single generic engine part, perhaps added by RO itself
(someone promised to make us one), make it category -1 and under ORPHANS so
it won't show up, and use that as the config unlock icon

pap1723:
Yep @NathanKell that will work very well.

The only other issue I see with it (don't know how to get around it) is that whether you have a mod installed to include the engine or not, the Engine Config PARTUPGRADE tech unlock will show in the tree.

Is there a way around that? Otherwise, it becomes much more confusing than if we didn't add them.

NathanKell:
Use :NEEDS[] on the part upgrade?

pap1723:
That is what I thought as well with a NEEDS[]

Whoever creates them will just have to find all of the mods that include each engine.

NathanKell:
You might be able to do some magic with external node editing and placing
the patch in the engine global config

pap1723:
I think you meant that YOU (NathanKell) might be able to do that because I only understood some of your words. ;)

NathanKell:
Hehe. I'll see if I can come up with something

On May 24, 2017 1:02 PM, "Pap" <notifications@github.com> wrote:

I think you meant that YOU (NathanKell) might be able to do that because I
only understood some of your words. ;)

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/KSP-RO/RP-0/pull/657#issuecomment-303835456>, or mute
the thread

NathanKell:
@sarbian we can write to external things yeah? Idea is in a foo[] {} we
have %root/PARTUPGRADE[bar] {} to create that PARTUPGRADE (using % because
that patch may run many times) and then somewhere else we do
@PARTUPGRADE[bar]:FOR[later pass] and actually fill it in.

Does something like that work?

sarbian:
Yes you can do that. I nearly forgot that I added it...
http://forum.kerbalspaceprogram.com/index.php?/topic/50533-130-module-manager-280-may-26th-2017-better-late-than-never/&do=findComment&comment=2413546

pianojosh:
Comment from the cheap seats: Might this be a good place to describe what the tank types are? I don't think there's any in-game explanation of them.

Cryogenic: "Additional insulation when you need your cryogenic liquids to last days rather than hours. Useful if you need longer than usual coasts for departure stages or Cryogenic Mun/Minmus/Moon insertion and landers. For boosters and most departure stages, stick with Default and add a bit extra of your colder liquid(s), the mass penalty of Cryogenic isn't necessary for them.

ServiceModule: "Highly pressurized and heavily insulated. Use if you need cryogenics to last for more than a few days, such as fuel cell tanks"

Balloon/BallonCryo: "Much lighter than Default/Cryogenic, but at a higher cost. Used by Atlas and Centaur in real life. Rarely used in the real world due to complicated ground handling and structural weakness."

Maybe not necessary, exactly, but as long as all that description room is there, could put it to good use.

NathanKell:
@sarbian thanks, I'll have a poke when I get home.

@pianojosh that's an excellent point! We never do describe them ingame.

NathanKell:
Yep, ok, I think I have a way to do this.

1. in a normal (non-patching) cfg, create the PARTUPGRADE and have a count = 0 line.
2. In the engine global configs, use that linked syntax to increment count in that PARTUPGRADE. That means if the global config is applied once, count will be 1, if twice, count will be 2, etc.
3. Delete all PARTUPGRADEs with count == 0

Presto!

NathanKell:
See https://github.com/KSP-RO/RealismOverhaul/pull/1654

PR #659, opened 2017-05-27
Update netkan

Olympic1:
Changes to RP-0 netkan:
* Moved `KerbalConstructionTime` to dependencies
* Renamed `SXT` to `SXTContinued`
* Removed `TACLS-Config-RealismOverhaul` (config now uses MM)
* Removed overrides as these are an one-time use

**Merge after release**


stratochief66:
*I* read the fine print. But beforehand, I also nearly made that mistake. :P

PR #661, opened 2017-05-28
Added Bluedog Design Bureau parts to tech tree.

Jerre122:
Added BDB fasa clamps and towers, and BDB diamant rocket.
Added .vs directory to .gitignore


tree.yml
```diff
@@ -1393,7 +1430,7 @@ basicConstruction:
     r7_blok_bvgd_sput: &R7Booster
         cost: 100
     r7_blok_a_sput:
-        cost: 120	
+        cost: 120
```
stratochief66: @NathanKell 
will trailing changes like this have any effect, or no?

stratochief66:
Tentatively, looks good to me.

tree.yml
NathanKell: I don't think so. Travis would complain if the tree didn't compile.

pap1723:
I know this is an old PR, but with the Tree.yml being deprecated, and the old tree not being used anymore, this no longer applies. All BDB parts should be placed correctly in the new tree already.

PR #684, opened 2017-06-10
clearer that it's OK to transmit some of the scien

glilienthal:
so far it seems like any transmission would be harmful for completion.


NathanKell:
Merged to master.

PR #689, opened 2017-06-12
J-2T price

winged7:
Price of J-2 was lowered almost 4x but it seems like someone has forgotten about J-2T.


pap1723:
Tree.yml is deprecated, but these changes have been made with the new tree. There is a good chance that these will change as all pricing for engines has been changing a lot to coincide with new Launch Costs.

PR #693, opened 2017-06-14
Vulcain, LE-7  - price

winged7:
Both are ridiculously expensive compared to their performance. Look at RS-68 as a comparision.


pap1723:
I will have more comments on this later, but for now, I agree with you, however, I think that the costs that you have come up with might be a little low.

pap1723:
These changes have been included, however as the prices are changing on everything, these will probably be included in that as well.

PR #697, opened 2017-06-17
fix minPeA after completion

glilienthal:
in the current version, minPeA=400km, after the first crew departed, even though the text says 250k. I think that's a typo and set it to 250k.


pap1723:
Yep, this is a good fix, thanks!

PR #700, opened 2017-06-19
Fix LR91 ECMs.

ahmedcharles:


rsparkyc:
@NathanKell I know you fixed some of these before, and you seem to be the expert on such things, want to take a look?

NathanKell:
So, the FASA large LR91 skips the first config, and the miniLR91 only has the first config (IIRC). That's why they use subtractor not the usual multiplier. What was the issue encountered?

ahmedcharles:
At a high level, the ECM configs don't mention the FASAGeminiLR91Mini at all. It seems to have the same stats as the FASAGeminiLR91, but I didn't look at the engine configs themselves. I assumed they were the same, so I figured they should work the same as other parts that are the same engine with different looks.

NathanKell:
The ECMs for it were written when the -3 was only on the Mini, and the -5 and above were only on the regular. THat's why (as you can see in the original file) the Mini counts as a subtractor for the others, but not vice versa.

If both the mini and the regular now have all the same config options, then they should be treated identically and there should be three ECMs (one per part), each of which has multiplier 0 for the other two copies.

ahmedcharles:
As far as I can tell, the Mini and the regular have the same engine configs, because they both have engine = LR91 and the configs are the same there. So, I assume that means that this change is correct?

ahmedcharles:
@NathanKell Does this look good?

NathanKell:
Yup!

Issue #701, opened 2017-06-19
Procedural Avionics in the new tech tree


leudaimon:
I have devised a simple progression scheme for the procedural avionics modules in the new RP-0 tech tree. Instead of going the more accurate route of copying the performance of non-procedurals at all tech levels, I eyeballed the pattern of progressing for the different types of avionics and devised simple formulas for that. It will very likely demand some tweaking after first implementation, but I think it is a simple first step. The values I intend to use, in @rsparkyc 's spreadsheet formtat, is available [here](https://docs.google.com/spreadsheets/d/1MOYYTKPtr3TgfFHrgTUNjMn86nmjmYlchaZmIc3BFEY/edit#gid=1101913381).
 @pap1723 @rsparkyc @NathanKell and other people that is looking into the new tree, if you want to give suggestions before I make the first PR for that, please feel free.

rsparkyc:
Something that's worth adding is having a column for the cost of a science return unit.  I'm guessing this should not be scaled according to size.  The question is do we just leave that a fixed price, or do we lower the price as the technology gets better?  With payouts getting bigger, i don't know if lowering the cost is really an issue.

leudaimon:
@rsparkyc I think it can have a fixed price, just as any other experiment. Btw, one thing some people mentioned as suggestions for procedural probes is adding experiments. That could be nice, an option to add tech-unlocked experiments, with a cost similar to the standalone parts...

Bornholio:
close me?

PR #703, opened 2017-06-19
Proc avionics tech levels, per #701

leudaimon:
I still have some issues with a few details, especially unlock costs. Suggestions are welcome.


NathanKell:
@leudaimon please retarget on Master.

PR #715, opened 2017-07-07
Fixes to procedural avionics

leudaimon:
When costs were defined, I thought they were directly applied, but the costs of previous TLs are discounted from the values in the configs. Fixed that to have total cost in each TL.


pap1723:
There was a merge issue because I added the entries for the Tech Tree. There was also a missing } that didn't allow Probe Avionics to unlock. I am closing this, but your new stats are fully integrated into the tree.

PR #720, opened 2017-07-14
Add HG-3

Ash19256:
Using a 27% price increase estimate for the cost of the the HG-3 over the J-2, directly extrapolating from the thrust increase. This is just a placeholder in the extreme, and if @pap1723 is already working on something like this or something then this can just get closed. But I at least wanted to get the ball rolling. I figured that since T5 is where the RS-25 first pops up, and T3 is the home of the J-2, T4 worked well enough as a place for this engine. That being said, I could be doing this wrong, as this is my first RP-0 pull request, so feel free to call me out on stupid mistakes.

### DIFF
--- a/tree.yml
+++ b/tree.yml
@@ -3486,6 +3486,12 @@ hydroloxTL4:
         cost: 9000 # J-2 is 8500
     RLA_mp_large_spike: *J2T 
     NP_lfe_125m_AerospikeEngine: *J2T
+    
+    SXTLT80: &HG-3
+        cost: 3175
+        entryCost: 127000
+    SHIP_HG_3_SL: *HG-3
+    SHIP_HG_3_VAC: *HG-3
         
 precisionPropulsion:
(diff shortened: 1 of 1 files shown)


pap1723:
This has been added to the new Tech Tree, @NathanKell do we want to add it to the old one as well?

NathanKell:
I wouldn't bother, it's not like we're going to release or ever touch it again :]

Ash19256:
Out of curiosity, how would you go about adding it to an existing RP-0 install? Cosmonaut Crash on YouTube is currently using them in his RP-0 series, and I was kind of hoping to at least let him not have to deal with the "non RP-0" tags on all of the HG-3s.

NathanKell:
@Ash19256 you can copy and paste an existing entry from Tree.cfg (in a release of RP-0) and change the name and data, then send that to Cosmonaut Crash. :)

RP-0 no longer uses tree.yml so that's the best way to fix things up in a release (existing) version of RP-0.

PR #725, opened 2017-07-17
Entry Cost configs

smartdummies:
Added entry cost mosifiers for AJ10 early and mid and the Vanguard X405.
Also chnage the entry cost for the RN X405 Vernier so that the total
entry cost for the RN Vanguard X405 matches the SXT composite


NathanKell:
LGTM!

PR #730, opened 2017-07-20
Engine identicalPArts and Entry Cost Modifiers

smartdummies:
Updates to the IdenticalParts and EntryCostModifiers, moslty AJ10s

### DIFF
--- a/GameData/RP-0/Tree/EntryCostModifiers.cfg
+++ b/GameData/RP-0/Tree/EntryCostModifiers.cfg
@@ -48,7 +48,7 @@ ENTRYCOSTMODS
 		{
 			SSTU-AJ10-CustomEarly = 0
 			rn-aj10-37 = 0
-			bluedog-DeltaK-AJ10 = 0
+			bluedog_ableEngine = 0
 		}
 	}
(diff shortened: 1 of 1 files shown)


NathanKell:
Um, is the bluedog-DeltaK-AJ10 not actually the Delta K one? Because Delta K's AJ10 was like the last AJ10, so it'd be AJ10-Adv.

NathanKell:
Poke

smartdummies:
Ok let me go back and take a look

smartdummies:
Doh, bad copy past error.

smartdummies:
Did my updates get posted to this or do I need to do a new pull request?

NathanKell:
Ah dang, conflicts. Hmm, @pap1723 can you import the identicalpart changes into your xls?

pap1723:
Let me review later. I'll roll the changes in.

pap1723:
The Identical Parts changes have been added for the release. The Entry Cost Modifiers are no longer working the way that they were before. The whole system has been re-written so these ECM's will be applied, but in a completely different way.

PR #732, opened 2017-07-22
Price some decouplers/bases/nosecones.

ahmedcharles:
The prices are based primarily on similar parts or the old tech tree prices.

### DIFF
--- a/GameData/RP-0/Tree/TREE-Parts.cfg
+++ b/GameData/RP-0/Tree/TREE-Parts.cfg
@@ -13097,16 +13097,16 @@
 @PART[rn_sputnik3_base]:FOR[xxxRP0]
 {
     %TechRequired = materialsScienceSatellite
-    %cost = 200
-    %entryCost = 500
+    %cost = 10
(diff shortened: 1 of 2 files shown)


ahmedcharles:
I'm not listing them as being RP-0 configured because I suspect having the entry cost be 1 for everything will be contentious and all I really want to do is improve the costs of parts that aren't currently configured rather than spend lots of time debating. Hopefully that means this can just get merged and I won't have to keep local changes that do this.

NathanKell:
I believe these changes should be made to @pap1723 's xls file?

ahmedcharles:
How do I do that?

pap1723:
These are included with the new developmental updates. HOWEVER, there is no pricing adjustments for the fuel tanks. NathanKell has changed the costs of the PP Tanks to be more in line for the Launch Cost major changes that are coming. Therefore, these fuel tanks should be priced by using PP Tanks and determining their cost. That will have to come at a later time when these changes are fully complete.

PR #734, opened 2017-07-24
Fix config node unlocks for RN's R-7 engines.

ahmedcharles:
This doesn't work yet, but I'm not sure what else could be going wrong.

### DIFF
--- a/GameData/RP-0/Tree/TREE-Engine_Configs.cfg
+++ b/GameData/RP-0/Tree/TREE-Engine_Configs.cfg
@@ -974,6 +974,34 @@
         { %techRequired = orbitalRocketry1972 }
     }
 }
+@PART[rn_r7_bvgd_engine]:FOR[xxxRP0]
+{
+    @MODULE[ModuleEngineConfigs]
+    {
+        @CONFIG[RD-107-8D74]
+        { %techRequired = orbitalRocketry1956 }
+        @CONFIG[RD-107-8D74PS]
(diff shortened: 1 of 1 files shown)


ahmedcharles:
This makes it so that the configs unlock in the correct nodes.

ahmedcharles:
Any feedback on this?

NathanKell:
This...doesn't seem to have anything to do with TF...?

Except to the extent that it looks like those configs are RN-specific, so I wouldn't be surprised if they have no TF configs because they're not using the global configs (where the TF configging is done).

ahmedcharles:
On, I just didn't change the comment.

ahmedcharles:
That's a better commit comment.

NathanKell:
Ah, that makes more sense. I wonder if we now have enough global configs that we can harmonize with raidernick's configs?

ahmedcharles:
All of the RN TF configs map to the stats of an existing TF config, i.e. he didn't create different reliability profiles from the 'global' ones.

However, the engine configs themselves are different, since they lack the thrust of the verniers, since they are separate engines. Granted, this also means that RN's version of the LV is more likely to have a failure, since there are more engines being modeled.

ahmedcharles:
Anyways, any unification would have to be done in RO, not RP-0, so this should be good as is?

pap1723:
This is taken care of with the new RD-107/RD-108 configs

PR #737, opened 2017-07-31
remove old Tree.cfg from .gitignore

lamont-granquist:
i believe this is cruft now and the old .cfg file should be removed
from any builds now, and if its present the git repo should properly
be marked as dirty?

Signed-off-by: Lamont Granquist <lamont@scriptkiddie.org>


lamont-granquist:
eh, wut travis?

PR #744, opened 2017-08-08
Sounding Rocket fixes

ppboyle:
Fixed sounding altitude errors
Set sounding density to .5 g/cc
Made all contracts read density to make further changes easier.


ppboyle:
Fixed sounding altitude errors
Set sounding density to .5 g/cc
Made all contracts read density to make further changes easier.

PR #766, opened 2017-09-17
Don't remove Ablator from crew pods.  They're likely to need it.

ec429:
I also noticed HTP was being removed from the FASA Mercury pod, but the list
of possible RCS propellants is quite long; there must be a better way to fix
this than just listing them all.  (It'd surely be simpler to list all the
resources to _remove_...)


m4gus88:
Should probably also have CharredAblator.

PR #767, opened 2017-09-19
Update for KSP 1.3.1

aw1621107:
Don't really know what I'm doing, so hopefully this isn't too off.

Obviously still waiting on a ton of other dependencies.


Bornholio:
Running this successfully. Used current master as base, same for RO, RSS and a few other requirements.

PR #770, opened 2017-10-07
Add more crew reports

ntimpe:


winged7:
Mimas doesn't have InSpaceHigh biome.

PR #772, opened 2017-10-08
Add Surface Outpost Contracts

ntimpe:
This folder adds surface outpost contracts for the Moon and Mars, for both standard outposts and outposts with science labs.


pap1723:
It was always the plan to create these in the future. Great job @ntimpe !!

PR #778, opened 2017-10-16
Add more science cores

aw1621107:
There were some other cores that didn't have science cores nor something saying how much vessel mass the avionics support. I wasn't sure whether these should have avionics added:

  - Ranger Block 1 Core (bluedog_rangerCore)
  - Galileo Atmospheric Probe (galileo_aprobe)
  - Alouette Core (SXTAlouetteI)
  - Surveyor Core (ca.landv.core)
  - Mariner 1/2 Probe (rn_mariner1_2)
  - Tons of others

Perhaps an alternative would be mark everything that doesn't explicitly say how much vessel mass they can control with a science core? Something akin to
```
@PART[*]:HAS[!MODULE[ModuleAvionics],!MODULE[ModuleScienceCore]]:FOR[RP-0]
{
        MODULE
        {
                name = ModuleScienceCore
        }
}
```

Also, should crewed parts (both those that require crew and those that don't) have avionics? Existing parts don't appear to be consistent on that.


aw1621107:
Dangit, wrong branch. Let me try this again.

PR #783, opened 2017-11-03
Fix NewCrewStations.cfg

ctiberious:
As I read this, the point of this contract is to send crew to an empty station.  But the current requirement is set to only allow the contract to occur when there are less than 3 vacant seats available.  Meaning if a station can support 5 crew and you have 3 on the station, the contract would become available because "EmptyCrewSpace()" would return a result of 2.  By using "v.CrewCount()==0" we should only allow the contract to occur when a station has no crew on board, regardless of how many available seats there may be.

### DIFF
--- a/GameData/RP-0/Contracts/Space Stations/New Crew Station.cfg
+++ b/GameData/RP-0/Contracts/Space Stations/New Crew Station.cfg
@@ -53,8 +53,8 @@ CONTRACT_TYPE
 		requiredValue = true
 		hidden = true
 		uniquenessCheck = GROUP_ACTIVE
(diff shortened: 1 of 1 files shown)


pap1723:
So I might have to change the description instead. The idea of the contract is to send a new crew to an empty Space Station, or a station that has empty space in it. Basically the idea is to fill up the station with a new crew. This will allow better crew rotation missions in the future.

ec429:
@pap1723 well, in any case the constraint of needing fewer than 3 vacant seats is wrong; if you have an empty four-seat station you're not offered the contract.

PR #784, opened 2017-11-16
Show 3 sig figs for proc avionics tonnage slider

aw1621107:
Makes the displayed value consistent with the values shown in the
avionics tab of the RP-0 GUI


ec429:
Merge conflict appears to be caused by upstream move of Source/{Procedural Avionics=>Avionics}/ModuleProceduralAvionics.cs.  Suggest you rebase the fix.
(Change itself looks good to me, but I'm not confident enough to merge it without someone else also reviewing.)

aw1621107:
Alright, rebased. That's what I get for using the web editor without checking what branch I'm starting on...

PR #785, opened 2017-11-16
Allow RP-0 UI in Sandbox

aw1621107:
UI still provides useful info in sandbox mode, such as whether a vessel
has sufficient avionics

### DIFF
--- a/Source/UI/UIHolder.cs
+++ b/Source/UI/UIHolder.cs
@@ -39,8 +39,6 @@ namespace RP0
 
         private void OnGuiAppLauncherReady()
         {
-            if (HighLogic.CurrentGame.Mode == Game.Modes.SANDBOX)
-                return;
             try {
                 button = ApplicationLauncher.Instance.AddModApplication(
                     ShowWindow,


aw1621107:
Looks like I actually need to work on this a bit more; the RP-0 button for the UI shows up, but the window itself is a puny cross thing (in yellow rectangle in the screenshot), due to exceptions of some kind:

screen shot 2017-11-16 at 10 17 21 am

Exception:
```
[EXC 10:19:02.239] NullReferenceException: Object reference not set to an instance of an object
	RP0.TopWindow.DrawWindow (Int32 windowID)
	UnityEngine.GUILayout+LayoutedWindow.DoWindow (Int32 windowID)
	UnityEngine.GUI.CallWindowDelegate (UnityEngine.WindowFunction func, Int32 id, UnityEngine.GUISkin _skin, Int32 forceRect, Single width, Single height, UnityEngine.GUIStyle style)
```

ec429:
Ok, reverted.  (I naïvely assumed you'd tested it before submitting.)

aw1621107:
Not at all an unreasonable assumption. I thought the change would be simple. I'll open a new PR with a better set of changes

PR #786, opened 2017-11-16
Recompile to incorporate proc avionics sig fig changes

aw1621107:
Forgot that the web interface can't do this...

Compiled against Modular Flight Integrator v1.2.4


ec429:
Superseded by #803.

PR #791, opened 2017-11-18
Set max utilization for proc tanks

aw1621107:
Without this, the player can set the utilization for any
non-service-module proc tank to 100%, partially defeating the purpose of
the tank progression

### DIFF
--- a/GameData/RP-0/Parts/RFProcTanks.cfg
+++ b/GameData/RP-0/Parts/RFProcTanks.cfg
@@ -16,6 +16,7 @@
 		name = ModuleFuelTanks
 		volume = 1060
 		utilizationTweakable = true
+		maxUtilization = 88
 		utilization = 88
 		type = Tank-I
 		typeAvailable = Tank-I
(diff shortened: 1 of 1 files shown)


Theysen:
Wasn't utilization slider already in relation to a fixed ratio of volume vs rest of tank structure? So that a value of 100 actually doesn't cheat but is just a maximum value of used tank volume? 

I can't look at MFI at the moment to make sure. I hope my question makes sense to you.

aw1621107:
If I'm understanding you right, it make sense, but it seems to me that if that were the case then having the default utilizations for the tanks be below 100% would be somewhat pointless since the tank progression would already be "built in" to the 100% value, and having the default utilization be lower would just be a needless penalty. I'm not really familiar with the code, though, so I could have easily missed something

NathanKell:
Now that we actually have different parts for different tank types, we definitely should limit utilization, I agree. Only balloon gets (and requires) 100%. The maxes for the other tank types should be not much more than their new defaults, IMO.

PR #795, opened 2017-11-19
Allow distinct proc avionics parts to have different tech levels

aw1621107:
Prior to this PR, distinct proc avionics parts that had the same
config (i.e. "booster", "upper stage", or "probe") were forced to have
the same tech level, since the tech level was set per-config, instead of
per-part (or per-PartModule, in this case).


aw1621107:
Don't merge this yet; I may have broken KRASH sims at least

aw1621107:
Alright, I think it's fine now. Just a brain fart on my end.

Also, is it necessary to have both `CurrentProceduralAvionicsConfig` and `currentProceduralAvionicsConfig`?

PR #797, opened 2017-11-24
Allow fuel crossfeed for radial decouplers

aw1621107:
Enabling crossfeed for radial decouplers require the "fuelSystems" tech
node to be unlocked, but that node doesn't exist any more.

Guess the implicit question here is whether fuel crossfeed should be enabled by default, or whether it should be hidden behind a tech node, like it is for stock


NathanKell:
This should not be enabled until quite late in the tree, probably about Shuttle era?

aw1621107:
Does the `orbitalRocketry1981` tech node sound good to you?

NathanKell:
Actually I'd say the materials science bluesky node of that tier might make more sense.

PR #799, opened 2017-12-04
Fix BDB Agena engine tree placement

aw1621107:
From #RO:

    21:36 <@Pap> awang: Agena A should be in orbitalRocketry1959, Agena B in
                 orbitalRocketry1961, Agena D in orbitalRocketry1962


aw1621107:
Github pls. Where did those conflicts come from. I'll close and try again.

PR #801, opened 2017-12-04
Fix RealEngines AJ10-137 tree placement

aw1621107:
AJ10-137 engine config is supposed to be in the 1967 orbital rocketry
node, so it only makes sense for the AJ10-137 part to be in the same
node


NathanKell:
Is that part only a -137 or is it one of the AJ10s we hijack with many configs?

PhineasFreak:
@NathanKell Yes, this engine refers only to the 137 model.

PR #802, opened 2017-12-05
Show three decimal places for tooling dimensions

aw1621107:
Proc parts sliders allow editing with precision of up to three decimal
places, so three decimal places should be shown when displaying sizes of
tooled/untooled parts


ec429:
Looks good to me; I'll pull it into my "testing a bunch of awang's PRs" branch and see how it goes.

PR #806, opened 2017-12-13
Add PARTUPGRADE for AJ10-118F to tech tree

ec429:


EagleEyeLogic:
Let me say one simple thing.
That's a good pull, but more for RO's AJ10_Adv_Config.cfg

PARTUPGRADES are in most cases the job for RO's Engineconfigs. For PARTUPGRADES we should only change icon, description, placement. In general, our configs should only contain content we change or we will change.

If we hook on RO and do this simple in every possible way, much more people can read the configs from RP-0, understand changes and find faster solutions for problems.

But that's only how I understand the way.

PhineasFreak:
PARTUPGRADES are useless under RO, since it is the "sandbox" mode of RSS. Part costing and tech placement only matter under the "carreer" mode of RSS (RP-0/1).

And no, the global engine configs have nothing to to with tech tree PARTUPGRADES. The first one defines how the part actually works and the second one just creates a visual cue where you can get part upgrades for the said part. It does not do anything else.

EagleEyeLogic:
@PhineasFreak 

But most PARTUPGRADES are defined in these EngineConfigs.
If we don't need it in RO, why RO take this part?

I mean the configs there are a copy for RP-0/1 and useless in CTT, but we can simply change the TechRequired for RO and RP-0/1 change only the TechRequired for our TechTree.

https://github.com/KSP-RO/RealismOverhaul/blob/master/GameData/RealismOverhaul/Engine_Configs/AJ10_Early_Config.cfg

and you can check every file with EngineUpgrades, wost are defined

PhineasFreak:
In what file(s) are you saying that we are duplicating PARTUPGRADES nodes? Because i am not seeing them...all that i see are the RF part **variants** and their tech tree part *placement* patches.

You seem to misunderstand what the RF upgrades vs the PARTUPGRADES do. RF manages **variants**, not **upgrades**. **These two things are NOT the same.**

In previous releases of RO we did exactly what you are saying: [defining the tech tree and costs inside the global engine configs](https://github.com/KSP-RO/RealismOverhaul/blob/c6df4f0f7989d661f6af06fe8857386f63b6e2f4/GameData/RealismOverhaul/Engine_Configs/AJ10_Early_Config.cfg#L78-L79). But this also means that we have to time the RO and RP-0 releases to happen at the exact same time, else there might be conflicts between them. So, we decided to separate them so that RO will manage the part physical and mechanical properties (what RO always did) and RP-0 will manage the part costing and tech placement (what RP-0 always did).

Defining RP-0 specific fields inside the RO configs was always wrong and i am glad that @pap1723 did all this wonderful work to finally separate them.

EagleEyeLogic:
I know both parts and what they do. If you look right at the link, it goes to RO-master and not RO-dev,

Was a bit strange because RO-dev for didn't show up as an updated branche.
So this now make sence.

Personally, I would have left variations and upgrades in one file, because changes have to be done in 2 files anyway. But that doesn't really matter, when you know what to do and where to find it.

PhineasFreak:
That makes sense now! Yes, every RO change now is done against the dev version, with master being the latest "stable" one for reference.

I also should be a bit more careful when reading. Sorry for failing to checkout what branch the question was based against.

NathanKell:
I think this implies the spreadsheet needs changing too. @pap1723  ?

PR #813, opened 2018-01-25
Add science core to everything without avionics or crew control

aw1621107:
This eliminates the need to explicitly enumerate all probe cores that
need to have science cores added.

### DIFF
--- a/GameData/RP-0/Avionics.cfg
+++ b/GameData/RP-0/Avionics.cfg
@@ -9,11 +9,8 @@
 
 
 
-// *** Non-control probe cores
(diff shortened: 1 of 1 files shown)


NathanKell:
That should probably be in a later pass to ensure it runs after all avionics are added?

aw1621107:
`:AFTER[RP-0]`, since it looks like all the avionics are in `:FOR[RP-0]`? Or will this require a `z` prefix, e.g. `zRP-0Avionics`?

NathanKell:
AFTER should be fine I think!

aw1621107:
Done!

PR #832, opened 2018-02-15
Fix placement of Atlas LV-3A

sgrif:
The part for AtlasLFTLong is described as the SLV-3 which I believe is
why this was placed in the wrong node. While SLV-3 didn't fly until '64,
the tank is identical to that used on LV-3A which flew the agena in '60.
Without this, there is no Atlas tank that can be used with the Agena on
the node where the Agena is unlocked. It certainly flew before LV3-C
which is on this node as well.


pap1723:
Thanks, this will be fixed on the spreadsheet and pushed with the next update.

PR #836, opened 2018-02-18
Slinqify ALL the LINQs

aw1621107:


jwvanderbeck:
Are you aware that Smooth is very old and defunct? Unless something resurfaced lately, I would be hesitant to rely on it in newer Unity versions.

aw1621107:
Yeah, I noticed that Smooth seems to have vanished from the internet. I was somewhat unsure about using it, since it seems few people on #RO have experience with it, and there wasn't much mention on the Unity forums, but being able to use the LINQ style without the GC penalty just seemed too good to pass up. Figured that (a.k.a. hoping that) the C# side of things is more likely to be backwards-compatible, too, since the source code (minus libraries) isn't specific to Unity. Not sure if it has any bugs though.

In any case, I'll defer to you guys on whether using Slinq is a good idea. If not, I'll just turn these into regular loops and submit another PR.

rsparkyc:
So, looks like this guy posted about it here on reddit about 3 years ago: https://www.reddit.com/r/gamedev/comments/24sx1x/smoothfoundations_mit_licensed_unity_package_incl/
The poster is smooth_p (https://www.reddit.com/user/smooth_p), and it looks like he's still active.  Perhaps you can reach out and he could post his slinq code to github?  That way we can at least see what he's doing.

aw1621107:
@rsparkyc  As far as I can tell the source code is already on github at [pdo400/smooth.foundations](https://github.com/pdo400/smooth.foundations). [PR #1](https://github.com/pdo400/smooth.foundations/pull/1) seems to provide the strongest hint that this is the right repo.

rsparkyc:
Ahh, perhaps.  I had assumed that the author's name would be Oren Novotny, since that's what's in this post: https://archive.codeplex.com/?p=slinq

aw1621107:
@rsparkyc Huh, haven't seen that page before. Looks like it's something pretty different, at least; the zip I got from the download button doesn't seem to contain any code whatsoever, oddly enough.

rsparkyc:
Yeah I noticed (and was disappointed) by that.  I'd have to look at the source you pointed at (which i'm pretty sure is the right source) and see how it works, but my guess is it's pretty non-unity specific (at least in the ways you're using it).  Finding time for such things lately is another story entirely...

I miss RO...

aw1621107:
Quick grep shows that the only mentions of Unity in non-test code are for logging, so hopefully you're right.

Come back! We miss you <3

pap1723:
Ok, what does this do and is it needed to merge?

pap1723:
Thanks, I will close this for now.

PR #839, opened 2018-02-20
Support new BDB LR89/101/105

blowfishpro:
Deprecated old parts still supported for now since BDB still includes
them

### DIFF
--- a/GameData/RP-0/Tree/TREE-Parts.cfg
+++ b/GameData/RP-0/Tree/TREE-Parts.cfg
@@ -3456,7 +3456,7 @@
     RP0conf = false
     @description ^=:$: <b><color=green>From KW Rocketry mod</color></b>
 }
-@PART[bluedog_atlasVernier]:FOR[xxxRP0]
+@PART[bluedog_atlasVernier|bluedog_Atlas_LR101_Radial]:FOR[xxxRP0]
 {
     %TechRequired = orbitalRocketry1956
     %cost = 15
@@ -3504,7 +3504,7 @@
     { name = ModuleTagEngineLiquidTurbo }
 
 }
-@PART[bluedog_atlasSustainer]:FOR[xxxRP0]
+@PART[bluedog_atlasSustainer|bluedog_Atlas_LR105]:FOR[xxxRP0]
 {
     %TechRequired = orbitalRocketry1956
     %cost = 275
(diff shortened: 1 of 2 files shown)


MikeOnTea:
I've already done these changes to the sheet so they will be there once that is rebuilt, as far as i can see just update the sheet since these files are auto-generated.
I did however remove the deprecated parts since they are...deprecated. :)

blowfishpro:
Okay, clearly I'm way behind on how we're doing things

* Is everything done via the spreadsheet now?
* If so, how does version control work since it's a binary file:
  * If a change breaks something, how do we figure out what commit broke it and what the change was
  * How do we resolve merge conflicts? (text files can be merged provided the changes don't conflict, not so with binary files)
* Where is the script that generates the configs?  What's required to run it?
* Precisely which files are auto-generated and therefore should not be manually edited?

MikeOnTea:
I'm no expert here either, so i hope i'm not telling lies, but here goes :)
The sheet wasn't committed until recently, it resides on google docs which also apparently provides version control. Recently there was a csv version of it committed, we also explore to replace the sheet by set of python scripts.
The sheet itself generates the configs via formulas.
TREE-Parts.cfg, TREE-Engines.cfg, ECM-Engines.cfg and identicalParts.cfg are autogenerated.

ec429:
@MikeOnTea I believe also ECM-Parts.cfg is autogenerated.

blowfishpro:
@MikeOnTea so should I close this PR  since you're going to submit the changes?

MikeOnTea:
Well, i guess you can close it, yes.
Before the next release, the configurations will have to be rebuilt from the sheet, so the changes will be in, then. If somebody wants to get them in earlier, i guess there could be done a PR with the files generated from the sheets.

blowfishpro:
Pap indicated to me that the spreadsheet is really just to build the individual configs rather than the entire files, so they still have to be added into the repository via PR.  Did you get different info?

MikeOnTea:
Well, what's the difference? It does generate individual configs of *all* parts in a single sheet as far as i can see, so you copy them into a file, filter out "zzz" lines and you have the new config file. That then has to be committed, or actually, doesn't even have to be committed strictly speaking as it is auto-generated.

pap1723:
This is correct. I am working on detailed instructions as to what the sheet
does and how it operates in order to explain it and clear up some of the
confusion.

pap1723:
Closed, updated with the most recent Spreadsheet pushes.

PR #843, opened 2018-02-22
tech tree placement of various parts

DocRockwell:
SSTU MUS and LV.
Coatl Aerospace Venera probe parts (somethign is overriding the skinmaxtemp  with a low value of the aeroshell parachutes making them useless)
Couple price fixes (some RN vernier engines costing 12500 instead of 20)

### DIFF
--- a/GameData/RP-0/Tree/TREE-Parts.cfg
+++ b/GameData/RP-0/Tree/TREE-Parts.cfg
@@ -10246,7 +10246,7 @@
 @PART[rn_proton_blockd_v]:FOR[xxxRP0]
 {
     %TechRequired = orbitalRocketry1966
-    %cost = 12500
+    %cost = 20
     %entryCost = 70
     RP0conf = false
     @description ^=:$: <b><color=green>From RN Soviet Rockets mod</color></b>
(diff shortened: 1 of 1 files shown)


GameData/RP-0/Tree/TREE-Parts.cfg
```diff
@@ -25359,3 +25359,197 @@
...
+}
+
+@PART[ca_fom_lander]:FOR[RP-0]
+{
+	MODULE
+	{
+		name = ModuleAvionics
+		massLimit = 2.0
+	}
+	@MODULE[ModuleRTAntenna]
```
PhineasFreak: That's...bad. You should never really need to change any RO parameters via RP-0 or vice versa. Is there a reason for that?

DocRockwell: Woops, that wasn't meant to be left in there. I put it together in a separate config and had some other changes to the same parts. Missed one when I was separating the techtree stuff.

PhineasFreak: Oh OK, that makes sense! Ping me if you feel that some RT values need update, since i am trying to push an update for CA that includes the new parts.

pap1723:
These have been added to the spreadsheet.

PR #846, opened 2018-02-26
Update Realengines parts in the tree

leudaimon:
I'm following the part names in RO PRs by @PhineasFreak and using the corresponding parts details from other mods for new parts.


pap1723:
This was done by @leudaimon in the Spreadhsset.

leudaimon:
@pap1723 I think the parts added by this PR are not in the current Dev branch. What would be the best way to include them? As you are not updating or using the csv in the rep, I guess it's directly in the spreadsheet?

PR #864, opened 2018-04-07
Update README.md

Bornholio:
RP-1 Readme


pap1723:
Good first step, we will need to take another look at this with updates right before release.

PR #868, opened 2018-04-08
Delete CTTree.svg

Bornholio:
no longer need CTT in RP-1


Bornholio:
everything works fine in 1.3.1, in 1.4.5 running DX-11 some tech icons go black regardless of CTT or not

PR #873, opened 2018-04-11
fix wrong MM patch in TREE-Engines.cfg

Schnobs:
fixes https://github.com/KSP-RO/RealismOverhaul/issues/1860
(all subtypes of an engine available regardless of research)


pap1723:
Already fixed. This file (per the comments in it) cannot be edited directly as it will be overwritten.

PR #880, opened 2018-04-29
ProcTankSpaceplane: added tooling

Schnobs:
Problem: 
Shielded Proctank was not part of the general RP-1 overhaul and still uses the old system of unlocking sizes by techlevel (no tooling). Also has 100% utilization.

Solution:
Retrofitted tooling and applied a max utilization.

### DIFF
--- a/GameData/RP-0/Parts/RFProcTanks.cfg
+++ b/GameData/RP-0/Parts/RFProcTanks.cfg
@@ -423,8 +423,40 @@
 	}
 }
 
+@PART[RO-ProcTankSpaceplane]:FOR[RP-0]
+{
+	@MODULE[ProceduralPart]
+	{
+                !TECHLIMIT,* {}
+                %TECHLIMIT
+                {
+                        name = start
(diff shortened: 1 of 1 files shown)


GameData/RP-0/Parts/RFProcTanks.cfg
```diff
@@ -423,8 +423,40 @@
...
+	}
+
+}
+
 @PART[proceduralTankRealFuels]:FOR[zzzRP-0]
 {
 	%category = -1
 	%techRequired = ORPHANS
 }
+
```
Tidal-Stream: Why is this line added? No need for it.

Schnobs: I honestly cannot remember anymore. I guess it's either carried over from whatever current config I copied, or a remnant of the old config that's escaped deletion.

Tidal-Stream: It's OK, it was merged already.

PR #882, opened 2018-04-29
More CrewScience Labs

Schnobs:
Equips more parts with the WBIConvertibleMPL part module. Requires https://github.com/KSP-RO/RP-0/pull/881 or it won't work.

I've used the incredibly scientific method of assigning them according to crew capacity: 1 - basic, 2 - 2ndGen, 3+ mature

### DIFF
--- a/GameData/RP-0/Science/CrewScience/CrewLabAdder.cfg
+++ b/GameData/RP-0/Science/CrewScience/CrewLabAdder.cfg
@@ -68,7 +68,7 @@ RESOURCE_DEFINITION
 //**********************************************************************************
 // 	SECOND GEN CAPSULES
 //**********************************************************************************
-@PART[FASAGeminiPod2|FASAGeminiPod2White|K2Pod|Voskhod_Crew_A|RO-Mk1CrewModule|RO-Mk1CockpitInline]:FOR[RP-0]
+@PART[FASAGeminiPod2|FASAGeminiPod2White|K2Pod|Voskhod_Crew_A|RO-Mk1CrewModule|RO-Mk1CockpitInline|RO-Mk1Cockpit]:FOR[RP-0]
 {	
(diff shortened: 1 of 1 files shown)


GameData/RP-0/Science/CrewScience/CrewLabAdder.cfg
```diff
@@ -68,7 +68,7 @@ RESOURCE_DEFINITION
 //**********************************************************************************
 // 	SECOND GEN CAPSULES
 //**********************************************************************************
-@PART[FASAGeminiPod2|FASAGeminiPod2White|K2Pod|Voskhod_Crew_A]:FOR[RP-0]
+@PART[FASAGeminiPod2|FASAGeminiPod2White|K2Pod|Voskhod_Crew_A|RO-Mk1CrewModule|RO-Mk1CockpitInline]:FOR[RP-0]
```
ec429: Any reason for no `RO-Mk1Cockpit` here?

Schnobs: Probably just an oversight. Or maybe, isn't that the X-15 cockpit? Possibly I just honored the new rule that the X-15 may not under any circumstances be suitable for spaceflight.

ec429: No, the X-15 is `Mark1Cockpit`, which is a one-seater.  `RO-Mk1Cockpit` and `RO-Mk1CockpitInline` are two-seater cockpits in tech Prototype Spaceplaces[sic].  Thus if the `Inline` version is here (as a two-seat space-capable cockpit) so should the non-`Inline` one.

pap1723: Added it

PR #888, opened 2018-07-19
NEW commercial satellite contracts

Relic13:
New Commercial Satellite (cmlSat) contracts:

Overview of these new contracts:

 - Player can set (increase/decrease) contract payload quantities for all future generated commercial satellite contracts.  (Payload quantity does affect rewards) Does not change the existing (already accepted) commercial satellite contracts.  Uses exponential stepping (500, 1000, 2000.... units etc) which gives a pretty good payload range while not being overly tedious on clicking and waiting for new ones to repopulate etc.  Should help facilitate designing a class of satellites and launchers for various payload/orbits combinations.

 - Each cmlSat contract offering has its own relevant unlocks; Example molniya requires first molniya sat to be complete, MEO requiring a LEO completion, other planets require their respective first orbit. There are also requirements shared across all: that comTestSat has been completed 3 times.

 - Rewards are tied to orbits and the payload quantity. They start at a lower base value and are scaled a bit stronger on payload quantity than the existing 'advComSat' contract; because larger payloads necessitate larger rockets which in turn have longer build times as well. (IMO Launching a payload that is 800% as large should have more incentive than the -current advComSat- 50% reward increase).

 - Earth orbits of: (Random)LEO, (Random)MEO, (Random)HEO, Molniya, Tundra, Synchronous, Stationary.

 - Inner Planets (Currently as one contract with rand choice between Mercury, Venus, Mars, Ceres)

 - Each outer Planet System, inclusive of those moons with weighted choice chance to favor choosing the main planet over moon.  (ex Jupiter and its respective moons)

These should especially help improve game play:
- for players who are using a launch site limiting orbit insertion inclinations (either from high latitudes sites, or players who self impose house rule of restricting their launch azimuths to safe corridors).   Allows them choose to prioritize orbits such as molniya/tundra over others.  
- Helps out with versatility of rocket size (heavies lifters vs small ones) by allowing player to effectively utilize them all.  Ie a player who only has access to smaller rockets can set the payload down, while a player who just built a new heavy lifter can up the payload to effectively utilize it.
- And gives something for late game missions (outer planet orbits, and/or increase the payload to very large sats requiring heavy lift launchers).

Cheers,

Relic.


Relic13:
Went through and updated all of the sortkeys (forgot a zero) which I guess failed checks since it updated build before the check finished.  :(

Bornholio:
Looks great will start reviewing

Relic13:
Deleting the files for now since it is far easier to tweak and test then re upload them all in a bit.

ec429:
Your design approach looks interesting and I look forward to trying this out, but _please_ use a proper Git client and make meaningful commits instead of (I assume) file uploads through the web interface.  The former not only makes for a more useful commit history, it also avoids sending dozens of emails to everyone in the project.

I do wonder if the approach of using contracts to change the parameters is really the friendliest UI; would it not be possible to add something to the RP-1 DLL to control this instead?

Relic13:
Sorry for the email spam!  I did not realize it would send them out to everybody if I was just doing changes on my fork.  While I've been editing other mods in (my) ksp install for a long time I am admittedly very new to Git.  So again, apologies for the unintended spam :(.

About the interface:
It is pretty simple for the user I think; they have 2 contracts placed above the rest of the individual commercial sat orbit contracts.  Basically it is an increase payload contract and a decrease payload contract (showing new desired payload quantity), with completion only requiring that they go back to space center (or do anything) while 60min time elapses, after this new commercial sat contracts will be issued using the newly set payload target (again, previously issued, accepted or not, are unaffected).  The player can also wait a little bit in mission control if they want all the existing (not accepted) commercial sat contracts to be fully cleared out as well. 


Also, finished testing/tweaking these contracts on my end (since some inadvertent problems crept in), and added the files.

So this should *actually* be finished and ready now. 
Hope to see what you all think about these contracts though.

Cheers,
Relic

hargn:
Maybe the choice of payload quantity could be set by a commercial strategy in the Administration building?
That'll finally make a reason to use this building in RP-0.

PR #889, opened 2018-07-28
Fix GEO Weather Sat Max Completions

Tidal-Stream:


Tidal-Stream:
It seems like the intention was to have 6 GEO weather sats to capture the whole planet and repeatable sun-synchronous sats for more precise and localised measurements. This is what the description says at least.

ec429:
I think we should give primacy to the description and therefore accept this change.
However (unrelated bug), in my experience this contract is uncompleteable for the same reason as the Earth Observation Satellite one, i.e. that horizontal-distance parameters in CC don't work.  I wonder if maybe CC can specify a longitude target rather than using a generated point on the surface?

Tidal-Stream:
Are you sure you got the orbit parameters right? It works fine for me no matter the longitude.
screenshot661

PR #901, opened 2018-09-20
Fix NewCrewStations.cfg (port be7a24b) (#783)

Tidal-Stream:
#783, merged in commit be7a24b, fixes the empty seat requirement in the contract. It was merged only in master and not in Developmental. This ports it to Developmental.


ec429:
@Bornholio there's a separate contract for rotating out an existing crew (`Crew Rotation Station.cfg`), AFAICT that one works fine.

PR #903, opened 2018-10-15
Why do I need <3 empty seats?

glilienthal:
Warning: when I had this locally, this contract somehow seems to have blocked the New Crew Station contract to appear. Do you have an idea why?


pap1723:
This was merged already with the previous push to master

glilienthal:
I think you are confusing the "new crew station" with the "update power module" contract. The latter one is still faulty in master.

PR #904, opened 2018-10-15
reduce Crazy rewards for flight duration

glilienthal:
The current rewards for flight duration are dominating all my contract rewards. 180days gives me 7,200,000 funds.  A year would double that.  (In my settings a moon landing is significantly less than that.) 

Considering that long-time spaceflights are not a  difficult thing in gameplay, I think this should be changed to an approach that is less than linear.


pap1723:
This makes a lot of sense.

PR #905, opened 2018-10-20
Added settings for disabling crew training and retirement features

siimav:
Added options into the game's difficulty settings menu for toggling the RP-1 crew training and automatic retirement features. I know that a lot of people (including me) would prefer to play without having to micromanage their crews. This commit adds the option for disabling either of these features. For new games, these options are now by-default disabled on 'easy' difficult but 'normal' and above  still has them turned on.

The training and retirement updates are still done in the background. Also beware that because retirement dates are unaffected, re-enabling the retirement feature after playing for 10 years can cause most of the old nauts to retire almost instantly. But I thought that it's a better option than increasing the retirement deadline that gets stored for each individual crewmember.

Here's how it looks ingame:
screenshot

### DIFF
--- a/Source/Crew/CrewHandler.cs
+++ b/Source/Crew/CrewHandler.cs
@@ -97,6 +97,8 @@ namespace RP0.Crew
 
         public Dictionary<string, double> kerbalRetireTimes = new Dictionary<string, double>();
 
+        public bool retirementEnabled = true;
+
         protected HashSet<string> retirees = new HashSet<string>();
 
         protected static HashSet<string> toRemove = new HashSet<string>();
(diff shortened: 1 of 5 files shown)


ec429:
Looks good to me, but I'll wait a few days to see if anyone else has comments before merging it.

Bornholio:
Can you make a test .dll
Also it should have a note to reduce rewards by some amount 

Looks great.

siimav:
The modified dlls are available under my own fork [here.](https://github.com/siimav/RP-0/releases/tag/dev-settings1)

Bornholio:
Seems to work fine, should add a bit of warning/instruction on the topic of enabling it after disabling.

siimav:
> Seems to work fine, should add a bit of warning/instruction on the topic of enabling it after disabling.

I think I added a warning about it into the hover tooltip. Not sure what else we could do. Or do you have something in mind already?

Bornholio:
Ah had not noticed the tool tip, my thought is that it could be body text beside the settings.
Tooltip is good though.

PR #910, opened 2018-11-16
Master to Developmental merges

pap1723:


pap1723:
Bring Developmental in line with Master

PR #913, opened 2018-11-26
Have Tank-III be cost-reduced by Tank-II, not Tank-I and SM-I

pianojosh:
@pap1723 

I'm not sure why SM-I would have been in the cost reduction list for a non-SM tank. Maybe something related to the fact that the original sounding rocket payloads are in Tank-I-HP, but if anything, I would see that as a reason to have SM-1 be cost-reduced by Tank-I. I'm guessing this was just an oversight.

### DIFF
--- a/GameData/RP-0/Contracts/Early Satellites/Tundra Satellite.cfg
+++ b/GameData/RP-0/Contracts/Early Satellites/Tundra Satellite.cfg
@@ -101,7 +101,7 @@ CONTRACT_TYPE
 		{
 			name = ReachOrbit
 			type = Orbit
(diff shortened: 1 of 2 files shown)


pap1723:
I changed it to also include SM-I as a modifier as well

pianojosh:
@pap1723 I'm curious why? None of the other tanks have service modules as cost reducers, and I think it makes sense that way. For cryo fuels, they're dewar tanks which is a fundamentally different technology, and as payload containers, it's sort of a different concept entirely.

pap1723:
@pianojosh My thinking was that it is just an improvement on tank design, materials design and learning from your past experiences. My thought is that engineers need to figure out how to make Service Modules as light as possible, so that technology transfers into the next tier of tanks. But I could be way off base and you might be right.

pianojosh:
Hmm. I could see that. Theres definitely some commonality. If we're going to go that route, we should probably do it for the other tank types too though.

PR #914, opened 2018-11-28
Fix loading screen images switching too quickly

siimav:
* Increase the time that each loading screen image is shown to 12 seconds (was 4 previously)
* Removed code that didn't appear to be doing anything useful. I have no idea why the code was previously calling Unity's random generator up to 99 times.

Test dll's are available here: https://github.com/siimav/RP-0/releases/tag/dev2


pap1723:
I changed it to 8 seconds. 12 is a little too long for me and I want to be able to see more (as we continue to add them as well)

PR #919, opened 2018-12-01
Add Speculative NK-9 Configurations

Tidal-Stream:
Add NK-9 upgrades in line with NK-15 and NK-33 with respect to TWR and
ISP. It should give players more options for light Soviet launchers.

### DIFF
--- a/GameData/RP-0/Tree/TREE-Engines.cfg
+++ b/GameData/RP-0/Tree/TREE-Engines.cfg
@@ -880,6 +880,24 @@
                 %cost = 100
             }
 
+            @CONFIG[NK-9-1969]
+            {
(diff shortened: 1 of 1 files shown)


Bornholio:
Blasphemy "Light Soviet" !  looks good to me.

Tidal-Stream:
In addition, given that AJ26-62 costs 300, maybe we could cost NK-9, NK-15, and NK-33. Something like NK-15 being 200, NK-33 250, NK-9 50. Or using RD-107 costing 470 as a benchmark, NK-15 could be 500, NK-33 520, NK-9 100.

pap1723:
These have been added to the new Parts Browser and will appear in the next update.

PR #920, opened 2018-12-02
Add Cassini-Huygens cost and entry cost

Standecco:
Adjusted costs for the cassini spacecraft and the huygens lander from Raidernick's US probes mod.
Sources: https://www.jpl.nasa.gov/news/press_kits/cassini.pdf
https://www.hq.nasa.gov/office/codeb/budget/LIFCY982.htm#CASSINI

### DIFF
--- a/GameData/RP-0/Tree/TREE-Parts.cfg
+++ b/GameData/RP-0/Tree/TREE-Parts.cfg
@@ -20987,9 +20987,9 @@
 @PART[rn_cassini]:FOR[xxxRP0]
 {
     %TechRequired = longTermAvionics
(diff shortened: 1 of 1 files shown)


Standecco:
The Huygens lander was funded entirely by ESA with $500 millions, or at least that's my understanding. Its development is therefore not included in the program costs by NASA alone, if that's the problem.

pap1723:
These will need to be added to the parts JSON files in order to take effect. This will require some simple work so I will close this, but I created a new branch with the changes:
https://github.com/KSP-RO/RP-0/tree/standeco-cassini-huygens-costs

PR #922, opened 2018-12-02
Fix procedural avionics getting stuck in an infinite update loop

siimav:
Fix proc avionics getting stuck in an update loop because of a weird issue with float field control

### DIFF
--- a/Source/Avionics/ModuleProceduralAvionics.cs
+++ b/Source/Avionics/ModuleProceduralAvionics.cs
@@ -422,12 +422,13 @@ namespace RP0.ProceduralAvionics
 
 				float largeIncFloat = (float)largeIncrement;
 
-				proceduralMassLimitEdit.incrementSlide = largeIncFloat / 4000;
+                // There's some weirdness going on here that makes the slider not match up with min and max values.
+                // Because of that, need to ensure that this difference does not get larger than FLOAT_ERROR_ALLOWANCE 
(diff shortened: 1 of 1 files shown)


pap1723:
@siimav Did this fix the error you were having?

siimav:
Yes, previously it was causing a severe loss of framerate and the proc avionics UI became unresponsive. Only way to recover from that situation was to repeatedly tap the slider controls and hope that the clicks are finally registered. I have already seen the same issue reported in RO discord a couple of times.

Here's also a more detailed explanation of this issue that I wrote on discord:

> Basically the slider control works in predefined increments. For example if you set the slider interval at 2 and the actual value is 100, it may automatically snap the value to either 99 or 101.  To make things complicated, it doesn't appear to be doing that when you first manually set the values on the slider. It only starts happening if you close the right click menu and reopen it again. And after that if you try changing it in code, it always reverts back from the actual value to what the slider increments support.
> Now the next issue is that the proc avionics code checks whether the current tonnage value is within the minimum and maximum bounds and tries to set it back to within the allowed range. It does use a 0.002f times error margin for this check but if the slider increments are too small, it could still get outside the allowed range. In my case, 0.002 * 100 is  0.2 tons and the slider always reverted the value to 98.7 or something like that. So because the range check is unable to bring the minimum within bounds, it keeps running in an endless loop.
> Actually that issue should currently only happen on higher tier booster avionics because the maximum allowed tonnage (or the volume of the part itself) must be quite large to run into this issue.

PR #925, opened 2018-12-04
Make the crew training feature less complex

siimav:
Based on the feedback of RO Discord, make proficiency trainings no longer expire and remove mission trainings. This commit also removes all code that's related to training expirations.

Test DLLs are available here: https://github.com/siimav/RP-0/releases/tag/dev3
But be sure to back up your save before loading a game with these DLLs.
It looks like my changes will cause the training expiration blocks to be removed from saves. So even after going back to old DLLs, all trainings would still lack expiration dates.

### DIFF
--- a/GameData/RP-0/CrewHandlerSettings.cfg
+++ b/GameData/RP-0/CrewHandlerSettings.cfg
@@ -17,10 +17,8 @@ CREWHANDLERSETTINGS
 	retireCourageMax = 3
 	retireStupidMin = 1
 	retireStupidMax = 0
-	trainingProficiencyExpirationYears = 4
 	trainingProficiencyStupidMin = 1.5 // multiplier to how long proficiency lasts at 0 stupidity
 	trainingProficiencyStupidMax = 0.5 // multiplier to how long proficiency lasts at 1 stupidity
-	trainingProficiencyRefresherTimeMult = 0.25
 	trainingProficiencyXP = 1
 	trainingMissionExpirationDays = 120
 	trainingMissionStupidMin = 0.5 // multiplier to how long the training course takes at 0 stupidity
--- a/Source/Crew/ActiveCourse.cs
+++ b/Source/Crew/ActiveCourse.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using UniLinq;
-using System.Text;
 
 namespace RP0.Crew
 {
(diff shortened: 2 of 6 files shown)


NathanKell:
hashtag DELURK Quick note: I think proficiency training not expiring makes a lot of sense, but I'm leery of removing mission training because the point of having it at all is so that you can't do instant launches. In particular, without mission training, it's possible to keep a built LV ready to go on the pad and launch a rescue (or other crewed mission) just-in-time, rather than having to plan ahead.

What's the major gripes people have with mission training?

RoboRay:
The only issue I have with Mission Training is that I can't _schedule_ it to begin on a specified future date...  I can only start it _now_.

If it's July and a vehicle will be ready to launch in mid October, I have to be careful to not blow past the date I need to start training on, or manually set up a KAC reminder.  In reality, I would just tell the crew and instructors in July to start training in early October and not need to worry about anymore.

If the training system could generate KAC or stock event system reminders, that would help eliminate the need to babysit the calendar while timewarping or doing other missions.

I do find that Mission Training adds another layer of planning to the game and think it's worth saving.  I just find it tedious to use as implemented _right now_.

lpgagnon:
For me, the worst feature of mission training is that it gets lost on a launch failure. Now that launchpad ignition failures can be disabled, that's greatly mitigated; but it still makes no sense to me that the astronauts need to train again for the "same" mission

NathanKell:
Should be able to not clear training on recovery if MET was below threshold?

Sent by my thumbs, slowly.

pap1723:
@NathanKell here is a different approach to the question...Looking at the code, you pretty much already have the constraints built in with the Training Time limit of (is it 120?) days and also the mandatory inactivity period afterwards. If the flight is very short, the inactivity period will be short and they will be able to fly again since the inactiveTime is:

`double inactiveTime = elapsedTime * multiplier + constant * 86400d;`

I think having training expire just because a mission is accomplished doesn't make sense. Instead, we should leave the expiration date to the current setting and get rid rid of the removal of training proficiency upon recovery. So I think we should remove this section completely:

```
if (e.type == "TRAINING_mission:)
   SetExpiration(pcm.name, e, Planetarium.GetUniversalTime());
```

pap1723:
Proficiency is also something that can be set by the config file if I remember correctly? I think we should move the proficiency expiration dates into the RP-0 Settings Fields so if people want to, they can set it to much different values if they would like.

pap1723:
Closing this for now, please see issue #942 to continue discussion.

PR #932, opened 2018-12-09
Fix CKAN identifiers for RCSBuildAid and HangarExtender

DrRaphello:


pap1723:
@DrRaphello are you sure those are the proper identifiers for the 1.3.1 versions of those mods? I thought I had checked them (but there is a good chance I missed it or pulled old stuff)

DrRaphello:
@pap1723 Yes, I've just double checked on NetKAN - both were renamed when taken over by LGG and updated to 1.3

DrRaphello:
One more question - is dmagic orbital science a hard dependency now?

pap1723:
Yes it is. Many of the new science parts are modified versions of the
DMagic Orbital stuff.

Bornholio:
not sure how but can the links be made virtual so if the original is restored (like FAR vs FAR cont. in RO now) you can use either?

pap1723:
Great if it can, but I don't enough about CKAN.

DrRaphello:
hmm, I can see nothing in RO netkan file that does such a substitution. They seem to have a pretty standard definition: "depends" : [ { "name" : "FerramAerospaceResearch" }. I have not found any obvious solution in netkan specification either.
I'll ask in ckan thread anyway

DrRaphello:
it seems possible for FAR but not for these two mods: [link](https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1254-kennedy/&do=findComment&comment=3502231)

DrRaphello:
So, is seems possible after all based on further response from HebaruSan. However, I see possibility of old mod version revival so low that it does not make sense to ask him and LGG to spend time on this.

pap1723:
Couldn't we leave both the old one and the new one as links? That way if the old one is ever reactivated, LGG drops the new one. Any drawbacks to this?

DrRaphello:
We could do this too. The only risk is an unlikely scenario where both original dev and LGG continue their separate versions.
Anyway, I consider RP-1 a finished work in the sense that vast majority of mods will never be updated for 1.3.1. It is therefore enough to install all currently available versions for 1.3.1. For this purpose both my approach will work and your approach will work. 
I don't pretend to have such modding experience as you do so I leave the choice to you :)

pap1723:
Commited directly to master as it affects the main netkan read by CKAN

PR #953, opened 2018-12-25
Patch 1

Bornholio:


Bornholio:
This should be added via part browser not a patch .shrug  
Will close when i do a pr that way

Bornholio:
If anyone wants these in feel free to make a correct PR for them.  I won't have time in several weeks.

pap1723:
Added through Commit: https://github.com/KSP-RO/RP-0/commit/321aa5172da4a288cf4d6c195098541027670ba7

PR #957, opened 2018-12-27
MaintenanceHandler performance fix

siimav:
Fix MaintenanceHandler causing performance issues for vessels that are outside Earth's SOI.

// TODO: needs some further investigations into why facility levels are not accessible. Maintenance fee calculations are most likely also incorrect.


pap1723:
After some testing by @vallin91 and others, this has proven to fix the issue, great job @siimav !!

NathanKell:
What exactly was the problem here?  Looks like the find-objects got moved?

siimav:
Added #961 to fix the maintenance fee calculations.

PR #958, opened 2018-12-28
Add SSTU booster tooling, reimport nose/mount model lengths

MikeOnTea:


GameData/RP-0/Parts/SSTUToolingTanks.cfg
```diff
@@ -305,7 +305,112 @@
...
++PART[SSTU-SC-TANK-MFT-D]:FOR[RP-0]:NEEDS[SSTU]
+{
+	@name = SSTU-D-Tank-II
+	%RSSROConfig = True
+	
+	@title = Modular Booster Tank, II
+	@manufacturer = SSTU
+	@description = Level II-equivalent booster fuel tank. Used on the R-7 and similar to tanks on the Redstone, R-5, and Able rockets. Can be configured with normal or high pressurization. High pressurization tanks are required by some engines and they are heavier than normal tanks. <b><color=green>Max Utilization: 92%
+	
+	@maxTemp = 773.15
```
PhineasFreak: Why are you configuring the physical properties via RP-1? These all **must** be part of RO.

GameData/RP-0/Parts/SSTUToolingTanks.cfg
```diff
@@ -305,7 +305,112 @@
 	}
 }
 
-@PART[SSTU-A-Tank-*]:AFTER[RP-0]
++PART[SSTU-SC-TANK-MFT-D]:FOR[RP-0]:NEEDS[SSTU]
```
PhineasFreak: The NEEDS check is not required since patches do not apply if the actual parts are not installed.

GameData/RP-0/Parts/SSTUToolingTanks.cfg
MikeOnTea: Good question/catch, i guess we were just copying the parts from above. I assume pap also only copy/pasted it, so i'll remove all the physical properties in the file.

MikeOnTea: Well, that's another good question. For RP-0, NK created 4 "levels" of fuel tanks, level 1-4, to have a tech progression. Without any progression, you technically don't need those 4 tank levels in RO, but i can see your point, they're physical things and you still might want to recreate an early rocket with procedural parts just using RO. We should probably get some other opinions on this.

PhineasFreak: I had a similar situation with the "X-Ray" probe core. It was an RP-0-only part and it was impossible to actually test a craft file that included it under just RO.

MikeOnTea: Right, so we do actually change the physical properties of the descendant tanks, so the values itself belong to the part. About the move, the plan seems to be to go ahead with this PR, and move the parts over to RO later, together with procedural avionics i assume.

PR #964, opened 2018-12-29
Soviet and NTR Engine ECMs

leudaimon:
Adds ECMs to the three main Staged Combustion engine families, RD-58 upper stages, and the N1 and Proton rockets engines and the NTR engines suggested by @Bornholio


Bornholio:
looks good

pap1723:
Implemented with https://github.com/KSP-RO/RP-0/commit/4dd1c3e1458aeab46bbe5528e8f25b74a3f73efc

PR #973, opened 2019-01-01
 Added CC extension for downrange SR contract

siimav:
To use in CC contract configs, add the following block:
```
PARAMETER
{
	name = DownrangeDistance
	type = DownrangeDistance
	title = Downrange Target
	distance = @/targetDistanceKM * 1000
}
```


theonegalen:
What about submitting this as a pull request to contract configurator, so other mods can take advantage of it as well?

PR #975, opened 2019-01-02
Added highlight to 2nd gen navsat contract for unexpected resource type

gazpachian:
The second generation Navsat contracts have ComSatPayload as a requirement, which is not intuitive. Added highlighting in the contract blurb to aid players somewhat.


gazpachian:
Done and done. #FF0000 isn't pretty from a design perspective, but it sure stands out which I guess is more important! I got the style from other contract blurb highlights in RP-1 where white is used, should they be changed to red also?

PR #977, opened 2019-01-03
Added ECM's and Ident parts info for the Bluedog and FASA Agena avion…

mattwrobel:
…ics parts.


Bornholio:
Looks good, have test files?

PR #986, opened 2019-01-11
Missing texture for the HRIR experiment truss

gazpachian:
Someone just forgot to bundle it from the original mod, easy fix.


pap1723:
Thanks!

PR #988, opened 2019-01-13
Fix BDB RL10

Standecco:
Fixed a wrong config that prevented the part to be correctly placed in the tree


Standecco:
wrong bdb version installed, ignore

PR #992, opened 2019-01-18
Fixed typo in explorer core ECMs

mattwrobel:


pap1723:
I had been messing with ECM's all day so I just added your changes to this commit: https://github.com/KSP-RO/RP-0/commit/1dccc3fb7251cdddf1e3aaad9bc5b67000ce9975

PR #995, opened 2019-01-20
 Fixed a few ECM's with periods (should be dashes), fixed RealEngines…

mattwrobel:
… AJ10-137 fields


pap1723:
We'll figure these all out eventually!

PR #999, opened 2019-01-22
Update BodyScienceParams.cfg

Bornholio:
Add missing moons of Uranus and Ceres and Vesta


pap1723:
Incorporated by #1005

Bornholio:
thanks!

PR #1001, opened 2019-01-22
Module tag updates

mattwrobel:


mattwrobel:
> 
> 
> Also looks like a bunch of non-tag stuff happening too? periods/underscores/dashes stuff

That's just a result of generating the files from the latest code which hadn't been done for the period to dash stuff.

PR #1002, opened 2019-01-24
Allow proficiency training while the part is still being researched

siimav:
Creates proficiency courses for all crewed parts that are in the KCT's research list. Mission trainings are not available before the part is actually purchased. If the research is finished but the user does not unlock the part, then the proficiency course will also disappear. I think it might actually be a good thing because this way the list wouldn't get cluttered up with entries for parts that will never be used.


Bornholio:
Working good, early game doing x-1 with no issues

PR #1006, opened 2019-01-28
Module Tag Maintenance:  Removed EngineLiquidPF tag from services mod…

mattwrobel:
…ules and combined parts.


NathanKell:
Looks like the SMs still have avionics? Those tags should be stripped too. Otherwise, LGTM so merging, the avionics can be stripped in another PR.

PR #1021, opened 2019-02-19
Updated Tuning

ppboyle:
Updated tuning to make early distance contracts a little more valuable


pap1723:
This was rolled into a different PR, so good to go!

PR #1022, opened 2019-02-19
Create Downrange Records

leudaimon:
* Uncrewed downrange records based on the Altitude records
* The reward is the same as for altitudes
* Must go suborbital (above Kármán line) - avoid completion by plane
* Must be uncrewed

### DIFF
--- a/GameData/RP-0/Contracts/Records/DRRecord.cfg
+++ b/GameData/RP-0/Contracts/Records/DRRecord.cfg
@@ -8,7 +8,7 @@ CONTRACT_TYPE
 	genericTitle = Uncrewed Downrange Record
   genericDescription = We want you to set a new uncrewed downrange record!
   synopsis = Set an uncrewed downrange record of @uncrewedTargetDRKM km.
-  completedMessage = Congratulations! You've set a new uncrewed downrange record and is closer to your first Orbital Vehicle!
+  completedMessage = Congratulations! You've set a new uncrewed downrange record and are closer to your first Orbital Vehicle!
 	
 	agent = Federation Aeronautique Internationale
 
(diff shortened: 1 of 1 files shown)


Bornholio:
how about +50km instead, since we have dead kerbs in AC cockpits above that.  
I think long distance flights and cross country speed records would be interesting with a <50km limit.

leudaimon:
You can still fly a plane even without kerbals (or even a cockpit)... but the main reason I asked for the >100km apoapsis is so that this is a progression into orbit in fact.
The long distance flight thing I don't see much reason. It's more of a prize for patience, as you can complete it easily with a very simple plane and a lot of fuel.

GameData/RP-0/Contracts/Records/DRRecord.cfg
```diff
@@ -0,0 +1,118 @@
...
+{
+  name = recordUncrewedDR
+  group = Records
+	
+  title = Uncrewed Downrange Record of @uncrewedTargetDRKM km.
+  description = We want you to set a new uncrewed downrange record! Reach @uncrewedTargetDRKM km downrange in a suborbital trajectory (going above the Kármán line).
+	genericTitle = Uncrewed Downrange Record
+  genericDescription = We want you to set a new uncrewed downrange record!
+  synopsis = Set an uncrewed downrange record of @uncrewedTargetDRKM km.
+  completedMessage = Congratulations! You've set a new uncrewed downrange record and is closer to your first Orbital Vehicle!
```
Rockstar04:
```suggestion
  completedMessage = Congratulations! You've set a new uncrewed downrange record and are closer to your first Orbital Vehicle!
```

pap1723:
Cannot get the setup to work, instead NK put in a Distance Milestone for pre-First Orbit that is a good balance.

PR #1029, opened 2019-02-26
Attempt to fix CI by increasing clone depth

blowfishpro:
Travis defaults to last 50 commits, but this does not include the most recent tag so `git describe` failed.  Increasing the clone depth should allow it to find the tag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: c
 
+git:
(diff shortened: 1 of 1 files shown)


blowfishpro:
Looks like this fixed it, it'll break again once we reach 500 commits since the last tag though

PR #1030, opened 2019-02-26
Added json files for support of Deep Freeze and SmartParts

thesciencenut:
Created configs for Deep Freeze mod. 

Added them to Near Future Life Support (2030). 

Change as needed.


pap1723:
There is not enough information including with these files to determine if what you are doing is correct. Please provide more information in the JSON so we can review to add it.

PR #1034, opened 2019-02-27
Update Asteroid Landing rewardScience

Rockstar04:
From the rest of the changes made, it looks like this should be a float.


pap1723:
Thanks! Nice catch!

PR #1038, opened 2019-03-05
Prerelease todos

leudaimon:
* Fix the placement of the Early Avionics procedural Booster avionics unit 
* Adjust the cost of the early probes (20in and Explorer 1) for the removal of their experiments


pap1723:
Rolled your changes into previous 2 commits since I messed up my JSON files!!

PR #1039, opened 2019-03-05
Added json files for support of SmartParts

thesciencenut:


Source/Tech Tree/Parts Browser/data/SmartParts.json
```diff
@@ -1,15 +1,15 @@
 [
     {
         "name": "FuelController",
-        "title": "",
-        "description": "",
+        "title": "KM Fuel Controller",
```
pap1723: This should not be included. Leave it in here, but change "orphan": true

Source/Tech Tree/Parts Browser/data/SmartParts.json
```diff
@@ -24,15 +24,15 @@
     },
     {
         "name": "kb-fuel-breaker1",
-        "title": "",
-        "description": "",
+        "title": "Fuel Flow Breaker 1.25m",
```
pap1723: This should not be included. Leave it in here, but change "orphan": true

Source/Tech Tree/Parts Browser/data/SmartParts.json
```diff
@@ -47,15 +47,15 @@
     },
     {
         "name": "kb-fuel-breaker15",
-        "title": "",
-        "description": "",
+        "title": "Fuel Flow Breaker 1.875m",
```
pap1723: This should not be included. Leave it in here, but change "orphan": true

Source/Tech Tree/Parts Browser/data/SmartParts.json
```diff
@@ -70,15 +70,15 @@
     },
     {
         "name": "kb-fuel-breaker2",
-        "title": "",
-        "description": "",
+        "title": "Fuel Flow Breaker 2.5m",
```
pap1723: This should not be included. Leave it in here, but change "orphan": true

Source/Tech Tree/Parts Browser/data/SmartParts.json
```diff
@@ -93,15 +93,15 @@
     },
     {
         "name": "kb-fuel-breaker3",
-        "title": "",
-        "description": "",
+        "title": "Fuel Flow Breaker 3.75m",
```
pap1723: This should not be included. Leave it in here, but change "orphan": true

Source/Tech Tree/Parts Browser/data/SmartParts.json
thesciencenut: While not exactly the same thing, the Falcon Heavy was originally envisioned to have fuel crossfeed. You could use this to accomplish that, or even to shut down engines when you reach a specific fuel level (so that you could save some for a landing). 

I'll mark it as an orphan part for now, but just something to keep in mind.

PR #1047, opened 2019-03-08
Fixed the names of parts in NFSpacecraft

thesciencenut:
credit to Kerrigan778 on discord for spotting the issue and putting together the bulk of this fix. 

Basically all of the names for each part as well as the folder structure appears to have changed since the last implementation of this config, all this does is fixes the names to the actual ones so that the parts will work properly.


thesciencenut:
Yes, I know that there isn't a description, but that should carry over from the original parts.

thesciencenut:
@pap1723 But doesn't the original description from the part carry over when you add it to the tree?

And how do you typically add things that are a work in progress? Just leave out RP-0?

I suppose the better question is what can I do to add the configs first, but then continue to tweak later?

pap1723:
Yes, the full description is not necessary for anything in game. BUT if things need to be tweaked for balance later or if someone else wants to take a shot at pricing them, including a description is useful. The JSON is used for RP-1 devs and players to easily place parts and price parts into the tree. Without building the JSON will as much info as possible, it is useless as a tool.

Yes, WIP parts should not get an RPO tag. Then thry are on the tree properly, but not costed properly

thesciencenut:
@pap1723 I made the changes and set the RP-0 flag to false. I'm happy to add descriptions as I go, but those other parts should be disabled now.

pap1723:
This will need to be resubmitted. Please do when it is fully formed.

PR #1050, opened 2019-03-10
Update SSTU.json I added pricing for SSTU LC3 and LC5 pods both were RP-0 no cost and moved them to earlier tech nodes

karlack26:
The LC3-POD, I moved from interplanetary Rovers (mid 90s) tech to advanced un-crewed landing.(1975-80) 
I view this pod to be used for expanded lunar missions. With possible touch and go landings to the martian moons and for the daring a martian landing.  With astronaut training players really wont be able to use this pod until the late 70s
I priced the LC3-Pod a little more then the LC-2-POD(larger vehicle and more avionics) Unlock costs are a little less. (similar parts but increased crew capacity and larger vessel size some re-engineering involved)

The LC5-POD I moved from SIAD (2019) to Interplanetary Rovers (mid 90s) a option for players with a commitment to interplanetary exploration.  Capable of longer surface excursions and greater crew capacity. 
Pricing about double of earlier lander pods. It is much larger vehicle with greater crew capacity. Avionics can support large vehicles
Unlock price much higher then early lander pods, more then double.  More of a new design then evolution of previous lander, Plus increased safety and design standards  increases development costs.  

a note perhaps changing the names of those tech nodes may be a good idea. To take into account crewed interplanetary mission. This task is beyond me at the moment.


pap1723:
Looks good!

PR #1055, opened 2019-03-15
Fix the deep space upper avionics

leudaimon:
The way I implemented it before does not work. Now they are a separate procedural avionics config.
I won't be able to test until Sunday though. If anyone wants to give it a try, please do.


siimav:
Appears to be working correctly in my install

PR #1062, opened 2019-03-30
 Don't lock RCS toggle if no avionics

blowfishpro:
It can end up being on anyway because it was on while avionics were attached, and we want to be able to do velocity trim on axial thrusters


siimav:
Sounds good to me

PR #1065, opened 2019-04-02
Some more tweaks to RN Skylab

Brummi1234:
I went over Raider Nicks Skylab mod once more and tweaked the costs of the individual parts to make them a bit more usefull.

Solarpanels got a bit more expensive, but got ECMs so they should be cheaper once you need them.
The fairings and mm-shields got the same entry cost and per part cost as a procedural fairing would have.
The dockingport also got ECMs and are now way cheaper once you have the apollo ones.
The ASAS module got way cheaper, because all it does is make it remote controllable which can also be done with every other controll core, so you don't get punished for using it. (I don't know why this even exists)
The actual skylab got its cost adjustet so the complete station costs the same as before.

To summarize:
Befor my PR from yesterday: Entry cost 720.000  Cost: 50.212
Now WITHOUT ANYTHING else that has some ECMs: Entry cost 1.144.000 Cost 50.212
WITH tech unlocked you would have before you need skylab (one node before were the parts are for the most part): Entry Cost 528.000

Oh and can somebody explain to me how to keep my fork up with the KSP-RO one without making PRs to mine and have all the extra commits in it?
However, by comparing the files changed this should work, there are just a few extra (useless) commits in it.


Toothspit:
> Oh and can somebody explain to me how to keep my fork up with the KSP-RO one without making PRs to mine and have all the extra commits in it?

You have to add this repo as a remote upstream of your fork:
`git remote add upstream https://github.com/KSP-RO/RP-0.git`
Then you can locally `git pull upstream master` to pull over all commits on this repo's master to your fork, basically udpating it locally, and `git push` to update it on GitHub.

PR #1067, opened 2019-04-04
Contract: First Space Station

Brummi1234:
Changed the required crew from 4 to 3 so it can be completed with Salyut 1, the actual first space station.

### DIFF
--- a/GameData/RP-0/Contracts/Space Stations/First Space Station.cfg
+++ b/GameData/RP-0/Contracts/Space Stations/First Space Station.cfg
@@ -58,7 +58,7 @@ CONTRACT_TYPE
 		name = vesselGroupLaunch
 		type = VesselParameterGroup
 		title = Put your first Space Station in Orbit
-		notes = The station must include at least 1 Docking Port and support for at least 4 Crew
+		notes = The station must include at least 1 Docking Port and support for at least 3 Crew
 		define = spaceStation
 		defineList = spaceStationsEarth
 
@@ -82,8 +82,8 @@ CONTRACT_TYPE
 		{
 			name = HasCapacity
 			type = HasCrewCapacity
-			minCapacity = 4
-			title = Space for at least 4 crew
+			minCapacity = 3
+			title = Space for at least 3 crew
 			hideChildren = true
 		}
 		PARAMETER


NathanKell:
Doesn't it only require two crew, but space for four? Salyut, by KSP terms,
had space for 6 (or more) with Soyuz docked.

Sent by my thumbs, slowly.

Brummi1234:
It needs to be uncrewed and have space for 4. So if you bring 2 or 3 up you fail the uncrewed check.
Pap mentioned in the Disord it was at 4 to stop people using Apolle (or Soyuz i guess) as a space station

pap1723:
Any thoughts on this? What @Brummi1234 says is true, but there is also the point that the new version of the contract works by putting an Apollo capsule in orbit and putting crew there.

What other requirements should there be for something to be considered a "Space Station". There has never been one launched without Solar Panels, correct? What else? A specific amount of a resource?

ec429:
Can we not just have a ModuleTagStationHabitation, or can CC not see those?

pap1723:
Yes, we would just need to add the tag and then tag the parts. It makes sense.

Brummi1234:
Hi, yeah that's what i had in mind with issue 1068 (don't know how to link it on mobile) but forgot to close this pr :)

lpgagnon:
If someone _wants_ to keep an apollo crewed and powered long enough to qualify as a space station, what's the rationale for preventing it?

Only obvious one I see is "too cramped". Kerbalism deals with that among its stress factors; what would be the natural non-kerbalism way to handle the concept?

SerinaTsukaya:
Well if you put an empty Apollo capsule in orbit as your station then dock another Apollo craft to it, then i mean, it's kind of a station in a sense? Maybe we should change the contract goal instead to require you to have two craft launched at two different times dock in orbit with twice as much crew capacity as there is actual crew. You won't stop people from docking an apollo with an empty apollo, but at the very least some effort has to go into it. On a different, note, what use is there for non kerbalism space stations? Does the lab function as it does in stock?

Brummi1234:
The reason for not allowing apollo/soyuz as a station is probably that you should have to research and pay for a station part to get the contract money.

Issue #1097, opened 2019-06-07
RP-0 plugin does not declare dependency on RealFuels plugin


yump:
This causes the loading of some classes related to tooling to fail.  On my machine (KSP 1.6.1, Fedora Linux 29) the symptoms include:

- No prompt about crew retirement on starting a new career

- No £ toolbar button

- No tooling button in VAB part action menu

- This message in KSP.log:

>[ERR 00:06:21.162] ADDON BINDER: Cannot resolve assembly: RealFuels, Culture=neutral, PublicKeyToken=null

>[ERR 00:06:21.163] ADDON BINDER: Cannot resolve assembly: RealFuels, Culture=neutral, PublicKeyToken=null

>[ERR 00:06:21.165] AssemblyLoader: Exception loading 'RP-0': System.Reflection.ReflectionTypeLoadException: The classes in the module cannot be loaded.
>  at (wrapper managed-to-native) System.Reflection.Assembly:GetTypes (bool)
>  at System.Reflection.Assembly.GetTypes () [0x00000] in <filename unknown>:0 
>  at AssemblyLoader.LoadAssemblies () [0x00000] in <filename unknown>:0 

>Additional information about this exception:

> System.TypeLoadException: Could not load type 'RP0.ModuleToolingROTanks' from assembly 'RP0, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.

> System.TypeLoadException: Could not load type 'RP0.ModuleToolingPTank' from assembly 'RP0, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.

Seeing as this wasn't uncovered earlier, it might be a load-order-dependent problem that is only triggered on Linux due to file system differences.

siimav:
When did you last update ROTanks? Was it before https://github.com/KSP-RO/ROTanks/commit/34aa573de8e3156047b802424e5c51f8c1b77dac ?

yump:
The problem occurs without ROTanks present, using CKAN RealFuels as described in [the guide](https://github.com/KSP-RO/RP-0/wiki/Installation-for-1.6.1-(WIP\)).

To be sure, I tried it just now with the current ROTanks and bundled RealFuels.  Same symptoms, same error in the log.

PR #1098, opened 2019-06-07
Declare dependency on RealFuels

yump:
This fixes issue #1097.

### DIFF
--- a/Source/Properties/AssemblyInfo.cs
+++ b/Source/Properties/AssemblyInfo.cs
@@ -37,3 +37,4 @@ using System.Runtime.InteropServices;
 
 [assembly: KSPAssembly("RP-0", 1, 00)]
 [assembly: KSPAssemblyDependency("ModularFlightIntegrator", 1, 0)]
+[assembly: KSPAssemblyDependency("RealFuels", 1, 0)]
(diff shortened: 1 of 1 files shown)


siimav:
I think it would be better to require RF version 12.7 or 12.7.4. Anything lower than that wouldn't work anyway.

yump:
TBH I stumbled on this solution while trying to get it to build to start troubleshooting the problem.  Both ModularFlightIntegrator and RealFuels were in the references, but only MFI was listed here.  I copied and pasted on a hunch, and the log errors and breakage went away.

Further investigation reveals that even though MFI is currently on version 1.2.6, it claims to be 1.0.0 through this mechanism.  So that was apparently a deliberate choice of arguments rather than just a maximally accepting default or boilerplate.

siimav:
https://github.com/sarbian/ModularFlightIntegrator/blob/master/Properties/AssemblyInfo.cs#L39
MFI doesn't change either the AssemblyVersion or KSPAssembly numbers on releases to prevent updates from breaking dependant mods.

yump:
Ah, I should mention that this change requires a recompile to be effective, and I didn't include the rebuilt .dll in the PR for the obvious security and, "had to tweak the .csproj to make it build and can't test on Windows," reasons.

Issue #497, opened 2016-03-28
Disable MechJeb when avionics locked


NathanKell:
@sarbian added a property so we can disable MechJeb. Let's use it!

stratochief66:
I will make sure to reverse this for my personal play, as I am a filthy cheater.

Schnobs:
Careful there. RT still lets us rotate the craft even without a link. Good thing, too, as maneuver nodes tend to wander (both before and during the manuever), so one can't line up the vessel ahead of time. The only thing that MJ is capable of on top off that is handling the throttle. So that's the only capability that needs to be taken away.
...and I'm not even sure if that really has to go.

jwvanderbeck:
The problem though is that with MJ installed, Avionics essentially means absolutely nothing.  You have full control over the rocket regardless of whether you have any active avionics or not.

pap1723:
Would this fix MJ working despite not have sufficient avionics?

NathanKell:
That's what the issue is about, yes. :)

lamont-granquist:
The property to use is here:

https://github.com/MuMech/MechJeb2/blob/0f13d838430c5aa1a052708197032032fa6f67fb/MechJeb2/MechJebCore.cs#L62-L76

You can find any MechJebCore on the ship and send `DeactivateControl = true` to it, and that API will look up the master MJ for the vessel and disable it.

Obviously if the vessel splits the new child vessel with a new core won't get the right persisted value.

It is likely best to hammer on this API once every tick, then when vessels dock or undock it gets asserted to be correctly true or false based on avionics, for every vessel.

Note also that vessels in physics range that lack avionics should still be hammering on this setting as well even if they're not the active vessel.

lamont-granquist:
What that API will do is inactivate the modules that hook OnFlyByWire, which are things like the ThrustController and AttitudeController.  Most other modules and all the display modules hook FixedUpdate instead.  So PVG will still run in FixedUpdate and send throttle and attitude commands to the ThrustController and AtittudeController, but those modules will be disabled, so appropriate levels of uncontrollable vessel hilarity will ensue.  There might be bugs lying around with modules that twiddle with the thrust or the attitude directly, but those would just be bugs that would hopefully get exposed after this is added to RP-1.

PR #1105, opened 2019-06-20
Added control locking for MechJeb

RCrockford:
Added control locking for MechJeb


RCrockford:
Fixes #497.

Bornholio:
needs a switch for filthy cheaters, preferably.

PR #1116, opened 2019-08-01
move the AJ10_Mid configuration to 1960

### LINKED ISSUE #1115: AJ10-104 unlocks in 1960, but the AJ10_Mid unlocks in 1961
Looks like this engine configuration was flown in 1960 (http://www.astronautix.com/a/aj10-104.html), and while we have the config unlocking at that time, the engine for in doesn't unlock until 1961.  We should update the part to unlock in the 1960 node

rsparkyc:
This fixes #1115 
Tagging @pap1723 since he first committed this


siimav:
You also need to edit the config here: https://github.com/KSP-RO/RP-0/blob/master/Source/Tech%20Tree/Parts%20Browser/data/ROEngines.json
Otherwise your changes will be overwritten the next time someone regenerates the part configs.

rsparkyc:
I can update that, it looks like it also has an entry cost modifier of the AJ10_104.  Not sure how that part works - does that mean if you have the 104 that this part is cheaper, or does having this make the 104 cheaper?  Either way, I don't think you can get the 104 config without the Mid engine

siimav:
The ECM basically means that you get the 104 config for free if you unlock the AJ10 part itself

rsparkyc:
@siimav ok, that makes sense (and seems desirable in this situation).  I've updated the PR

PR #1117, opened 2019-08-04
Reduce tech node & entry costs for early solid motors

siimav:
Part entry cost for upper stage solids was reduced by 25K. The biggest change is probably in the tech tree where early nodes got a sci point reduction in the range of 2-3x. I would've liked to reduce the first node to 0.5 points but unfortunately KSP doesn't support floating point values there. I chose the new values based on how much useful tech you get by researching it. Because of that the 1956 node ~~got a higher cost than 1958.~~ has the same cost as 1958.


siimav:
Changed solids1958 from 4 to 5 and solids1959 from 6 to 7

PR #1118, opened 2019-08-05
Made waypoint naming for the lunar rover contract consistent

gazpachian:
In certain config points the second waypoint was labeled 'Beta', in others 'Bravo'. Now it's all NATO compliant. _Loud cheering in the background_

### DIFF
--- a/GameData/RP-0/Contracts/Moon Exploration/Lunar Rover.cfg
+++ b/GameData/RP-0/Contracts/Moon Exploration/Lunar Rover.cfg
@@ -94,7 +94,7 @@ CONTRACT_TYPE
 
 		PARAMETER
 		{
-			name = WaypointBeta
+			name = WaypointBravo
 			type = VisitWaypoint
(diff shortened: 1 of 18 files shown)


siimav:
This isn't the only contract that has waypoints

gazpachian:
@siimav I went and replaced the occurrences in the other rover contracts as well, as far as I know the rover contracts are the only ones with the aforementioned naming convention in place(?). Please let me know if I missed any!

siimav:
I wonder if this change is going to break  any active rover contracts in existing careers...

gazpachian:
That's a valid concern, but from my understanding (and someone please correct me if I'm wrong) this change does not affect active contracts, only the generation of new contracts. Accepted contracts are stored in a node in the sfs save file together with all relevant data such as waypoint names as they were when you accepted the contract. Someone more experienced with CC specifics can confirm or deny this.

PR #1123, opened 2019-08-09
Reduced Nike Part Cost

nepphhh:
Despite the (I assume) historical accuracy of this cost, it's well over double that of the Castor-1 and over seven times as expensive as the AJ2.5. For a reasonable Aerobee 200 analogue (we don't have an AJ60-92, but the AJ10-27 produces comparable thrust), the Nike alone is up to 3/4 the total cost.  Halving the cost will keep it expensive compared to similar motors but not stupidly so.

### DIFF
--- a/GameData/RP-0/Tree/TREE-Engines.cfg
+++ b/GameData/RP-0/Tree/TREE-Engines.cfg
@@ -535,7 +535,7 @@
 
             @CONFIG[GCRC]
             {
-                %techRequired = basicSolids
+                %techRequired = solids1956
                 %cost = 0
             }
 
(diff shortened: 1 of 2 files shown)


siimav:
@mattwrobel Since you configured the part, do you happen to remember whether cost per unit was sourced or whether it was just a guesstimate?

nepphhh:
jesus christ I gotta stop accidentally pruning that branch

siimav:
That should have been a separate PR. Besides, you also need to update the JSON file from which the actual config files are generated.

nepphhh:
Sorry, I'll clean this up.

PR #1127, opened 2019-08-20
DECQ Saturn V/Apollo Compatability

lukecologne:
Just copied the values from FASA


siimav:
Has anyone tried out these configs?

PR #1128, opened 2019-08-20
Reduced Nike Part Cost

nepphhh:
Despite the (I assume) historical accuracy of this cost, it's well over double that of the Castor-1 and over seven times as expensive as the AJ2.5. For a reasonable Aerobee 200 analogue (we don't have an AJ60-92, but the AJ10-27 produces comparable thrust), the Nike alone is up to 3/4 the total cost.  Halving the cost will keep it expensive compared to similar motors but not stupidly so.


siimav:
I'm going to merge it soon unless someone finds actual cost figures for that booster.

PR #1142, opened 2019-10-17
Autopilot-resilient rotational locking!

DRVeyl:
On insufficient avionics, prohibit all rotation and translation requests other than FWD/REV ("H"/"N"), even from autopilots.
Use GameEvents to detect modification instead of part counting
Move setup to Awake() and Start() instead of constructor
Stylistic edits (use newer language features)

### DIFF
--- a/Source/Avionics/ControlLocker.cs
+++ b/Source/Avionics/ControlLocker.cs
@@ -1,7 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Text;
-using KSP;
 using UnityEngine;
 using KSP.UI;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


siimav:
Merged in dbf9fa5d20ee7ae65987ad24292f715d3e2df39c

Source/Avionics/ControlLocker.cs
```diff
@@ -13,28 +11,25 @@ class ControlLockerUtils
...
+            maxMass = vesselMass = 0f;  // These are unreliable if the calculation exits early!
 
             if (parts == null || parts.Count <= 0)
                 return false;
+            if (!HighLogic.LoadedSceneIsFlight && !HighLogic.LoadedSceneIsEditor) return false;
 
-            for (int i = parts.Count - 1; i >= 0; --i)
+            int crewCount = (HighLogic.LoadedSceneIsFlight) ? parts[0].vessel.GetCrewCount() : CrewAssignmentDialog.Instance.GetManifest().GetAllCrew(false).Count;
+
+            foreach (Part p in parts)
```
ec429: Umm, I think there was a reason (performance related, probably garbage generation) why `foreach` was avoided here.

DRVeyl: Foreach is safe for List<> types for both GC and execution time.  Also this implementation has previously been capped at one execution/second.

ec429: OK, I just wanted to be sure it had been considered :)

siimav: For reference: https://jacksondunstan.com/articles/3805

PR #1143, opened 2019-10-18
Use ClickThroughBlocker for GUI

todi:
I use this locally to prevent click through on the RP-0 windows.

This would make ClickThroughBlocker a hard dependency instead of a indirect one (e.g through KRASH). I don't know the policy on that so you might not want to merge this. But I thought it might be useful to others.

### DIFF
--- a/Source/Avionics/ModuleProceduralAvionics.cs
+++ b/Source/Avionics/ModuleProceduralAvionics.cs
@@ -3,6 +3,7 @@ using RP0.Utilities;
 using System;
 using System.Linq;
 using UnityEngine;
+using ClickThroughFix;
 
 using static RP0.ProceduralAvionics.ProceduralAvionicsUtils;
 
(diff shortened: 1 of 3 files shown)


siimav:
@pap1723 @MikeOnTea @ec429 or someone else - any opinions on whether we should add ClickThroughBlocker as a build dependency or not?

pap1723:
I am torn on this. Has LGG made ClickThroughBlocker a requirement for KCT
yet? I know he basically does for all his mods. When setting up an RP-1
install, it is basically impossible to not already have it installed so I
think requiring it would be a small impact to the player.

I just hate having tons of dependencies as we are always reliant on them
for all updates.

siimav:
Yep, it's included in KCT now and thus the player will have it anyway. 
I'm not entirely sure on whether it will cause any assembly versioning issues for us in the future.

pap1723:
I am hesitant to allow this as a hard dependency. Thoughts @siimav ??

siimav:
Not sure either. I've occasionally seen other mods break until I install the latest version of blocker or toolbar.

pap1723:
So have I. For example, I cannot use the current Dev version of MechJeb as it is broken with Toolbar Controller

lukecologne:
Hm. Maybe, before you decide, I should say that I am *trying* to redo the GUI with the Unity GUI stuff, and with that, ClickThroughBlocker should (I think) not be required. But don't expect too much, this is my first "real" work on modding in Unity

siimav:
Finally added in 1770c5e1a94e2551776d969c36db2016c215bff5

PR #1150, opened 2019-11-15
No X-1 Shenanigans

nepphhh:
X-1 is now unpressurized. Inaccurate but prevents super ahistorical suborbital, or, worse, orbital shenanigans with the X-1 cockpit.


SerinaTsukaya:
Why not just lock sub-orbital contracts behind the first manned capsule?

siimav:
We're going to handle that slightly differently

PR #1157, opened 2019-11-25
Add ECM and tech tree to LF2 Agena

Capkirk123:
Add ECM and tech tree configs for LF2 Agena that I added to RO
Link to RO engine config: https://github.com/KSP-RO/RealismOverhaul/pull/2148


siimav:
We should also add KCT Resource_Variable for LF2. It's slightly unfortunate that we can't make it global.

PR #1158, opened 2019-11-27
Add ECM and Tree Configs for XLR43

Capkirk123:
Add ECM and Tech Tree configs for the XLR43, the prototype Phase III and IV Navaho boosters. The XLR43-NA-1 was a single chamber derivative of the A-4, and was the basis of the NAA 75-110. It was hot-fired in early 1950, although it never flew. The XLR43-NA-3 was a heavily upgraded variant, and flew in pairs as the XLR71 on Navaho, and when converted to burn Kerosene became the LR43-NA-3 on Altas.
Learned my Lesson, adding tech tree configs before adding the engine to RO to prevent shenanigans.

RO engine config: https://github.com/KSP-RO/RealismOverhaul/pull/2164
ROEngine part: https://github.com/KSP-RO/ROEngines/pull/34


lpgagnon:
Seeing a few problems with the ECMs in my install
- the Navaho-PhaseIII-TP ECM isn't actually showing up in-game for the xlr43 part; I can't tell why
- if it did, the cost would jump from 10k to 43k (same as A6); seems awfully steep for something intended to compete with rd101/102
- the -NA-3 config costs 130k to unlock, which is very high for that point in the tech tree (part of that may be the same missing 30k, but even 100k seems problematic). I'm thinking the Navaho-TP ECM might need to be split in 2, one for the xlr43-na-5 level, one for the lr79/etc level

pap1723:
Needed to manually add due to conflicts with https://github.com/KSP-RO/RP-0/commit/da8ee3897b0afd3a4cae472a32fcceb369051f8a

PR #1162, opened 2019-12-04
Added No avionics power draw

SerinaTsukaya:
Added a condition where procedural avionics that have no controllable mass draw power similar to that of disabled avionics in later tech. rates are subject to change, but it allows the player to create dummy avionics that only serve as a data transmission hub.


SerinaTsukaya:
Merge was premature as code changes in the dll caused a duplicatoin of the RP-1 icon for unknown reasons, will also investigate improvements in terms of weight and cost so that no avionics becomes more than just a more practical/nicer looking 20inch xray core with way worse stats

SerinaTsukaya:
Have not managed to duplicate the bug with duplicate rp-1 icons, will continue to investigate possible changes to the cost and weight.

PR #1173, opened 2019-12-19
Unlock First EVA with any EVA-enabling tech

lpgagnon:
First EVA milestone was recently hidden behind 'must unlock 2nd gen
capsules', to protect players from thinking they can do it with
Mercury

But 2ndgen capsules aren't the only way to go EVA, prototype spaceplanes
are an alternate path. It'd be silly if the player could go EVA and not
hit the milestone.


lpgagnon:
my own preference would actually be not to hide the contract behind techs at all, because it hides goals from the player...

PR #1181, opened 2020-01-06
Nerf starting level science core

nepphhh:


siimav:
I would personally prefer to set it (and the prefab core) to 50kg but increase the engine mass by 5kg.

nepphhh:
That sounds excellent. Worth noting that reduces the total mass by 5 kg. Is that intended?

siimav:
Alternatively we could still increase the engine mass; leave the starting core as-is; and add a procedural science core upgrade to the first avionics node that is currently empty anyway. The upgrade would bring the mass down from 60kg to 40kg, for example.

MikeOnTea:
I'd like to somewhat question the density and mass changes. Procedural avionics mass is basically balanced to not include case/structural mass. Now it looks like visually our SR core models also don't include a nose cone, if that's also true for its mass, not sure what makes it so heavy. Either there was a huge mass progress between that and early satellite avionics like explorer (However currently our normal avionics don't represent that), or it contains stuff that simple sat avionics - which the science core is primarily meant to represent - doesn't need. In any case that means i'd probably prefer to have it rather light. That said, we could make the first config heavy and assume it contains SR guidance stuff and nobody's going to fly SRs anymore when they reached orbit.

siimav:
Since we haven't managed to source any actual mass figures for Bees, then I'd suggest going for values that make sense for the gameplay and balancing. I have always assumed that the early couple of levels are more like SR cores rather than Explorer-like science cores that run experiments in LEO. 
Adding an avionics upgrade to the first node would give another way to improve SRs in addition to better engines and tanks. Besides, we have an empty node and nobody likes those. :)

siimav:
Implemented in #1194

PR #1194, opened 2020-01-26
Early Science-Core proc avionics overhaul

siimav:
Changed the progression to be closer to prefab parts. Also added an additional level to the postWarAvionics node which was currently empty.
The mass figures are now:
* lvl 0 (starting) - 60kg
* lvl 1 (new) - 40kg
* lvl 2 (Sputnik) - 20kg
* lvl 3 (Explorer) - 10kg
* lvl 4 (Early controllable probe core) - 2kg

### DIFF
--- a/GameData/RP-0/Tree/ProceduralAvionics.cfg
+++ b/GameData/RP-0/Tree/ProceduralAvionics.cfg
@@ -544,25 +544,34 @@
 			{
 				name = start
 				techLevel = 0
-				massFactor = 20
+				massFactor = 58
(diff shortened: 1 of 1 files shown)


MikeOnTea:
Just for future reference, according to https://solarsystem.nasa.gov/news/285/10-things-to-know-about-explorer-1-americas-first-satellite/ 40 % of explorer's mass were batteries, that amounts to ~5.6 kg, but i see you already lowered its mass from 10 to 4 kg which is a much better value i think.

PR #1203, opened 2020-02-17
Remove 'solar power' param from Comm Sat contracts

lpgagnon:
Because the requirement is just a 'gotcha'.
- All those contracts are gated behind First Scientific Satellites,
  so solar panels are necessarily available
- Without any specific power output needed, the requirement can
  be satisfied with a single tiny panel that doesn't actually provide
  enough power for anything

So the requirement doesn't increase contract difficulty in any way
other than punishing players who fail to notice it

### DIFF
--- a/GameData/RP-0/Contracts/Early Satellites/Communications Satellite - Early.cfg
+++ b/GameData/RP-0/Contracts/Early Satellites/Communications Satellite - Early.cfg
@@ -92,15 +92,6 @@ CONTRACT_TYPE
 		}
 		PARAMETER
 		{
-			name = HasSolarPower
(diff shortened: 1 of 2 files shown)


siimav:
I'm happy to remove it from contracts other than *First Communications Satellite*. Maybe this one should still have the solar panel requirement since control of the sat isn't given over to a customer. Btw, the contract description also says to include a solar panel.

lpgagnon:
push -f: reverted change to First Comm Sat (I'm not convinced by the 'not given over to customer' argument, but don't care that much), and fixed the description that still talked about power generation. (and rebased to latest master)

PR #1206, opened 2020-02-22
Generic value contracts

nepphhh:
This pull request removes explicit reference to `rewardFunds`, `advanceFunds`, and similar nodes in all contract definitions. A MM patch catches references to a new field, `genericValue`, which abstracts the total reward offered by a contract, and patches it before ContractConfigurator sees it. 

Besides making contract value much more human-readable (I've taken the occasion to simplify what had become layers and layers of modifications to some original contract reward) and reducing the number of lines by a lot, it lays the groundwork for making quarterly budgets a toggle in the difficulty settings. Currently, adding the empty folder `RP1DoUseBudgets` toggles reward paradigm between cash and reputation. The details of this switch over are highly tweakable.

As the MM patch provides a much cleaner way to mass-edit contract value, I have removed the myriad references to `RP0:globalHardContractMultiplier`. Instead of having this lengthy reference be present in every single line of reward definition, it can be a simple `* 2.4` addition to a single line in the MM patch. 

The current 2.4x multiplier has been incorporated into `genericValue`. I've gone over my changes twice making sure I didn't change the value of any contracts (except for one series in which there is a 0.8% difference in value). Some advance/reward splits may have been *slightly* altered, but the total value is still the same.

This PR will be form the base upon which I'll develop the new sounding rocket contracts as directed by Siimav.

### DIFF
--- a/GameData/RP-0/Contracts/Advanced Satellites/Earth Imaging Satellites.cfg
+++ b/GameData/RP-0/Contracts/Advanced Satellites/Earth Imaging Satellites.cfg
@@ -27,12 +27,8 @@ CONTRACT_TYPE
 	
 	
 	prestige = Trivial       // 1.0x
-	advanceFunds = 20000 * @RP0:globalHardContractMultiplier
+	genericValue = 72000
 	rewardScience = 0.6
-	rewardReputation = 5
-	rewardFunds = 10000 * @RP0:globalHardContractMultiplier
-	failureReputation = 5
-	failureFunds = @advanceFunds * 0.5
 	
 	// ************ REQUIREMENTS ************
 	
--- a/GameData/RP-0/Contracts/Milestones/First Science Sat.cfg
+++ b/GameData/RP-0/Contracts/Milestones/First Science Sat.cfg
@@ -27,12 +27,8 @@ CONTRACT_TYPE
 		
 	// ************ REWARDS ************
 	prestige = Exceptional   // 1.5x
-	advanceFunds = 117779 * @RP0:globalHardContractMultiplier
-	rewardScience = 0
-	rewardReputation = 40
-	rewardFunds = 55556 * @RP0:globalHardContractMultiplier
-	failureReputation = 40
-	failureFunds = @advanceFunds * 0.5
+	genericReward = 416000
+	genericAdvanceMultiplier = 1.0 / 3.0
 		
 	// ************ REQUIREMENTS ************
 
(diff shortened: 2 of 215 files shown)


GameData/RP-0/Contracts/Milestones/First Science Sat.cfg
```diff
@@ -27,12 +27,8 @@ CONTRACT_TYPE
...
 	// ************ REWARDS ************
 	prestige = Exceptional   // 1.5x
-	advanceFunds = 117779 * @RP0:globalHardContractMultiplier
-	rewardScience = 0
-	rewardReputation = 40
-	rewardFunds = 55556 * @RP0:globalHardContractMultiplier
-	failureReputation = 40
-	failureFunds = @advanceFunds * 0.5
+	genericReward = 416000
+	genericAdvanceMultiplier = 1.0 / 3.0
```
siimav: Does that mean that advance is 1/3 of the completion reward? If yes, then the advance used to be twice as big as the reward. Why are you changing that?

nepphhh: Ah, my mistake.

GameData/RP-0/Contracts/Advanced Satellites/Earth Imaging Satellites.cfg
```diff
@@ -27,12 +27,8 @@ CONTRACT_TYPE
 	
 	
 	prestige = Trivial       // 1.0x
-	advanceFunds = 20000 * @RP0:globalHardContractMultiplier
+	genericValue = 72000
```
Standecco: shouldn't it be
> genericReward = 72000?

GameData/RP-0/Contracts/Milestones/First Science Sat.cfg
```diff
@@ -27,12 +27,7 @@ CONTRACT_TYPE
 		
 	// ************ REWARDS ************
 	prestige = Exceptional   // 1.5x
-	advanceFunds = 117779 * @RP0:globalHardContractMultiplier
-	rewardScience = 0
-	rewardReputation = 40
-	rewardFunds = 55556 * @RP0:globalHardContractMultiplier
-	failureReputation = 40
-	failureFunds = @advanceFunds * 0.5
+	genericReward = 416000
```
siimav: With the default 2/3 ratio, the advance is 10k smaller than it used to be.

siimav:
Before this PR has any chance of getting merged, I would need someone to review all neph's calculations. Since we've already found a couple of mistakes, it's safe to assume that there are more of them hidden inside this PR.

PR #1210, opened 2020-03-02
Avionics check parameter

nepphhh:
Well-tested


siimav:
Why is this PR changing other non-relevant files? It should be structured in the same way as the existing codebase is. Also please do not commit DLLs.

siimav:
Oh, and to make matters worse - your PR completely nukes the Downrange distance parameter carryover that I recently implemented for distance records.

nepphhh:
Sorry, not familiar with GH protocol. Un-consolidated the other parameter definitions & un-recompiled the .dlls. 

I don't believe I made any changes to the downrangeDistance parameter any more, but I was unfamiliar with your recent change to start with, so you'd have to check against it for me.

siimav:
Changed files list is still showing 2 DLL modifications. 
You should split it into 2 files for it to be similar to the other custom CC parameters.

nepphhh:
Dang, I'll just start a new branch. 

Is there any particular reason that we don't want to put the parameter factories and the parameter param classes in the same file? When trying to parse the exising CC parameters, I found having them in the same file improves readablility substantially. Change bad?

siimav:
In general you do not slap different classes together into the same file unless one of them is a small model that is only used by the other type. Otherwise you will sooner or later end up with a mess similar to [that](https://github.com/Starwaster/DeadlyReentry/blob/master/Source/DeadlyReentry.cs).

PR #1211, opened 2020-03-02
Part tooling setting

EdwinHoksberg:
Make part tooling optional. First PR here so let me know if something can be improved 😄 

Tooling is disabled on easy mode and enabled for normal and up, might want to look at that to make sure it's balanced.


EdwinHoksberg:
I wrote my code quite a while ago so I see there were some changes made in the `ToolingDatabase.cs`, namely the `GetToolingLevel ` method used to return an Enum but now it is just an integer(in commit a107763f0abe51d1ada9de388c690e3ed1919b2c). I couldn't infer from the code what is the maximum value for this int(so I can return it when tooling is enabled). Does anyone know what that value is? Also interested to find out why this was changed in the first place, an enum seems much nicer than magic numbers.

siimav:
> Does anyone know what that value is?
My guess is that the int signifies how many tooling parameters are met. For example avionics is tooled based on tech, length, width and tonnage.

EdwinHoksberg:
> > Does anyone know what that value is?
> 
> My guess is that the int signifies how many tooling parameters are met. For example avionics is tooled based on tech, length, width and tonnage.

Yes that's what I thought too, if i just return `parameters.Length` it should work. I tried to get this project to build in linux again so I can test it but no luck yet.

siimav:
Also, disabling the tooling feature entirely on Easy difficulty doesn't sound right to me.

EdwinHoksberg:
> > Also, disabling the tooling feature entirely on Easy difficulty doesn't sound right to me.
> 
> Same opinion. In easy you should have enough money to be able to tool more liberately, so why remove it completely? It's a fundamental feature to push for reusing part designs and thus think/plan a bit more about what you want to use where.

Agreed, I changed updated this.

PR #1214, opened 2020-03-05
Add AvionicsCheck parameter round two

nepphhh:
electric boogaloo


nepphhh:
Thanks, siimav, for the feedback on the last one. Learning to work with a shared codebase is a steep learnin curve

PR #1215, opened 2020-03-06
Hide deprecated parts from the editor

Standecco:
No need to keep showing them if they're deprecated and shouldn't be used anyway. This will make the part list less cluttered. Completes https://github.com/KSP-RO/RealismOverhaul/pull/2222


Bornholio:
Looks good, I would suggest a mechanism like NoNonRP0 folder or better yet settings switches for hiding these part.

Standecco:
The issue with a NoNonRP0 system would be that it would be disabled by default; these old parts don't have anything to be preferred over the new ones, so they aren't needed anywhere but in old crafts.

PR #1219, opened 2020-03-10
No non proc 1.8 test

nepphhh:
lets you murk non proc tanks with a magic folder. configs haven't been generated yet, make sure that any merges after this one have is_non_proc = False in the .json if they were made with a part browser older than this one


siimav:
Hmm, that got automatically closed after I deleted the 1.8-TEST branch

PR #1224, opened 2020-03-25
Orion

Capkirk123:
Add ECMs, tech nodes and avionics to Orion


DocRockwell:
Been testing this in my career game. Looks great! Just a few notes:
-solar panels are in the main Electronics Research branch rather than the Power Generation subbranch
-the RCS thrusters should, imo, be moved back from Near Future Flight Control to as early as Space Station Attitude Control, which is when lvl 4 RCS tech is unlocked and already provides higher ISP generic MMH+MON3 thrusters. 
-parachutes don't appear in game at all (I think the problem is on the ROC side though)

Capkirk123:
Solar panels have been moved. The RCS is associated with the service module, so I'll keep it where it is. If they were separate pieces then I would probably move them earlier. I'll look in the parachute issue too.

PR #1225, opened 2020-03-27
KAS/KIS compatibility

Capkirk123:


Source/Tech Tree/Parts Browser/data/KIS.json
```diff
@@ -0,0 +1,452 @@
...
+        "ro": true,
+        "orphan": false,
+        "rp0_conf": true,
+        "spacecraft": "",
+        "engine_config": "",
+        "upgrade": false,
+        "entry_cost_mods": "",
+        "identical_part_name": "",
+        "module_tags": [
+            "",
```
Standecco: Should be deleted.  This blank field fails config generation

PR #1237, opened 2020-04-11
Fix payload contracts by requiring 0.01 less than stated

leudaimon:
This fixes the rounding errors from different floating point approximations in CC and RF.


leudaimon:
Remade the PR because I had started from the wrong branch. Should be correct now

PR #1240, opened 2020-04-13
Move Cavea-B to the same node as bipropellant RCS

leudaimon:
Given this fuel was never really used, it was placed to early in the tech tree given its performance.
Please confirm that this is all that needs to be changed to implement this.


siimav:
Looks like nobody has screamed in horror so I'm going to merge this

PR #1241, opened 2020-04-13
Move hydrazine back a node

nepphhh:
Hand-in-hand with https://github.com/KSP-RO/RP-0/pull/1240.

Once Cavea-B is moved back, the "stability" node will be nearly empty. NASA document 19680006875 says that hydrazine monoprop was fired in space in 59 as a test but not used operationally until 1961 with Ranger and Mariner. This moves it from the '59 node to the '61 node.

### DIFF
--- a/GameData/RP-0/Tree/TREE-Engines.cfg
+++ b/GameData/RP-0/Tree/TREE-Engines.cfg
@@ -764,7 +764,7 @@
 
             @CONFIG[Hydrazine]
             {
-                %techRequired = earlyFlightControl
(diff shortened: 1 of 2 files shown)


pap1723:
Thoughts on this? It makes sense to me and makes the SUPER early Lunar Landers much more difficult.

Capkirk123:
Seems reasonable, HTP will still work fine for attitude control for most things, and it'll make super early lunar orbiters/ interplanetary orbiters much harder.

PR #1242, opened 2020-04-14
Warning that the sat needs to work for 91 days

glilienthal:
Without the warning the player could think that 120min would be enough. 

Alternative: require not all, but only some of the Cosmic ray data sent, so that the contract can be completed in 120min. But I don't know how to do that.


leudaimon:
I'm quite sure this is not the intended behavior for this contract. The issue is that there is no easy way to make it work with kerbalism right now. I don't think changing the description fixes it.

siimav:
Yeah, this isn't how we're going to solve that issue. Ideally we'll add a reference to KerbalismContracts at a later that which allow that parameter to complete as soon as the experiment starts running.
If you want to add an intermediate solution then append a note about cheat completing it to the description in case Kerbalism is installed.

glilienthal:
Makes sense. I know wrote a note that gives moral licence to cheat-complete the contract.

PR #1243, opened 2020-04-14
Fix the TechTree Icons for Baloon tank upgrades

leudaimon:
The Tech Tree Icons were not matching the tank definitions. 
Looks like there was some Copy-Paste error, the first upgrade had the technology for the base balloon tank, and the second upgrade had the technology of the first upgrade.

### DIFF
--- a/GameData/RP-0/Tree/ResourceTechs.cfg
+++ b/GameData/RP-0/Tree/ResourceTechs.cfg
@@ -370,7 +370,7 @@ PARTUPGRADE
 {
 	name = RFTech-Balloon-SteelAlCu
 	partIcon = RFTank-I
-	techRequired = materialsScienceSatellite
+	techRequired = materialsScienceSpaceplanes
 	entryCost = 0
 	cost = 0
 	title = Balloon Tank Upgrade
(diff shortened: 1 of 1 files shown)


siimav:
Should AlCu really be in the spaceplane matsci node? Regular tanks get it at materialsScienceAdvCapsules already.

leudaimon:
I just updated the icons to match the PARTUPGRADE nodes. 
However, I noticed that the Separate Structure tanks get AlCu at materialsScienceAdvCapsules and Integrated ones get only at materialsScienceLunar, while the balloon ones get at materialsScienceSpaceplanes. So is it supposed to be consistent, or do these come indeed at different eras?

siimav:
@Capkirk123 @nepphhh

leudaimon:
By the way, looking at the upgrade's descriptions, I'm quite sure they are indeed supposed to be in these more advanced nodes. Balloon AlCu is stated to be first used in Centaur-G (first launch 1984), and AlLi was used only in later versions of Atlas III and Atlas V.

nepphhh:
@siimav it should not. Must have been a mistake when writing the RP-1 configs. We were planning to put Al-Cu in the Spaceplanes Mat Sci node (1981).

Capkirk123:
The material upgrades are intended to be staggered, since it is much easier to switch materials on the simpler tanks than it is on more complex structures, which is reflected in real life.

PR #1244, opened 2020-04-14
Per-difficulty-preset configurable contract deadlines

siimav:
Added a slider to the RP-1 difficulty settings which will be used to give Easy and Normal difficulty presets longer contract deadlines than Moderate. Currently those are set to 1.5x for Normal and 2x for Easy.
The reasoning behind this change is that the deadlines are currently balanced for expert players and coupled with random failures, beginners have too high chance of failing big milestone contracts like FSO.  In addition to that the short deadlines and career-ending penalties will scare them away from the high-paying milestone contracts and instead start grinding other low-paying filler contracts.

I would like to get some feedback on this PR from other RP-1 players.


norcalplanner:
My vote would be 1.25x for normal and 1.5x for easy. Some vestige of the "I don't know if I'll be able to meet this deadline" panic needs to remain at all difficulty levels.

leudaimon:
I like this idea, and I think it could be paired with smaller penalties as well. I think we as a whole have been trying to balance the game on very experienced and min-maxing players. RO/RP-1 introduce so many mechanics, that I think it makes sense to have the easy difficulty really easy, where the player can learn the mechanics and do things on their own pace. Then normal can be still more forgiving but requiring some optimization.
Another suggestion I would have, is that given these reductions in difficulty, the rewards and advances could be slightly reduced in the lower difficulties as well.

siimav:
> My vote would be 1.25x for normal and 1.5x for easy. Some vestige of the "I don't know if I'll be able to meet this deadline" panic needs to remain at all difficulty levels.

Yeah, my initial 1.5x and 2x multipliers might have been a bit too much. Although I'm not sure if 1.25x and 1.5x would have much of an effect on encouraging people to accept these contracts earlier. Any more suggestions on what these multipliers should be?

norcalplanner:
Somewhere in between then, such as 1.3 and 1.6? And perhaps 0.8 for hard? [insert evil grin]

siimav:
Changed the values to:
* 1.7 for easy
* 1.3 for normal
* 0.8 for hard

PR #1263, opened 2020-05-07
Untagging unbalanced parts

Capkirk123:
Remove the RP-0 tag from prefab avionics and the Airstream Fairings. 

Parts untagged:
AE-FF1 Airstream Protective Shell
AE-FF2 Airstream Protective Shell
AE-FF3 Airstream Protective Shell

Reason:
No tooling support, and weight half as much as the Proc Fairings

Parts Untagged:
Early Controllable Core (Stock)
Guidance Unit Early (1m) (Stock)
Guidance Unit Early (2m) (Stock)
Guidance Unit Early (3m) (Stock)
Guidance Unit Starting (1m) (Stock)
Guidance Unit (1m) (Stock)

Reason:
Superseded by, and unbalanced compared to, procedural avionics.


jwvanderbeck:
Github makes it really hard to see what parts this actually affects :(

leudaimon:
I would suggest the avionics parts based on real rockets should be kept and reconciled with the procedural avionics. The reason is twofold: 
* AFAIK many of these rockets/stages were used to balance procedural avionics in the first place
* These parts were created based on the real rockets - if procedural avionics is way off of these parts something is wrong with them.

The parts that apply to this argument are:
Able Avionics Package (Stock)
Agena Avionics Package (Stock)
Centaur Avionics Package (Stock)
Delta Avionics Package (Stock)
Mariner Core (SXT)
Ranger Block III core (Stock)
Saturn IB/V Instrument Unit (Stock)
Surveyor Core (Stock)

Capkirk123:
Ok, I'll just untag the generic avionics parts. Those seem to be the trap for newer players anyway, and are the most unbalanced.

PR #1264, opened 2020-05-08
Tech tree adjustments for Kerbalism parts

Standecco:


siimav:
Could you add configs too?

PR #1271, opened 2020-05-20
Retirement + R&R rework

siimav:
The previous system was flawed because it used the number of flights flown for calculating retirement. This meant that someone who had flown 10 plane flights got almost nothing for getting to orbit compared to a naut whose first flight was orbital. New system counts each situation individually and thus new achievements will still advance the retirement substantially. There's now also a cap of 15 years that can get added to the base retirement date.

For R&R the flight duration will affect the total time non-linearly now. 

Big thanks to @leudaimon who ran the spreadsheets and figured out the new balance. For future reference, the speadsheet itself is [here](https://docs.google.com/spreadsheets/d/1YE4-OhompZlQO4k7kb1a3zuoxmR6tYJ7d9h_sOfepXo).


fat-lobyte:
Would there be a way to get a little more info into the Astronaut tab about the retirement status? For example what the "Not Later Than" date for the cap, and how much a flight could push that back?

siimav:
> Would there be a way to get a little more info into the Astronaut tab about the retirement status? For example what the "Not Later Than" date for the cap, and how much a flight could push that back?

I honestly have no idea how to give the player some meaningful feedback on how different flights could affect the retirement increases. On the other hand I at least added *+X days* information to the dialog that pops up after each flight.

fat-lobyte:
> On the other hand I at least added +X days information to the dialog that pops up after each flight.

This is already very helpful, thanks!

Currently, there's the NET (Not Earlier Than) date for the retirement when you go into the astronaut tab of the RP-1 menu and click on the Astronaut. Maybe there could also be a NLT (Not Later Than) date right next to it?

PR #1273, opened 2020-05-23
RSB + KK SpaceX pack avionics support

pt1243:
Adds avionics control to RealScaleBoosters and KK's SpaceX mods

### DIFF
--- a/GameData/RP-0/Avionics.cfg
+++ b/GameData/RP-0/Avionics.cfg
@@ -935,7 +935,7 @@
 }
 
 // FIXME: Do this for other small upper stages.
-@PART[FASAAgenaProbe|RP0probeAvionics1-5m|RSBtankAgenaD|bluedog_agenaProbeCore|RSBtankAgenaD]:FOR[RP-0]
+@PART[FASAAgenaProbe|RP0probeAvionics1-5m|RSBtankAgenaD|bluedog_agenaProbeCore]:FOR[RP-0]
 {
 	MODULE
 	{
--- a/GameData/RP-0/Avionics.cfg
+++ b/GameData/RP-0/Avionics.cfg
@@ -923,14 +923,23 @@
 //  Centaur astrionics.
 //  Centaur guidance provides control for both Centaur and Atlas stages
 
-@PART[RP0probeAvionics3-05m|RSBtankAtlasCentaur]:FOR[RP-0]
+@PART[RP0probeAvionics3-05m]:FOR[RP-0]
+{
+	MODULE
+	{
+		name = ModuleAvionics
+		massLimit = 45.0
+		toggleable = true
+		disabledkW = 0.02
+	}
+}
+
+@PART[RSBtankAtlasCentaur]:FOR[RP-0]
 {
 	MODULE
 	{
 		name = ModuleAvionics
 		massLimit = 600
-		toggleable = true
-		disabledkW = 0.02
 	}
 }
 
--- a/GameData/RP-0/Avionics.cfg
+++ b/GameData/RP-0/Avionics.cfg
@@ -921,7 +921,7 @@
 }
 
 //  Centaur astrionics.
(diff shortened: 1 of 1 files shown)


GameData/RP-0/Avionics.cfg
```diff
@@ -921,21 +921,21 @@
...
-		massLimit = 45.0
+		massLimit = 600
 		toggleable = true
 		disabledkW = 0.02
 	}
 }
 
 // FIXME: Do this for other small upper stages.
-@PART[FASAAgenaProbe|RP0probeAvionics1-5m|RSBtankAgenaD|bluedog_agenaProbeCore]:FOR[RP-0]
+@PART[FASAAgenaProbe|RP0probeAvionics1-5m|RSBtankAgenaD|bluedog_agenaProbeCore|RSBtankAgenaD]:FOR[RP-0]
```
siimav, line 938: Duplicate now

pt1243: Should the configs for the new parts be in separate entries from the existing parts? (sorry, not sure exactly what you mean @siimav)

siimav: `RSBtankAgenaD` is twice on that same row

pt1243:
Ah, my bad - [this commit (3e29dbc)](https://github.com/KSP-RO/RP-0/pull/1273/commits/3e29bdcfd766beafeaed4624d0c40c7205da94ed) should fix that

GameData/RP-0/Avionics.cfg
```diff
@@ -921,14 +921,14 @@
...
+//  Centaur guidance provides control for both Centaur and Atlas stages
 
-@PART[RP0probeAvionics3-05m]:FOR[RP-0]
+@PART[RP0probeAvionics3-05m|RSBtankAtlasCentaur]:FOR[RP-0]
 {
 	MODULE
 	{
 		name = ModuleAvionics
-		massLimit = 45.0
+		massLimit = 600
```
siimav, line 931: That's actually a significant change for the existing RP-1 avionics part. I wonder if it would be better if RSB config would be separate. Or how do those 2 parts compare to each other mass-wise?

pt1243: The dry mass of the RSB model with one engine and including the 600t guidance capability is 2405 kg - slightly less than the 2245-ish it weighs according to [the Atlas V user guide from ULA](https://www.ulalaunch.com/docs/default-source/rockets/atlasvusersguide2010.pdf), but still within the range.

The Saturn V RP-1 avionics part controls 3000t with a mass of 2t, or 1.5t per kg. Scaling this down to a part mass of 300kg of the existing RP-1 Centaur avionics would mean controlling 450t. I don't think its unreasonable to assume that the difference between technology which flew in the 1970's vs the 2000's could make up for that gap.

pt1243:
@siimav are there any other changes which need to be made?

GameData/RP-0/Avionics.cfg
siimav: My issue is that you're also changing a part called `RP0probeAvionics3-05m` which lies in the `Improved Avionics (1962-1963)` tech node. So if your part is unlocked at a later node then it shouldn't change the config for `RP0probeAvionics3-05m`.

pt1243: Didn't notice that - I'll separate them into two different configs.

GameData/RP-0/Avionics.cfg
```diff
@@ -921,7 +921,7 @@
 }
 
 //  Centaur astrionics.
-//  Approximates the maximum current Centaur + payload combined mass (23t + 25t).
+//  Centaur guidance provides control for both Centaur and Atlas stages
```
siimav, line 924: This comment change doesn't fit for `RP0probeAvionics3-05m`

PR #1281, opened 2020-05-31
Planet orbits

leudaimon:
Fixes the payout from most planetary orbit contracts, which were missing the @RP0:globalHardContractMultiplier and adds a new planetary orbit contract for the 7 main planets.
Also changes the Requirement for the flybys Mercury, Ceres, Vesta and Jupiter, to be either Mars or Venus instead of only Mars.

These repeatables have slightly less payout than the first orbit "milestone" contracts, and add a high inclination and apoapsis (5 * planet radius for the rocky planets, and a higher preset "reasonable" requirement for the gas giants) requirements.
As they require a new vessel, only one of each can be launched at a time, avoiding a shotgun into each planet every window. The intention is to stimulate people to return to the planets with more advanced experiments later on after the first orbit.

### DIFF
--- a/GameData/RP-0/Contracts/Flyby Contracts/Ceres Flyby.cfg
+++ b/GameData/RP-0/Contracts/Flyby Contracts/Ceres Flyby.cfg
@@ -37,11 +37,24 @@ CONTRACT_TYPE
 	// ************ REQUIREMENTS ************
 
 	REQUIREMENT
-	{
-		name = CompleteContract
-		type = CompleteContract
-		contractType = flybyMars
-	}
+    {
+        name = Any
+        type = Any
+        
+        REQUIREMENT
+	    {
+		    name = CompleteContract
+		    type = CompleteContract
+		    contractType = flybyMars
+	    }
+	    
+	    REQUIREMENT
+	    {
+		    name = CompleteContract
+		    type = CompleteContract
+		    contractType = flybyVenus
+	    }
+    }
 
 	// ************ PARAMETERS ************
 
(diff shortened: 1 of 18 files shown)


norcalplanner:
Just a quick thought - the ScanSat contracts already provide an incentive for low eccentricity polar orbits, so make sure these are complementary rather than duplicative

jwvanderbeck:
> Just a quick thought - the ScanSat contracts already provide an incentive for low eccentricity polar orbits, so make sure these are complementary rather than duplicative

Does anyone actually do the scansat contracts? Last time I looked at them they seemed to require a very large amount of scan data which I just assumed would take years to accumulate at best.  They seemed a lot harder than normal orbital science.

norcalplanner:
I always do, but I acknowledge that I may be an outlier. It just makes sense to me to definitively determine where biomes are before doing extensive landings.

jwvanderbeck:
Oh I definitely agree on doing scans for landers.  I just have never bothered with the actual contracts as they seemed to require a large amount of work and I was never sure I could actually get them done in time.

leudaimon:
Regarding the orbital parameters, SCANSat requires much more circular orbits than the ones I'm asking for in these contracts.
About the usefulness of the SCANSat contracts, I have completed the first altimeter contract for the Earth and Moon without any effort, it completes fast and easy. I didn't get to the more advanced ones yet, but I expect it to be equally simple?

norcalplanner:
Sounds like they're complementary! Love the idea, btw.

pap1723:
I always do the ScanSat contracts, but they need to be fixed in RO-Kerbalism (unless they have since I have been around more) as they are not nearly worth it in terms of total science gained.

leudaimon:
I had the same impression about the science amount, but iirc @Standecco already buffed the science from ScanSat experiments a few weeks ago.

PR #1282, opened 2020-05-31
Update Orbital Recovery.cfg

smartdummies:
Increase the required parameters for completion based on discussions in discord thread. Fist completion is unchanged, second requires a "typical" crewed orbit (set to min paerigee of 200km) to test in actual situation after diong initial test. Thrid requires a 1000km apogee to test higher safety margins

### DIFF
--- a/GameData/RP-0/Contracts/Milestones/Orbital Recovery.cfg
+++ b/GameData/RP-0/Contracts/Milestones/Orbital Recovery.cfg
@@ -28,9 +28,9 @@ CONTRACT_TYPE
 	// ************ REWARDS ************
 	prestige = Significant   // 1.25x
 	advanceFunds = 50000 * @RP0:globalHardContractMultiplier
-	rewardScience = Random(0.6, 1.8)
+	rewardScience = 2
 	rewardReputation = 10
-	rewardFunds = (Random(40000, 60000)) * @RP0:globalHardContractMultiplier
+	rewardFunds = 45000 * @RP0:globalHardContractMultiplier
 	failureReputation = 10
 	failureFunds = @advanceFunds * 0.5
 
@@ -131,9 +131,9 @@ CONTRACT_TYPE
 	// Same as initial return
 	prestige = Significant   // 1.25x
 	advanceFunds = 50000 * @RP0:globalHardContractMultiplier
-	rewardScience = Random(0.6, 1.8)
+	rewardScience = 2
 	rewardReputation = 10
-	rewardFunds = (Random(40000, 60000)) * @RP0:globalHardContractMultiplier
+	rewardFunds = 50000 * @RP0:globalHardContractMultiplier
 	failureReputation = 10
 	failureFunds = @advanceFunds * 0.5
 
@@ -242,9 +242,9 @@ CONTRACT_TYPE
 	// Same as initial return
 	prestige = Significant   // 1.25x
 	advanceFunds = 50000 * @RP0:globalHardContractMultiplier
-	rewardScience = Random(0.6, 1.8)
+	rewardScience = 2
 	rewardReputation = 10
-	rewardFunds = (Random(40000, 60000)) * @RP0:globalHardContractMultiplier
+	rewardFunds = 55000 * @RP0:globalHardContractMultiplier
 	failureReputation = 10
 	failureFunds = @advanceFunds * 0.5
 
--- a/GameData/RP-0/Contracts/Milestones/Orbital Recovery.cfg
+++ b/GameData/RP-0/Contracts/Milestones/Orbital Recovery.cfg
@@ -123,7 +123,7 @@ CONTRACT_TYPE
 	maxExpiry = 0
 	maxCompletions = 1
 	maxSimultaneous = 1
-	deadline = 365  // 1 year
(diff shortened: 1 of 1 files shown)


jwvanderbeck:
The requirement for suborbital on the first contract seems off to me.  I would say have the speed requirement for sure but it shouldn't really matter if I am suborbital, or go all the way to orbit.

smartdummies:
I have let the first one unchanged from the current contracts. Once that has been completed it requires higher orbits to complete. While I may disagree with the allowance of the 6500m/s trigger as it can be gamed quite easily, form a historical perspective it makes sense as these were more for ICBM development. So for the contract progression the flow is test at a speed that requires a heat shield to survive re-entry, which is less than orbital velocity (this is the minimum 6500m/s). Next require re-entry from the "typical" crewed orbit altitude of 200km (this is the perigee required to complete the contract) as this is greater than the initial contract and makes sense to me for testing the safety of the heat shields. I included the stable orbit requirement as well to help prevent the gaming that I have done in the past on the second and third contracts. Finally it requires a 1000km apogee as a test at a high end of the spectrum to test the limits of the safety margins (this has a 10 minute orbit check instead of the standrd 2 minute used commonly in RP-0 to require to to get to a much higher apogee before returning) as this is common throughout space flight testing. The inclusion I put in for the second and third contracts to block them if you complete the first crewed orbit is that these would be tests that would be done before the first flight as these would be safety related. Once you do the first crewed orbit you do not need to do the tests. Just my thoughts and reasoning for this and I hope that helps you understand the "why" of what I did here and apologies if it is rambling.

jwvanderbeck:
My only point was I don't see a reason to make the player fail the first contract if they put the test vehicle all the way into orbit. Either way should be fine.

A couple of other notes:
1) These don't seem to honor the contract deadline slider
2) If you fail one in the series (which I did on the second because I didn't notice that it wasn't honoring the slider), then you can't retry it or continue the series.

smartdummies:
it will not fail, it remains unchanged, getting to orbit will exceed the 6500m/s requirement so the contract can be completed successfully.

Regarding the contract deadline slider - I am not sure what you are referencing. I did most of this this by copying the contract and then editing the requirements (I have a limited understanding of the workings of Contract Configurator and have not found comprehensive documentation so I find similar requirements and use that as a template for the requirements I wanted). If I need to add something so that it is offered again if you fail let me know where I can look to see where this is done (pointing me to a contract that is setup for this will work) and I will gladly make necessary changes.

jwvanderbeck:
> it will not fail, it remains unchanged, getting to orbit will exceed the 6500m/s requirement so the contract can be completed successfully.

Unless you changed it after I tested one of the requirements was **suborbital situation** so when I got into orbit the contract did not complete.

jwvanderbeck:
As for contract deadline sliders, here is an example, just grabbed one at random but it should be the same, with the only difference between the "base" time you desire:

https://github.com/KSP-RO/RP-0/blob/master/GameData/RP-0/Contracts/Early%20Satellites/First%20Com%20Sat.cfg#L23

For being able to try and complete the series.. That I don't know. @siimav ?

smartdummies:
Change for the slider added

siimav:
That looks correct to me now. 
@jwvanderbeck Not entirely sure why it didn't complete for you when returning for orbit. The parameter should complete itself and stay completed as soon as you go over 150km while travelling at >6500m/s.

Another odd thing about this contract is that the sci and funds rewards are randomized while the actual requirements are always the same. Should probably get rid of that since it doesn't make any sense.

smartdummies:
For science I suggest 2 (I know it is at the high end, but less than that makes it feel meaningless) for all three contracts,  and 45000, 50000, 55000 * multiplier for the funds as this increases the funds for the successive contracts while keeping the total amount paid equivalent to the average of doing all three contracts with the randomized amount.

GameData/RP-0/Contracts/Milestones/Orbital Recovery.cfg
```diff
@@ -110,3 +99,226 @@ CONTRACT_TYPE
...
+		}
+		PARAMETER
+		{
+			name = Orbit
+			type = Orbit
+			minPeA = 200000
+			disableOnStateChange = true
+			title = Achieve Orbit within the parameters
+			
+			//Do we wat to kee this? I lean to yes but is not really necessary
```
siimav, line 189: Might as well keep it since it will prevent you from doing a 180 and burning retrograde with the same stage. Not sure if increasing the duration to 10 minutes for the 3rd contract does anything useful though.

smartdummies: I put in the 10 minute duration as a means to prevent achieving the high apogee and then deorbiting. 10 minutes will at least ensure that they achieve an actual altitude sufficient to do a more strenuous test. Perhaps this can be achieved by added an altitude parameter?

siimav: All right, keep the 10 minute requirement then. I would appreciate it if you removed those 2 comment-questions though.

smartdummies: Done

PR #1286, opened 2020-06-15
Move TD-339, Fix F-1B

Capkirk123:
Move the TD-339 from Early docking to Early landing, move F-1B to NF Orbital Rocketry


siimav:
This kinda conflicts with #1290 because both adjust the F-1 cost

Capkirk123:
Merge that first then I suppose, and I'll fix this later

PR #1290, opened 2020-06-22
More Price Fixes

Capkirk123:
Another pass of price fixes, focused mainly on the Apollo era (due to me getting caught in a rabbit hole of the F-1s cost). Changes include

- F-1: Reduced cost from 4555 to 2100. Apollo era documents show that F-1s cost little more than 2 million each (and even then, Rocketdyne was willfully overcharging to increase their own profits).

- F-1A: Reduced cost reduction from 800 to 260, corresponding with lower base F-1 price.

- F-1B: Reduced cost from 3300 to 1448, corresponding with lower base F-1 price.

- J-2: Reduced cost from 2000 to 1355, according to Apollo era documents.

- J-2S: Reduced cost reduction from 500 to 339, corresponding with lower base J-2 price.

- M-1: Reduced cost from 13600 to 9200, corresponding with lower base J-2 price.

- RL10A-1 through A-3-3: Reduced cost increase by approximately half. The RL10A-1 and A-3 were largely still prototypes, and didn't reach the goal performance of the RL10 until the A-3-3. Consequently, a large price difference between these versions doesn't make much sense. Later model RL10 prices are untouched

- LR87 and LR91 Kero: Reduced 1000+ costs down to the same as their hypergolic equivalents. A significant price increase doesn't make sense for simply converting an engine from hypergolic to kerolox (especially when that engine was originally kerolox in the first place)

- LR87LH2: Reduced modification costs slightly, corresponding with lower base J-2 price.

- NK9: Reduced cost from 450 to 350, to bring the cost/kN more in line with other Soviet engines.

- RD-0124: Reduced cost from 1200 to 500, according to Russian documents on its planned successor, the RD-0125.

- RD-0146 Reduced cost from 2000 to 1500. It is based off of the RL10A-4, and has similar performance.

- RD-180: Reduced cost from 2700 to 1240, according to ULA reports.

- RD-170: Reduced cost from 5400 to 2004, corresponding with lower RD-180 and RD-191 price.

- RD-191: Reduced cost from 1350 to 858, according to Russian documents.

- AJ10-133: Moved from 1963 to 1966, the same as the AJ10-137 Apollo SPS, since it was a competitor for the Apollo SPS.

- XLR81-LF2-SPS: Moved from 1963 to 1966, the same as the AJ10-137 Apollo SPS, since it was a competitor for the Apollo SPS.

- LR87LH2-Vac: Moved from 1967 to 1966, to be in the same node as the J-2. It was a competitor to the J-2, and Aerojet already had a working engine when the contract was announced.


Capkirk123:
Fixed conflicts

PR #1293, opened 2020-06-29
Tank Reworks

Capkirk123:
- Remove all "generic" tanks. Tanks that at least pretend to be based on a real part can stay, but stock, generic tanks are neither balanced nor necessary, and should be completely superseded anyway.

- Move Procedural Battery to Orphan node to prevent it from being used

- "Separate Structure" tanks have been renamed to "Stringer" tanks to avoid confusion among new players.

- Removed tech requirements for being able to put water in a tank.


Source/Tech Tree/Parts Browser/data/Procedural_Parts.json
```diff
@@ -48,6 +48,27 @@
             "NoResourceCostMult"
         ]
     },
+    {
+        "name": "proceduralBattery",
+        "title": "Procedural Battery (Deprecated)",
+        "description": "DO NOT USE THIS. Use Service Modules instead",
+        "mod": "Procedural Parts",
+        "cost": "999999",
```
siimav, line 56: Not sure that's wise. The combination of title, description and non-RP0 should already be enough.

Capkirk123: I suppose I don't need to add any extra motivation

PR #1296, opened 2020-07-01
Update SoundingDifficult.cfg

norcalplanner:
Updated sounding payload to reference HP tanks and the Tank UI.


norcalplanner:
My apologies for the pull request spam - I'm still new at this whole github thing.

PR #1297, opened 2020-07-01
Update SoundingBio.cfg

norcalplanner:
Updated sounding payload to reference HP tanks and the tank UI.


siimav:
Merged with #1296

PR #1298, opened 2020-07-01
Update DistanceSoundingIntermediate.cfg

norcalplanner:
Updated sounding payload to reference HP Tanks and the Tank UI.


siimav:
Merged with #1296

PR #1301, opened 2020-07-01
Fixed kOS upgrades adding mass

RCrockford:
Fixed upgrade process breaking kOS disk space and causing increased mass.
Tweaked early progression to be a bit more useful.


siimav:
Noticed some mixed tabs and space

RCrockford:
Spaces should be tabs now.

siimav:
Hmm, now GH is saying that the entire `UPGRADES` section was modified...

PR #1306, opened 2020-07-05
Move KIS mountable container to same node as its mount

lpgagnon:
The Container Box Mount is explicitly meant for attaching the
Container Box; the box shouldn't become available 2 nodes later

(I conveniently chose to make both available early rather than both late, but could go either way)


lpgagnon:
dunno what the f-1b cost change is about, but the update matches what's in the existing .json

PR #1315, opened 2020-07-21
Rebalance repeatable crewed contracts

leudaimon:
Reduces payouts along with preventing taking several contracts at once in
exploity ways. Also fixes a couple of mistakes/writing in these contracts.

~~This will be followed by a future PR improving the station rotation and
resupply contracts to make these worth it and more reasonable to trigger.~~
Those are being done in the same PR, with commits below


leudaimon:
I now added the LEO resupply mission. This contract changes the previous ones in several ways:

 * It is made only for LEO stations - a Lunar (and maybe Mars) version will
be added later.
 * Reward is based on a soft cooldown with a nominal time of 120 days, and
a more flat curve to raise above 100% much more slowly.
 * The requirement is now to bring enough supplies to keep the current crew
already in the station for 6 months.
 * I added a new vessel requirement to avoid exploits

leudaimon:
Rewrite of the crew rotation contract:
It changes the original contract by being more flexible: rotate at least half of the current crew in the station.
No checks on individual astronauts is made, but especially with Kerbalism the player will want to change the crew in the station anyway.

NathanKell:
So, uh, can this be revived? Or not worth it for next release?

leudaimon:
I kind of stopped midway, but would like to finish it. If you want to have a look and help me tracking what else needs to be added, I can try to crack it. My idea was to have a LEO series of contracts here, and then make Lunar (and maybe Mars) separate, instead of the generic ones currently implemented.

NathanKell:
Sure! Sorry, lost track on this. My feedback on the LEO stuff is that it is, at present (and presumably in this PR since I don't see changes about this) very US-centric. In particular a player using historic Soviet spacecraft simply cannot complete the 2-person contract until Soyuz (because it requires maneuvers), and at that point the 3-person contract pays better. Now, given that the deathtrap that was Voskhod only flew twice that is perhaps not a big deal, but it can be frustrating, because the 2+ contract is _clearly_ written in a Gemini-centric way.

I don't have an immediate suggestion here, though I think you're right it should be a progression. It might do to couple things to the records contracts, so the number of crew and the duration required in the contracts track what you've achieved so far (with maybe a bit extra duration to encourage you to go longer), and offering "with maneuvers" or "long duration" alternatives?

pap1723:
I still think it is alright. The Voskhod only flew to set records. It was most people in orbit and then to complete the EVA. Both of those are still possible. Yes, the maneuvering contracts are very Gemini-like, but isn't that because it is the only bridge to the Moon that we have reference for?

The repeatables should require maneuvering IMO and if you are going the Soviet route, you'll need to build a proper SM for your Voskhod. Just my 2 cents.

leudaimon:
Yeah, given so many people build their own service modules anyway, I don't think the limitation to maneuvers in the Voskhod is a hard one. 
I don't know how much it's worth dividing the contract into two though, with one for endurance and the other for maneuvers... might be more work than it's worth. 
How about stations? Is what is there an improvement on the current contracts? Is there any contract type missing?

pap1723:
Can this be combined with #1593 in anyway to create a MUCH better LEO experience?

NathanKell:
Closing due to Programs.

PR #1320, opened 2020-07-25
First attempt at packaging the parts browser into an executable (So f…

mattwrobel:
…olks don't have to have Python installed), this works on Windows for me.  Assuming it works for everyone, I'll add one for MacOS and hopefully linux as well.

How I did it:
  pip install pyinstaller
  pyinstaller app.py -F

Then the app.exe will be in the dist/ folder.


jwvanderbeck:
If someone can't figure out the very simple process of setting up Python and running this, do we really want them mucking about in it?

mattwrobel:
Some people run Python 2.7 and don't want to mess up their set ups with Python 3.  Also, you can't really do much damage with it as anything you change in it has to be PR'd and approved anyway no?  Mainly it's just for convenience and accessibility.

leudaimon:
I always understood it as a way for people without a programming background to be able to contribute. In this sense making it into an executable is a huge improvement

siimav:
Appears to work fine except for one mysterious issue. Apparently after regenerating configs from the standalone part browser, git is reporting that all lines in cfg files have whitespace changes. Oddly enough, after commiting the files all those whitespace changes disappear again.

lpgagnon:
'all lines have changes' is presumably because the standalone exe's choice of windows vs unix line endings was baked-in when the exe was generated, and doesn't match what your git clone used.

'disappear after committing' would be because the repo is configured with the obnoxious (but convenient in this case) `autocrlf=true` option

PR #1341, opened 2020-08-15
Add support for 7m heatshield

lpgagnon:
Arbitrarily placed in same node as 10m shield.

Goes with KSP-RO/RealismOverhaul#2347


DRVeyl:
Different part names were created.  Will close this and raise an issue.

PR #1346, opened 2020-08-18
RD-8, RD-120 tweaks

lpgagnon:
- make all RD-8 and RD-120 variants have the same settings (cost, tags,
  ECMs, etc)
- make the RD-8 cost 150 per vernier, instead of a crazy 1000
  (600 on the config, divided by engineTypeMult=0.25)

rd-8 cost isn't based on anything, pure balance. 4x 1000 to give
verniers to a rd-120 that costs 750 is just insane. 4x 150 is probably
still high (esp. with the turbopump tag); didn't want to accidentally
create a new king of small engines

Also removed the dead ROE-RD8-RE config block; that part name lived for all of 1 month

### DIFF
--- a/GameData/RP-0/Tree/ECM-Parts.cfg
+++ b/GameData/RP-0/Tree/ECM-Parts.cfg
@@ -142,6 +142,7 @@
     GuidanceEarly2m = 1000,         avionicsBoosterEarly
     GuidanceEarly3m = 1500,         avionicsBoosterEarly
     GuidanceStart1m = 1
+    HA3SLRD120 = RD-120
     HeatShield0 = heatshieldsLunar
     HeatShield1 = heatshieldsLunar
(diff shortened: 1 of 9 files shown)


siimav:
For the RD-8 and RD-120, why have an additional `entry_cost` if both parts already have an engine ECM? 
I'm also a bit confused by the RD-8 engine config cost shenanigans. Would probably be best if @Capkirk123 chimed in here.

lpgagnon:
Why have an entry cost at all: no idea, I just stuck to what existed. 

Cost shenanigans: I just wanted to set the part cost on the config so it wouldn't need to be duplicated for each part; thanks to engineTypeMult, this will even work if a 'all 4 verniers in 1 part' part is ever added (it won't have engineTypeMult, so it'll get the full cost).

The fact that it works suggests it's exactly why engineTypeMult exists.

PR #1348, opened 2020-08-21
French Engines

Capkirk123:
Add the Veronique and Viking engines in a balanced, but ahistorical, position.

Due to a lack of any equivalent engine, and generally awful performance, the Veronique unlocks in PW rocketry, with a new config unlocked each subsequent node.

The initial viking configs unlock in 1959, the same as the broadly similar but higher performance LR87. The UH25 configs unlock in 1961, and the Ariane 4 5C/6 configs unlock in 1965.


siimav:
Related to https://github.com/KSP-RO/ROEngines/pull/82

PR #1351, opened 2020-08-23
Make parts in development available in the editor

Standecco:
Uses the stock experimental part system to unlock all the parts in the currently researched and in-development nodes.
Moves the locked parts check to build time, and blocks you from building a vessel that contains locked or experimental parts.
If the vessel contains parts in nodes that have been unlocked, but haven't been bought yet, it will open a prompt asking the user if he wants to unlock those parts and build the vessel.
Adds a button to enable/disable experimental parts in the editor.


Standecco:
The Experimental parts toggle in the editor now allows you to buy researched-but-not-unlocked parts as usual, instead of completely hiding them. Enabling the option still makes the part available to build with.

PR #1363, opened 2020-09-05
Add disclaimers relating to Gemini scrubber

felix-u:


felix-u:
Does not work, will try to figure out why and hopefully reopen.

PR #1365, opened 2020-09-10
Sat contract tweaks

lpgagnon:
per discord conversation:
- Make Early Nav Sats get destroyed on completion
- Gate Early Comm Network behind Comm Test Sats


ec429:
LGTM :+1:

PR #1380, opened 2020-10-19
American Upper Stage Engine Costs

Capkirk123:
Adjust the cost of the Agena and AJ10 based on historical documents.

- XLR81-BA-7: Reduced cost to 200, according to historical documents
- XLR81-BA-11, Model 8096-39: Reduced cost to 129, according to historical documents
- All other Agena models have had their cost reduced to better match the new XLR81-BA-7 and Model 8096 costs

Sadly, no documents are available directly stating the cost of any AJ10 variant. However, by comparing the costs of Thor Able/Delta to Thor Agenas of the same era, approximate costs can be derived

- AJ10 Early: base cost reduced to 100
- AJ10 Mid: base cost reduced to 150
- AJ10 Advanced: base cost reduced to 200
- AJ10-137: cost reduced to 350
- AJ10-190: cost reduced to 300
- AJ10 Transtar: base cost reduced to 400

- R-4D: Base cost reduced to 35, according to historical documents


Source/Tech Tree/Parts Browser/data/Engine_Config.json
```diff
@@ -4980,7 +4980,7 @@
         "spacecraft": "",
         "engine_config": "Agena",
         "upgrade": false,
-        "entry_cost_mods": "Model117ECM",
+        "entry_cost_mods": "7000, Model117ECM",
```
siimav, line 4983: Shouldn't that number go into Model117ECM instead?

Capkirk123: Maybe? The end result is the same, bit I can change it

DRVeyl: Is this resolved?

Capkirk123: I think it is fine as it

PR #1385, opened 2020-10-29
Configure RO-RSRM, RSRM-1986

lpgagnon:
- Place and cost new RO-RSRM part
- Place RSRM's post-challenger config in the right node
- Move unlock costs to ECMs and part costs to config costs,
  to avoid having to duplicate them for every part

Needs https://github.com/KSP-RO/RealismOverhaul/pull/2372 for the new part


NathanKell:
Can you fix the merge conflicts? Would like to merge down.

lpgagnon:
push -f: rebased, conflicts resolved.

Note that the new part still depends on the RO PR, which I don't foresee revisiting myself (to validate non-restock behaviour). But there's still worthwhile bits here even without that

lpgagnon:
oops, one merge conflict marker slipped in. will fix

PR #1387, opened 2020-10-31
Fix for lunar sample return contract

StonesmileGit:
Potentially fixes an issue where another craft can complete parts of the contract requirements.

Limited testing has been done by me at this time


siimav:
If this works fine then the other sample return contracts would need the same change.

StonesmileGit:
Reason for the current structure found in 5f262638b11c7cd908cd67103bcb3ef371c909ba, current change re-introduces that issue.

StonesmileGit:
Since the proposed change doesn't work, I'm closing the PR until a method to fix the issue is found

PR #1399, opened 2020-11-14
Gate Mars Station behind Mars Crewed orbit

lpgagnon:
Mostly because getting offered Mars Station before moon
station doesn't make a lot of sense. A different prereq
may make even more sense, but this is at least more consistent.


lpgagnon:
dupe of #1376

PR #1404, opened 2020-11-21
Fix unlocking experimental parts creates a new vessel in edit mode

### LINKED ISSUE #1395: Unlocking experimental parts when Saving an Edit queues a new vessel
Lazy discord paste:
```
[8:29 PM]  lpg:I think I just hit a glitch with Experimental Parts: Edit a vessel, add an experimental part to it, hit Save. "unlock parts?" prompt is shown; click yes. Nothing seems to happen. fine, hit Save again
[8:30 PM]  lpg:back at the space center scene, look at KCT queue: there's a 0%-progress copy of the vessel ahead of the edited one
[8:30 PM]  lpg:I'm guessing "click yes" just enqueued a new copy, as it normally would when the prompt is shown in a non-edit session
[8:35 PM]  The Destroyer:Yup, had that happen as well
[8:56 PM]  Cattman423:I had that happen to me as well
November 11, 2020
[2:38 AM]  Standecco:mh, I could see that happening
[2:38 AM]  Standecco:open an issue or something so that I don't forget fixing it
```

ValiZockt:
Currently, when editing ships with KCT and the prompt shows to unlock all experimental parts it will create a new entry in the build list

Fixes #1395

### Changes
* When prompt shows to unlock experimental parts, only add vessel to build list if `EditorShipEditingMode` is false, else edit the current ship
* Move KCT *"Save Edit"* to a method, so it can be called from everywhere not only the IMGUI button
* Plural/Singular switching: Fund**s** (the same thing that already existed for part**s**)

### DIFF
--- a/Source/KerbalConstructionTime/Utilities/Utilities.cs
+++ b/Source/KerbalConstructionTime/Utilities/Utilities.cs
@@ -894,11 +894,13 @@ namespace KerbalConstructionTime
 
                     var unlockableParts = devParts.Keys.Where(p => ResearchAndDevelopment.GetTechnologyState(p.TechRequired) == RDTech.State.Available).ToList();
                     int n = unlockableParts.Count();
+                    int unlockCost = FindUnlockCost(unlockableParts);
+                    string mode = KCTGameStates.EditorShipEditingMode ? "Edit" : "Build";
                     if (unlockableParts.Any())
                     {
                         buttons = new DialogGUIButton[] {
                             new DialogGUIButton("Acknowledged", () => { }),
-                            new DialogGUIButton($"Unlock {n} part{(n > 1? "s":"")} for {FindUnlockCost(unlockableParts)} Funds and Build", () => { UnlockExperimentalParts(unlockableParts); AddVesselToBuildList(blv); })
+                            new DialogGUIButton($"Unlock {n} part{(n > 1? "s":"")} for {unlockCost} Fund{(unlockCost > 1? "s":"")} and {mode}", () => { UnlockExperimentalParts(unlockableParts); if (!KCTGameStates.EditorShipEditingMode) AddVesselToBuildList(blv); else EditShip(KCTGameStates.EditedVessel); })
                         };
                     }
                     else
@@ -918,7 +920,6 @@ namespace KerbalConstructionTime
                         HighLogic.UISkin);
                     return null;
                 }
-
 
                 double totalCost = blv.GetTotalCost();
                 double prevFunds = Funding.Instance.Funds;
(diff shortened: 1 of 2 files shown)


Source/KerbalConstructionTime/Utilities/Utilities.cs
```diff
@@ -894,11 +894,13 @@ public static BuildListVessel AddVesselToBuildList(BuildListVessel blv)
 
                     var unlockableParts = devParts.Keys.Where(p => ResearchAndDevelopment.GetTechnologyState(p.TechRequired) == RDTech.State.Available).ToList();
                     int n = unlockableParts.Count();
+                    int unlockCost = FindUnlockCost(unlockableParts);
+                    string mode = KCTGameStates.EditorShipEditingMode ? "Edit" : "Build";
```
siimav, line 898: I think "save edits" or "save changes" works better there

Source/KerbalConstructionTime/Utilities/Utilities.cs
```diff
@@ -894,11 +894,13 @@ public static BuildListVessel AddVesselToBuildList(BuildListVessel blv)
...
                     if (unlockableParts.Any())
                     {
                         buttons = new DialogGUIButton[] {
                             new DialogGUIButton("Acknowledged", () => { }),
-                            new DialogGUIButton($"Unlock {n} part{(n > 1? "s":"")} for {FindUnlockCost(unlockableParts)} Funds and Build", () => { UnlockExperimentalParts(unlockableParts); AddVesselToBuildList(blv); })
+                            new DialogGUIButton($"Unlock {n} part{(n > 1? "s":"")} for {unlockCost} Fund{(unlockCost > 1? "s":"")} and {mode}", () => { UnlockExperimentalParts(unlockableParts); if (!KCTGameStates.EditorShipEditingMode) AddVesselToBuildList(blv); else EditShip(KCTGameStates.EditedVessel); })
```
siimav, line 903: That delegate action is getting way too long. Should split it to multiple lines.

Source/KerbalConstructionTime/Utilities/Utilities.cs
```diff
@@ -959,6 +959,43 @@ public static BuildListVessel AddVesselToBuildList(BuildListVessel blv)
...
+            }
+
+            ship.RemoveFromBuildList();
+
+            double origBP = (ship.IsFinished ? -1 : ship.BuildPoints) + ship.IntegrationPoints;
+            double buildTime = KCTGameStates.EditorBuildTime + KCTGameStates.EditorIntegrationTime;
+            double difference = Math.Abs(buildTime - origBP);
+            double progress = ship.IsFinished ? origBP : ship.Progress;
+
+            newShip.Progress = Math.Max(0, progress - (1.1 * difference));
```
siimav, line 979: The same code appears to be duplicated on GUI_Editor as well. Should extract as a method.

siimav (changes requested):

Source/KerbalConstructionTime/Utilities/Utilities.cs
```diff
@@ -973,6 +1011,16 @@ public static BuildListVessel AddVesselToBuildList(BuildListVessel blv)
             return protoTechNodes;
         }
 
+        public static double GetShipProgress(BuildListVessel ship, out double origBP, out double buildTime, out double difference, out double progress)
```
siimav, line 1014: Hmm, that got a lot messier than I originally thought. Maybe return new BP amount, original completion % and new completion % ?

PR #1405, opened 2020-11-21
Place more NFElectrical reactors in the tree

lpgagnon:
still uncosted.
snap-10a to 1966, when the first one was flown
snap-50 and kilopower to near-future, because they never have

### DIFF
--- a/GameData/RP-0/Tree/TREE-Parts.cfg
+++ b/GameData/RP-0/Tree/TREE-Parts.cfg
@@ -10062,12 +10062,48 @@
     { name = ModuleTagNuclear }
 
 }
+@PART[RO-reactor-kilopower]:FOR[xxxRP0]
+{
+    %TechRequired = nuclearPowerNF
+    %cost = 0
+    %entryCost = 0
+    RP0conf = false
+    @description ^=:$: <b><color=green>From Near Future Electrical mod</color></b>
+
+    MODULE
+    { name = ModuleTagNuclear }
+
+}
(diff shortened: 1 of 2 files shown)


siimav:
What even happens when you do not set a cost value? Pretty sure that missing the cost on at least engines has caused issues in the past.

lpgagnon:
snap10a & snap50 end up each costing zero plus the cost of the Uranium (14.86 for snap10a's 0.3 EnrU, 3318 for snap50's 21 UraniumNitride)

kilopower ends up with a base cost of 4 instead of 0, no idea why (plus 215 for 3.9 EnrU).

In any case I haven't seen it cause any issues.

siimav:
So is there any specific reason why you would want to use an empty string for cost instead of explicitly setting it as 0? Like I said, if there's an MM patch that operates on the cost field, errors are going to happen.

lpgagnon:
no reason, that's just what it defaulted to

DRVeyl:
Should this be fixed here, noting that RP-0 tags "no cost" via <@PART[*]:HAS[#RP0conf[*alse],#RSSROConfig[*rue]]:FOR[zzzRP-0]>  ?  Or does something like leaving a cost field blank need to be fixed in the RP-0 parts browser tool to disallow empty fields?

lpgagnon:
push -f: rebased, resolved conflicts with kirk's additions, and set costs to "0"

NOT tested in-game, I'm away from ksp machine.

PR #1410, opened 2020-11-25
Fix hiring Hero kerbals after they would have retired.

### LINKED ISSUE #1409: [1.10.x] Firing original kerbals then re-hiring them after their retirement date causes them to immediately retire
I mentioned this in the Discord a couple of days ago. I took some time today to categorize it a bit better.

KSP 1.10 introduced a change which prevents you from losing any of your original four kerbals by firing them - they are now always available in the to-hire list so you can get them back.

When those kerbals are re-hired, they retain their original retirement date. If that date is in the past, they'll immediately be marked as retired. Confusingly, RP-1 will calculate a new retirement date and pop up the dialog to tell you this, but it doesn't seem to actually take effect.

I don't know what properties the game is actually using to preserve these kerbals in the list of applicants. From my zero KSP modding experience and a few minutes reading the API docs, setting `veteran` and/or `isHero` to `false` on those kerbals when the retirement date is calculated may be a fix?

This is pretty easy to reproduce - just fire your original astronauts then fast-forward a decade or so.

branan:
KSP 1.10 introduced a change which causes the hero kerbals to return to the applicant list when fired. If they are then re-hired after their retirement date, they immediately retire.

In order to prevent re-hiring once they've retired, we can force kerbals to become "crew" when their retirement date hits. This will properly move them out of the applicant pool and into the "lost" list.

Additionally, only update retirement date if  a kerbal is brand new, or past its retirement date (which would be a duplicate name). This prevents firing a hero kerbal and re-hiring it to push back the retirement date.

Fixes #1409

### DIFF
--- a/Source/Crew/CrewHandler.cs
+++ b/Source/Crew/CrewHandler.cs
@@ -692,6 +692,7 @@ namespace RP0.Crew
                             _toRemove.Add(kvp.Key);
                             _retirees.Add(kvp.Key);
                             pcm.rosterStatus = ProtoCrewMember.RosterStatus.Dead;
+                            pcm.type = ProtoCrewMember.KerbalType.Crew;
                         }
                     }
                 }
--- a/Source/Crew/CrewHandler.cs
+++ b/Source/Crew/CrewHandler.cs
@@ -580,17 +580,16 @@ namespace RP0.Crew
 
         private void OnCrewHired(ProtoCrewMember pcm, int idx)
         {
-            double currentTime = Planetarium.GetUniversalTime();
             double retireTime;
-            // Skip updating the retirement time if this is an existing kerbal who has not yet retired.
-            if (KerbalRetireTimes.ContainsKey(pcm.name) == false || KerbalRetireTimes[pcm.name] < currentTime)
-            {
-                retireTime = currentTime + GetServiceTime(pcm);
-                KerbalRetireTimes[pcm.name] = retireTime;
-            }
-            else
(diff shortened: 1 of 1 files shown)


Source/Crew/CrewHandler.cs
```diff
@@ -580,8 +580,18 @@ private bool TryGetBestSituationMatch(string body, string situation, string type
 
         private void OnCrewHired(ProtoCrewMember pcm, int idx)
         {
-            double retireTime = Planetarium.GetUniversalTime() + GetServiceTime(pcm);
-            KerbalRetireTimes[pcm.name] = retireTime;
+            double currentTime = Planetarium.GetUniversalTime();
+            double retireTime;
+            // Skip updating the retirement time if this is an existing kerbal who has not yet retired.
+            if (KerbalRetireTimes.ContainsKey(pcm.name) == false || KerbalRetireTimes[pcm.name] < currentTime)
```
branan, line 586: I wasn't sure if it was possible for a duplicate name to be generated - if not, I can remove that part of this check.

branan: I realized this is unnecessary and I'm an idiot.

The entire reason I found this issue is because the retirement check happens _before_ this gets called to update the retirement date. So any duplicate crew would immediately retire anyway. And in fact if they can be generated, they should now immediately retire and be moved to Lost thanks to the changes I made in the first commit.

I'll clean this up.

PR #1412, opened 2020-11-28
Modular cargo bay tooling

lpgagnon:
ROTanks is giving it a ModuleToolingROTanks, which no longer exists.

### DIFF
--- a/GameData/RP-0/Parts/RFProcTanks.cfg
+++ b/GameData/RP-0/Parts/RFProcTanks.cfg
@@ -184,11 +184,6 @@
 	}
 }
 
-@PART[ROT-*Tank*|ROT-ServiceModule]:FOR[RP-0]
-{
-	!MODULE[ModuleToolingROTanks] {}
-}
-
 @PART[RO-RFTank*|ROT-*Tank*|ROT-ServiceModule]:FOR[RP-0]
 {
 	MODULE
--- a/GameData/RP-0/ProcCosts.cfg
+++ b/GameData/RP-0/ProcCosts.cfg
@@ -178,6 +178,9 @@
 	{
 		%costPerkL = 0.5
 	}
+}
+@PART[proceduralStructural|proceduralNoseCone|ROT-ModularCargoBay]:FOR[RP-0]
+{
 	MODULE
 	{
 		name = ModuleToolingPTank


GameData/RP-0/Parts/RFProcTanks.cfg
```diff
@@ -184,11 +184,6 @@
 	}
 }
 
-@PART[ROT-*Tank*|ROT-ServiceModule]:FOR[RP-0]
-{
-	!MODULE[ModuleToolingROTanks] {}
-}
-
```
DRVeyl, line 191: Leave this in here, to deal with any synchronization issues with which version of RP-0 and ROTanks someone is on?  Can add a TODO note to remove after next release eg "remove after next release after 1 DEC 2020" or similar.

lpgagnon: We're already past the next release, ModuleToolingROTanks hasn't existed since April

DRVeyl: Of ROTanks?
BTW, I think "CargoBay" was a separate tooling type / category, to give a custom tooling instead of "stringer."  I have no personal opinion on how necessary that is to differentiate, but it was intentional.

lpgagnon: Of rp-0. Releases of ROTanks are irrelevant, I think; whether ModuleToolingROTanks is added and left alone, or is added and removed, or isn't added at all, observable behavior will be exactly the same

I saw the "CargoBay" type, but it wasn't clear to me whether that would have any implications for the rest of the settings. Sticking to a known-well-configured set of values seemed safer

DRVeyl: Yep, you're right, version of ROTanks won't matter.
As best I can recall, and tell from the one reference in my Discord conversation with DocRockwell when he was adding the part, setting the tooling type just defines a category.  Looks like there's a lot of tooling types that don't have much customization to them.  (ie, there's only tanks in <https://github.com/KSP-RO/RP-0/blob/master/GameData/RP-0/ToolingDefinitions.cfg>)

Maybe someone can remind me if/how many of the other tooling info gets set versus defaulted.

PR #1414, opened 2020-12-11
Add RZ config

Capkirk123:
Add configs for the Rolls Royce RZ series engines


seymonsters:
The engines use on Blue Streak or the proposed Hydrolox Engines (or both?)

seymonsters:
Considered adding in the RZ.10? Was very much a paper rocket engine but wouldn't be the first such inclusion?

The interesting challenge would be the Stentor engine with its 2 different thrust chambers etc...

Capkirk123:
I was previously unaware of the existence of the RZ.20. However, given that it was test fired, it does meet the criteria for being added to RO. I'll have to make up most of its performance though unfortunately, there seems to be very little data on it.

seymonsters:
Maybe base it on the RL10A-3 due to similar predicted thrust? And use a similar developmental pathway ( in the same way RZ.1 and RZ.2 followed the S3/S3D/LR79 route?

PR #1416, opened 2020-12-13
Add Gamma and Stentor Engines

Capkirk123:
Add Gamma engines and Stentor. Due to the slow pace of British rocketry, they have been placed in the tree to best match their performance with contemporaries, rather than according to their actual service date.


siimav:
Related to https://github.com/KSP-RO/RealismOverhaul/pull/2393

Issue #1370, opened 2020-09-30
Midgame balance changes


siimav:
### Lunar snowball nerf

- [x] Flybys - 1 completion only, possibly create as a milestone/record?
- [x] Impactors - keep the 3 but every completion after first having lower reward
- [x] Orbiters - keep the 2 but the second completion having lower reward and tighter orbit requirements
- [x] Landers - 4 completions with lowered rewards after first completion
- [ ]  Require a 30d experiment on third completion and a 90d experiment on fourth one 
- [x]  Add a separate contract for far-side landing with 2 completions

### Science changes
- [x] Significantly reduce the tech node cost of nodes between 1957 and 1961
- [ ] Significantly bump up the tech node cost of everything beyond 1965
- [ ] ~~Reduce the science multiplier for all bodies other than Earth~~
- [x] Reduce the landed science multiplier for Moon, Venus and possibly Mars
- [x] Reduce the number of unique biomes for Phobos and Deimos or eliminate biomes entirely
- [x] HOI-like system where the research rate of nodes is adjusted based on whether the player is ahead or behind the curve

### Repeatable commercial satellite contracts
- [x] Do *something* to prevent grinding them over and over breaking the economy.
Some ideas that have been thrown around:
  - add inflation factor to rewards such that the same amount of payload will pay less later on in the career. Would be based off tech progression, for example.
  - multiply the BP cost of commsat payload, such that just building the commsat takes longer, without increasing the rocket cost
  - [x] add soft cooldowns

### Other changes
- [x] Reduce the max completions on Orbital Return to 2 and lower the advances
- [x] Slightly increase the rewards on early sat contracts that have a single completion
- [ ] Add a contract for Lunar HS re-entry tests
- [ ] Reduce the rewards on atmospheric probe contracts
- [ ] Reduce the advance and/or deadline on first manned orbit
- [ ] Shorten the deadlines on manned Lunar milestones
- [x] Gate Crewed Lunar Landing by Crewed Flyby

norcalplanner:
Adjust maximum science point costs for different levels of R&D complex in conjunction with the post-1965 science cost increase

Sumguy62:
Consider making the science multipliers for high and low in space around phobos and deimos 0, since there are no stable orbits around those bodies irl or with principia.

ec429:
[...]

PR #1418, opened 2020-12-16
Lunar Snowball Nerf

StonesmileGit:
Implements the suggested changes in #1370, section 1. Lander revamp is to come.

Feedback about values or implementation is very welcome!

### DIFF
--- a/GameData/RP-0/Contracts/Moon Exploration/Lunar Flyby.cfg
+++ b/GameData/RP-0/Contracts/Moon Exploration/Lunar Flyby.cfg
@@ -4,7 +4,7 @@ CONTRACT_TYPE
 	title = Lunar Flyby (Uncrewed)
 	group = MoonExploration
 
-	description = Design and successfully launch a probe on a flyby of the Moon with a closest approach of no more than 5000 km, and successfully transmit or return some scientific data from the Moon's vicinity. You may complete this contract up to two times.&br;&br;<b>Number of Contracts Completed: $first_MoonFlybyUncrewed_Count</b>
+	description = Design and successfully launch a probe on a flyby of the Moon with a closest approach of no more than 5000 km, and successfully transmit or return some scientific data from the Moon's vicinity.
 	genericDescription = Flyby the Moon.
 
 	synopsis = Send a craft on a flyby of the Moon and receive data
@@ -18,7 +18,7 @@ CONTRACT_TYPE
 	autoAccept = false
 	minExpiry = 0
 	maxExpiry = 0
-	maxCompletions = 2
+	maxCompletions = 1
 	maxSimultaneous = 1
 	deadline = 730 * RP1DeadlineMult()  // 2 years
 
--- a/GameData/RP-0/Contracts/Moon Exploration/Lunar Flyby.cfg
+++ b/GameData/RP-0/Contracts/Moon Exploration/Lunar Flyby.cfg
@@ -49,9 +49,9 @@ CONTRACT_TYPE
 		type = Expression
 		
 		CONTRACT_COMPLETED_SUCCESS
-        {
-            first_MoonFlybyUncrewed_Count = $first_MoonFlybyUncrewed_Count + 1
-        }
+		{
+			first_MoonFlybyUncrewed_Count = $first_MoonFlybyUncrewed_Count + 1
+		}
 	}
 
 	// ************ PARAMETERS ************
--- a/GameData/RP-0/Contracts/Moon Exploration/Lunar Flyby.cfg
+++ b/GameData/RP-0/Contracts/Moon Exploration/Lunar Flyby.cfg
@@ -4,7 +4,7 @@ CONTRACT_TYPE
 	title = Lunar Flyby (Uncrewed)
 	group = MoonExploration
 
-	description = Design and successfully launch a probe on a flyby of the Moon with a closest approach of no more than 5000 km, and successfully transmit or return some scientific data from the Moon's vicinity.
+	description = Design and successfully launch a probe on a flyby of the Moon with a closest approach of no more than 5000 km, and successfully transmit or return some scientific data from the Moon's vicinity. &br;&br;<b>You may only complete this contract once</b> // Remove this note for experienced players at a later time!
 	genericDescription = Flyby the Moon.
 
 	synopsis = Send a craft on a flyby of the Moon and receive data
--- a/GameData/RP-0/Contracts/Moon Exploration/Lunar Flyby.cfg
+++ b/GameData/RP-0/Contracts/Moon Exploration/Lunar Flyby.cfg
@@ -43,17 +43,6 @@ CONTRACT_TYPE
 		contractType = first_OrbitScience
 	}
 	
-	BEHAVIOUR
-	{
-		name = IncrementTheCount
-		type = Expression
-		
-		CONTRACT_COMPLETED_SUCCESS
-		{
-			first_MoonFlybyUncrewed_Count = $first_MoonFlybyUncrewed_Count + 1
-		}
-	}
-
 	// ************ PARAMETERS ************
 	
 	PARAMETER
--- a/GameData/RP-0/Contracts/Moon Exploration/Lunar Landing Far Side.cfg
+++ b/GameData/RP-0/Contracts/Moon Exploration/Lunar Landing Far Side.cfg
@@ -5,7 +5,7 @@ CONTRACT_TYPE
 	group = MoonExploration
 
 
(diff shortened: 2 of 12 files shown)


StonesmileGit:
Sample return and rovers might also get some attention.

GameData/RP-0/Contracts/Moon Exploration/Lunar Rover.cfg
```diff
@@ -158,7 +172,7 @@ CONTRACT_TYPE
 			minDistance = 500.0
 			maxDistance = 2000.0
 		}
-	   RANDOM_WAYPOINT_NEAR
+		RANDOM_WAYPOINT_NEAR
```
siimav, line 175: Since you figured out how to show latitude and longitude in contract texts, maybe add that for this particular contract as well. This would allow playing the slot machine without having to visit the TS all the time to check the waypoint location.

GameData/RP-0/Contracts/Moon Exploration/Lunar Flyby.cfg
```diff
@@ -49,9 +49,9 @@ CONTRACT_TYPE
 		type = Expression
 		
 		CONTRACT_COMPLETED_SUCCESS
```
siimav, line 51: Remove that if there's only 1 completion

GameData/RP-0/Contracts/Moon Exploration/Lunar Landing Far Side.cfg
```diff
@@ -0,0 +1,156 @@
...
+	failureReputation = 25
+	failureFunds = @advanceFunds * 0.5
+
+	// ************ REQUIREMENTS ************
+
+	REQUIREMENT
+	{
+		name = CompleteContract
+		type = CompleteContract
+		contractType = first_MoonFlybyUncrewed
```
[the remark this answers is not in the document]
StonesmileGit: The idea is the same as for the normal landing contract; to fail this one if the crewed contract is accepted, however is this the correct implementation in both contracts?

GameData/RP-0/Contracts/Moon Exploration/Lunar Landing Far Side.cfg
```diff
@@ -0,0 +1,156 @@
...
+		type = CompleteContract
+		contractType = landingMoon
+		minCount = 2
+		title = Complete 'Lunar Landing' contract at least @minCount times
+	}
+	REQUIREMENT
+	{
+		name = CompleteContract
+		type = CompleteContract
+		contractType = first_MoonLandingCrewed
```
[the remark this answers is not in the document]
StonesmileGit: The idea with this is to fail the contract if the crewed landing is accepted (I think), however is this the correct implementation? (copied from the normal lander contract)

GameData/RP-0/Contracts/Moon Exploration/Lunar Landing Far Side.cfg
```diff
@@ -0,0 +1,156 @@
...
+		PARAMETER
+		{
+			name = VisitWaypoint
+			type = VisitWaypoint
+			index = 0
+			horizontalDistance = @targetBody.Radius() / 1000 * 1.57 // takes distance in km, close to pi/2
+			showMessages = true
+			disableOnStateChange = false
+		}
+		PARAMETER
```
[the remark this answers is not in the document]
StonesmileGit: This should be solved with the latest commit, but not sure

GameData/RP-0/Contracts/Moon Exploration/Lunar Rover.cfg
StonesmileGit: The implementation I added has the same question attached about the sign of the longitude as the GEO Weather sat PR

StonesmileGit:
The payout for the new 'Far Side Lunar Lander' contract might need to be updated, since they are copied from another contract currently.
However, I have no idea what reasonable numbers are and would like some input.

StonesmileGit:
@siimav Could you take a look when you have some time over?

GameData/RP-0/Contracts/Moon Exploration/Lunar Landing Far Side.cfg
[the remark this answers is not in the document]
siimav: But in this case the type should be AcceptContract (or something like that) instead

GameData/RP-0/Contracts/Moon Exploration/Lunar Rover.cfg
siimav: I think knowing the latitude could be useful as well in case you want to come in directly from the Earth-Moon plane

siimav:
A few of the REQUIREMENT indentations appear to be off by 1 level

GameData/RP-0/Contracts/Moon Exploration/Lunar Rover.cfg
StonesmileGit: The waypoint is forced equatorial, is this intended (no other rover contract is), or is it still useful to know that the waypoint is at 0° North?

GameData/RP-0/Contracts/Moon Exploration/Lunar Landing Far Side.cfg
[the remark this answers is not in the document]
StonesmileGit: Removed the REQUIREMENT with the latest commit. If it is needed to prevent this contract from being accepted while any of the crewed contracts are accepted, I can add that.

GameData/RP-0/Contracts/Moon Exploration/Lunar Rover.cfg
siimav: Oh huh. I never realized that it's always equatorial. I think adding 0°N to the description would still be useful to highlight that.

StonesmileGit: Fixed with latest commit

leudaimon:
I'm currently testing these (might take some time to make through them in-career to check the rewards and so on.
Regarding the flyby contract. I think it's necessary to add in the description that it has a single completion, especially because the others mention the number of completions.

GameData/RP-0/Contracts/Moon Exploration/Lunar Landing.cfg
```diff
@@ -28,10 +28,10 @@ CONTRACT_TYPE
 
 	// ************ REWARDS ************
 	prestige = Significant   // 1.25x
-	advanceFunds = 0.625 * 25000 * @RP0:globalHardContractMultiplier
+	advanceFunds = @/advance * @RP0:globalHardContractMultiplier
 	rewardScience = 0
 	rewardReputation = 50
-	rewardFunds = @advanceFunds
+	rewardFunds = @/reward
```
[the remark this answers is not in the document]
StonesmileGit: Good catch! Thanks for the feedback!

GameData/RP-0/Contracts/Moon Exploration/Lunar Orbiter.cfg
```diff
@@ -43,16 +43,50 @@ CONTRACT_TYPE
...
+		type = int
+		advance = @advances.ElementAt($first_MoonOrbitUncrewed_Count)
+		reward = @rewards.ElementAt($first_MoonOrbitUncrewed_Count)
+	}
+
+	DATA
+	{
+		type = List<float>
+		minPeList = [20000, 20000]
+		maxApList = [3000000, 2000000]
```
leudaimon, line 65: These two values for the Apoapsis are kind of a meaningless progression. The dv difference is negligible. I would suggest to go for 5,000 km in the first and 500 km in the second, at least.

StonesmileGit: @siimav Your thoughts on this would be nice

siimav: I agree with @leudaimon that 3Mm -> 2Mm makes almost no difference. 5Mm is incidentally what we had before Pap went in and made the contract harder. Setting the Ap requirement of second completion to <=500km would almost completely prevent unguided shenanigans here if that's what we want.

siimav:
I think the only thing preventing this from getting merged is the point @leudaimon brought up about the orbiter contract max Ap requirements. Do others feel that 500km for second completion is fair or should we try to strike a middle ground between that and what we have now?

lpgagnon:
500km doesn't seem unreasonable; but a quick look at historical missions suggests such a low Ap is fairly unusual. 

I wonder if requiring a _higher Pe_ might make sense? Min 20km isn't much of a constraint.

StonesmileGit:
If the Pe isn't going to be changed, then this should be mostly done; if something else should be changed, let me know

siimav (approved):

PR #1420, opened 2020-12-21
Add Waypoint Title to GEO Weather sat contract

StonesmileGit:
Adds a title to the Waypoint in 'GEO Weather sat' contract displaying target longitude.


StonesmileGit:
The waypoint title currently displays negative longitude for the western hemisphere, but this could be changed to show positive western longitude if prefered.

StonesmileGit:
Title now displays magnitude of the longitude, making western tarets display as a positive number

PR #1424, opened 2021-01-26
Add J47, J48 and VK-1 jet engines

Capkirk123:
Add J47, J48 and VK-1. Currently on hold until a decision can be made on how to deal with supersonic flight on early jets


siimav:
Related to https://github.com/KSP-RO/RealismOverhaul/pull/2403

PR #1426, opened 2021-01-26
 Add missing surveyor legs to Tech tree

TheBigLee:
The legs of Raidernick's Surveyvor probe are currently missing completely in the tech tree. Therefore can't ever be unlocked when doing a career run.
This commit will add the missing leg part to the tech tree.

I'm not really sure about the costs and copied the values from the surveyor retro part.
We can gladly discuss the costs of the part before merging this PR.
Furthermore I'm not sure if I used the correct module tag (`ModuleTagEngineLiquidPF`). I'll be happy to correct this if it's wrong.

### DIFF
--- a/GameData/RP-0/Tree/TREE-Parts.cfg
+++ b/GameData/RP-0/Tree/TREE-Parts.cfg
@@ -30698,6 +30698,18 @@
     { name = ModuleTagNoResourceCostMult }
 
 }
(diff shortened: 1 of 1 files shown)


lpgagnon:
This will have to be redone using https://github.com/KSP-RO/RP-0/wiki/Parts-Browser:-How-to-edit-part-configurations-for-RP1

siimav (changes requested):
Add to Part Browser json files

PR #1433, opened 2021-02-25
Add ROE RD-57

Capkirk123:
Add configs for the ROE-RD57, and fix the RD-57 and RD-56/KVD-1 configs.


siimav:
Are the RZ20 changes supposed to be in this PR?

Capkirk123:
Not really, but it's not an issue. I apparently didn't properly regenerate the cfgs after merging the RZ/RZ.20 PR, so when I generated the cfgs for this the parts browser shuffled them around. Nothing has actually been changed, just their positions in the .cfg files.

PR #1439, opened 2021-03-14
Fix First Rendezvous Contract

Teykn:
Previously, you could rendezvous with a spent stage which would satisfy the Rendezvous contract parameter. Fixed by adding a VesselParameterGroup that creates a specific vessel to rendezvous with, rather than rendezvousing with any object in space.

### DIFF
--- a/GameData/RP-0/Contracts/Milestones/Rendezvous.cfg
+++ b/GameData/RP-0/Contracts/Milestones/Rendezvous.cfg
@@ -115,6 +115,7 @@ CONTRACT_TYPE
 			type = Rendezvous
 			distance = 100
 			vessel = RendezvousTgt // To prevent unintentional rendezvous with other debris or "probes"
+			title = Rendezvous with the Target Vessel
 			hideChildren = true
 			disableOnStateChange = true
 		}
--- a/GameData/RP-0/Contracts/Milestones/Rendezvous.cfg
+++ b/GameData/RP-0/Contracts/Milestones/Rendezvous.cfg
@@ -5,7 +5,7 @@ CONTRACT_TYPE
 	group = Milestones
 	agent = Federation Aeronautique Internationale
 
-	description = The first successful rendezvous between two spacecraft was when Gemini 6 and 7 met in 1965. Wally Schirra maneuvered Gemini 6 to within 30cm of Gemini 7, close enough and precise enough to prove that docking two craft together in space was possible. Using your own knowledge of orbital mechanics bring two craft to within 100 meters of each other while in orbit around Earth. At least one of the spacecraft must be a new launch so make sure you have time to build and launch both the target and rendezvous craft before the deadline.
+	description = The first successful rendezvous between two spacecraft was when Gemini 6 and 7 met in 1965. Wally Schirra maneuvered Gemini 6 to within 30cm of Gemini 7, close enough and precise enough to prove that docking two craft together in space was possible. Using your own knowledge of orbital mechanics bring two craft to within 100 meters of each other while in orbit around Earth. Both of the spacecraft must be a new launch so make sure you have time to build and launch both the target and rendezvous craft before the deadline.
 
 	synopsis = Perform the First Rendezvous of two craft in space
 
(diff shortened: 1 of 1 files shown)


Arodoid:
Do you think it would be better for the contract to not lock you into a historical play through? I.E. the docking target should be allowed to have crew.

Teykn:
Hmm, I would think most people would have the docking target be historical. There's basically no incentive to have a crewed docking target, since you not only need to have more crew which require training, but you also need another crew-rated vessel, which takes significantly longer to build. I think most people would not mind the uncrewed requirement of the target vessel.

Arodoid:
I completely agree that there's basically no incentive. I just see a potential (very unlikely) issue if someone were to think "first docking" was actually "first docking with uncrewed target".

leudaimon:
1) I agree with Arodoid, no reason to require an uncrewed target
2) I'm not sure if the way you designed it fixes the issue, but I have a memory that just requiring a new craft on both doesn't solve the issue of a separated stage automatically completing the contract like it does now. Afaik KSP treats a separated stage as a new craft, no?

Teykn:
> 2. I'm not sure if the way you designed it fixes the issue, but I have a memory that just requiring a new craft on both doesn't solve the issue of a separated stage automatically completing the contract like it does now. Afaik KSP treats a separated stage as a new craft, no?

Ah, it's not just the new vessel that prevents it. I also have the vessel set in the Rendezvous Parameter so that the crew vessel **must** rendezvous with **that** specific vessel. That too, should prevent rendezvousing with debris. The vessel with the "core" part (for example, an avionics core or a crew capsule) is the one that CC locks onto. Anything that's decoupled off becomes a separate vessel, and thus, doesn't satisfy the requirement.

leudaimon:
Ok, my concern is that a separated upper stage with its own avionics would satisfy these conditions to be the target, wouldn't it?

Teykn:
> Ok, my concern is that a separated upper stage with its own avionics would satisfy these conditions to be the target, wouldn't it?

It would satisfy **most** conditions except for the Rendezvous parameter, since in it, it’s required that the capsule rendezvous with a specific vessel. CC locks onto said specific vessel as the vessel you want to rendezvous with, even if you switch to another vessel. Anything decoupled off becomes a separate vessel from the Rendezvous target. Once CC chooses a vessel, it will never change. 

tl;dr: wont satisfy Rendezvous Parameter cuz of “vessel = RendezvousTgt”. Debris/probes decoupled aren’t RendezvousTgt.

Hope that clears it up a bit.

StonesmileGit:
As the contract is currently, if you loose the target vessel, you are unable to complete the contract

StonesmileGit:
You can also still rendezvous with a vessel sent up on the same launch, you just have to switch to it to get it to register as the target vessel, and then switch back to the crewed vessel

lpgagnon:
comments in this conversation keeps referring to "docking"... bear in mind this _isn't_ the "first docking" contract

leudaimon:
I'm missing the understanding of how does CC define the target vessel. Because if it can't change at all, what happens if CC decides to define the target as some unrelated craft launched for other purpose?

lpgagnon:
Looks to me that this is going to make the contract broken in ways similar to some others: CC will latch on to the _first_ launched satellite after accepting the contract, regardless of whether the player intends that launch to have anything to do with it. With a 1 year deadline, it shouldn't be assumed that the next launch after accepting will be for the contract.

siimav:
There's still the option of splitting the contract into 2:
1) A contract for launching the rendezvous target (Agena). Upon launching a vessel into the desired orbit, it will complete the contract and use the CC `define` feature to store the vessel in a global variable. Available all the time and can be repeated to designate a new target vessel. No advance or reward.
2) A contract for launching the vessel that does the rendezvous (Gemini). Only accepts the vessel defined in contract 1 as the rendezvous target. Can accept once and has the advance + reward.

ec429:
Regarding crewed vs. uncrewed... wasn't the first _rendezvous_ between two Gemini capsules, because the Atlas-Agena launches kept blowing up?

pap1723:
@siimav My experience is that style of contract is bugged. I did it for my
Historical Progression Contracts that way and it didn't work well. However
that was many years ago so things may have improved.

Teykn:
> As the contract is currently, if you loose the target vessel, you are unable to complete the contract



> There's still the option of splitting the contract into 2:
> 
> 1. A contract for launching the rendezvous target (Agena). Upon launching a vessel into the desired orbit, it will complete the contract and use the CC `define` feature to store the vessel in a global variable. Available all the time and can be repeated to designate a new target vessel. No advance or reward.
> 2. A contract for launching the vessel that does the rendezvous (Gemini). Only accepts the vessel defined in contract 1 as the rendezvous target. Can accept once and has the advance + reward.

Might be possible to somehow add a behavior or requirement that cancels the contract if the Rendezvous contract is split into two. First contract could have a requirement that the Second contract is both unavailable and uncompleted in order to appear. Second contract has the behavior/requirement.

> Looks to me that this is going to make the contract broken in ways similar to some others: CC will latch on to the _first_ launched satellite after accepting the contract, regardless of whether the player intends that launch to have anything to do with it. With a 1 year deadline, it shouldn't be assumed that the next launch after accepting will be for the contract.

I am unsure how that would be fixed though. You CAN see what CC locked onto, but having it lock onto a specific vessel is slightly annoying. There’s not exactly a good specific part that we could require the Target vessel to have, unlike with docking.

Teykn:
Split the contract into two, added more requirements, gave the option for the Rendezvous target to be crewed, and made it so the 2nd contract automatically fails/becomes unavailable if the vessel disappears.

Hope this resolved more of the issues you guys had. Please list more if there are any that concern you.

PR #1441, opened 2021-03-15
Add the RD-0110 vernier to the tech tree.

leudaimon:
That part was not added to RP-1 yet, so I created a new part in the parts browser, copying from the stats of the vernier from RealEngines.

Also, apparently there were some changes to the json files that hadn't been
included in the configs yet, thus the other changes.


leudaimon:
The changes to the RL10 are dubious but seem legit. I repeated the config creation and they are still there

PR #1445, opened 2021-03-18
Add RSSSeaDragon and SpaceX Launch Towers compatibility

kyleander:


siimav:
In general all parts should be added through the Part Browser. The MLP pads are a bit of a special case where adding all the different parts invidivually doesn't make much sense.
https://github.com/KSP-RO/RP-0/wiki/Parts-Browser:-How-to-edit-part-configurations-for-RP1

siimav (changes requested):
Add to Part Browser json files

kyleander:
Will do now, didn't know of this tutorial until now!

DRVeyl:
Following up a year later, did you get a chance to do this and push the commit?

PR #1453, opened 2021-03-27
Move ALL AJ10-Mids to 1960

lpgagnon:
follow-up to https://github.com/KSP-RO/RP-0/pull/1116
which only handled the ROE- variants


DRVeyl:
Why did you clear the solids entry costs?

lpgagnon:
I didn't. _someone_ apparently forgot to regen .cfgs after updating .jsons.

Capkirk123:
Oops

lpgagnon:
That would be because _somebody_ didn't actually test his changes. Year was changed but not the actual node, which is the thing that matters. Will fix.

lpgagnon:
clearly everything is correct and no testing is needed:

Screen Shot 2021-03-30 at 09 59 37

lpgagnon:
fixed, and actually confirmed the SXT version shows up in the right node

PR #1454, opened 2021-03-28
More antennas

lpgagnon:
3 restockplus antennas and one stock that have RO and RA configs
but were left out of RP-1

Placement and costs set to match "equivalent" parts already configured


lpgagnon:
Worth noting that while these _have_ RO and RA configs, I haven't validated them. Gains look about right, but some of the masses look a little out of whack. Not RP-1's problem, though.

PR #1456, opened 2021-03-30
Add config for restockplus Viking

lpgagnon:
Part was configured in RO a long time ago, and apparently forgotten.

https://github.com/KSP-RO/RealismOverhaul/blob/fe5f66506ed60e27db78b62cec2894e1f95f5310/GameData/RealismOverhaul/RO_SuggestedMods/RestockPlus/RO_RestockPlus_Engine_Part_Config.cfg#L78-L83

Also fixed: the stock-engine-as-rd-0105 part got switched from using
liquidEngine3_v2 directly to using a clone of it recently, but
rp-0 wasn't updated to match.

https://github.com/KSP-RO/RealismOverhaul/blob/991ba498d8d15d09b886eb43e28b0e49223105aa/GameData/RealismOverhaul/RO_SuggestedMods/Squad/RO_Squad_Engines.cfg#L24

### DIFF
--- a/GameData/RP-0/Tree/ECM-Parts.cfg
+++ b/GameData/RP-0/Tree/ECM-Parts.cfg
@@ -426,6 +426,7 @@
     RO-NIMS = 2500,RO-UVSpectrometer
     RO-PMIRR = 2500,RO-TwoChannelIR
     RO-QuadropoleMassSpectrometer = 2500, RO-MIMS
+    RO-RD-0105 = RD-0105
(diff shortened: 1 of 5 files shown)


siimav:
But the other Viking engines appear to be in 1976 node instead of 1959

lpgagnon:
I took the lead of the ROEngines vikings and the engine configs, both set to 1959:

Screen Shot 2021-03-30 at 11 58 08

PR #1458, opened 2021-04-01
Minor experimental parts fixes

Standecco:
Makes it so that parts aren't marked as experimental unless in the editor, as that's the only time when you'd care about experimental parts.
Should also allow you to buy parts directly in the R&D screen, which was previously impossible.


siimav:
In general try not to include refactoring changes to random other methods or files.

PR #1462, opened 2021-04-08
SCANSat v20 Early Contract Support

pap1723:


pap1723:
Since the old ones are broken anyway...

PR #1467, opened 2021-04-13
Add biome scan to cameras

Capkirk123:
Since visual scan is disabled in RSS, add biome scan to cameras instead. Most biomes in KSP are relatively simple (lowlands, mountains, craters, etc.) and should be able to be distinguished even with relatively primitive cameras, given enough data.


jgalak:
This would fix the huge time/tech gap that currently exists between "LowResAltimetry" contracts and "Biome" contracts.  Please merge this in if possible.

PR #1470, opened 2021-04-22
Move Hydrazine to 1968

Capkirk123:
Based on https://ntrs.nasa.gov/citations/19680006875, move Hydrazine to 1968.

Although Hydrazine monopropellant thrusters were available before 1968, the lack of a spontaneous catalyst for starting meant they required starter fluid and/or electric heaters to start. For example, the main hydrazine engines used on Ranger and Mariner required both a catalyst bed and NTO starting fluid, limiting them to 2 ignitions. The development of the first iridium based catalysts, such as Shell 405, allowed reliable, repeatable starts for hydrazine thrusters. This let them be used for attitude control for the first time starting in 1968.


lpgagnon:
This is moving hydrazine rcs well past bipropellant rcs, how much sense does that make?

Capkirk123:
I cannot find any examples of Hydrazine based RCS before 1970. Everything I can find either used bipropellant, HTP or cold gas. Hydrazine was pretty rare as a monopropellant before 1970, only really used on limited start maneuvering engines.

pap1723:
In the real world, what is the advantage to Hydrazine over using bi-props?

Capkirk123:
I would assume simplicity. Half as many tanks, pipes, valves, etc. and the performance is much better than any other monopropellant available at the time, and it can be stored indefinitely (Voyagers hydrazine thrusters still work fine apparently). Also no worries of hard starts, assuming you have a good catalyst.

Hydrazine seems to have mostly replaced very small attitude control thrusters that were traditionally cold gas or HTP, larger stuff like kick motors and RCS on manned spacecraft has stayed bipropellant

lpgagnon:
Where does this leave #1241 ? As far as I can tell, generic thrusters are the only stand-ins we have for ranger/mariner-class engines; doesn't seem like a good idea to eliminate them without a replacement

Capkirk123:
Ideally this would be accompanied by adding the Ranger/Mariner main engine and some Soviet equivalent to fill that niche.

PR #1482, opened 2021-05-14
Add small radial sep motor to tree

jwvanderbeck:
See also: https://github.com/KSP-RO/RealismOverhaul/pull/2454


jwvanderbeck:
Not sure why CI failed but looking at the logs it appears to be a problem with the CI itself and not my PR?

PR #1483, opened 2021-05-16
Disable tooling for Proc Fairing/Interstage bases.

NathanKell:
You're already tooling the tank, and tooling the fairing sides/stringers, as we discussed an extra cost for tooling the base doesn't make much sense.

### DIFF
--- a/GameData/RP-0/ProcCosts.cfg
+++ b/GameData/RP-0/ProcCosts.cfg
@@ -6,20 +6,6 @@
 	{
 		@costPerTonne = 400
 	}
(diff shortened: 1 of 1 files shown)


jwvanderbeck:
I like this, but still think tooling on the fairing sides should be higher than it is now to better encourage using "off the shelf" sizes and creative payload restrictions.

NathanKell:
I am cool with doing that as well, yes--some of the old cost was from the
base, might as well shift onto the fairings (and *not* onto the stringers).

PR #1484, opened 2021-05-19
Move Gemini LEO heatshield to capsules node

lpgagnon:
Following the lead of ROCMercury: it's not possible to unlock the
capsule node without unlocking Human Rated EDL, which already
provides a suitably-sized shield. Don't force players to unlock
Early Landing just to use the "right" shield


siimav:
> Don't force players to unlock Early Landing just to use the "right" shield

I only see them having to unlock the HS node as a good thing

siimav:
IRL Gemini HS is a significant upgrade over Mercury. I know there isn't any difference ingame between the 2 right now but I would like to change it at some point.

NathanKell:
Uh, since when is there no difference ingame? Was there a big heatshield rebalance pass that I missed? Gemini's shield was tuned to support (very careful) lunar returns, which the LEO shields can't.

lpgagnon:
approximately since
1) RO grew a generic scheme for applying properties to heatshield, and
2) ROC Gemini CM got split from its shield, and separate LEO- and Lunar-rate shields were added

NathanKell:
That would explain, yeah. Thanks!

NathanKell:
Snice we're in the process of adding a Gemini level of shield, closing this.

PR #1486, opened 2021-05-22
Earth flying low science removal

leudaimon:
Needs the PR in Kerbalism to add crewed supersonic experiments!

Removes all other science from Earth Flying Low, while slightly increasing the science reward from flying high.
It also lowers the high atmosphere threshold to 30km and changes the first photography experiment to only work in space like its higher levels.


siimav:
Related ROKerbalism PR: https://github.com/KSP-RO/ROKerbalism/pull/99

pap1723:
Should we also lower the time requirements on the first photography experiment? Something like 5 minutes so the full science data can be gained from about 2 launches per biome?

leudaimon:
That would go in the Kerbalism PR, right?

NathanKell:
Per discussion, sounds like we want 40km (maaaybe 35km?) as flying high threshold.

marsh1832:
I recently did a quick SR playthrough to test some things, included was removing the flying low science similar to this, and I admit that it felt a lot better, progression wise. A thought, is there any intent/ability to remove the science reward for the first recovery of a vehicle after flight? Similar to the other science, this is not revolutionary, and the reward should not start until at least flying high.

leudaimon:
The recovery science is a stock feature with very little room for customization (one single multiplier per body, that also affects contract rewards for that body)... I don't think we can remove it from the Earth for flying low only.

siimav:
Except maybe mark it as already collected with Configurable Start, @Standecco ?

Standecco:
I've never looked at it, but so far I haven't been able to complete experiments either. It may be harder than anticipated

leudaimon:
Messed up the branch.. I'll remake the PR in a moment

PR #1490, opened 2021-05-26
Balance the 1955 basic solids tech node

marsh1832:
As it stands the basic solids tech node is kind of worthless, too expensive for non-caveman careers, and cavemen don't use it regardless. These changes aim to add value to the non-caveman career.

-The 18KS7800 is moved from 1956 to 1955. This allows it to be used in it's historical purpose, as a 3rd stage for an aerobee rocket. The performance is inferior to other 1956 solids so there's no reason for it to be there other than not having an ECM.

-The Nike has it's ECMs removed and it's unlock cost adjusted to 2000, twice that of the 18KS7800. The ECM makes it far too expensive considering it's limited usefulness. It's cost is reduced from 360 to 100, putting it between the 18KS7800 and a castor. 

The above changes allow for career use of an aerobee 300/350. I'm not sure if these changes would affect the speedrun potential of the game. In my testing the Nike is pretty poor as a solid for an orbital LV.


leudaimon:
It definitely is necessary to make solids more worthwhile. The removal of the ECMs from the Nike is not very well founded though... we need these ECMs to make a proper progression and capture the cost of investment in solids. I think it's better to not change the Nike from its current status, and just keep the change to the 18KS7800. Also, we are planning to change the materials science node requirement for these nodes in the solids, to make them easier to get earlier.

marsh1832:
The ECMs for the Nike are far too expensive. They're balanced according to the kinds of funds you would put into orbital technology, and it's made worse by the fact that aerobee rockets are very lean in their cost, making it an even worse deal compared to DR rockets. 35k is a lot of funds to put into something that will only gain you a few hundred m/s at best. Having a build time greater than a castor is obviously wrong... Sacrificing the practical application of the Nike just for "progression" seems very strange to me. Doesn't research time also count as progression? Why don't earlier solids have any ECMs or progression? To me it seems like a pass was made to add progression to solids, but the Nike was pulled into the "expensive orbital era" when it should have been in the "cheap SR era."

NathanKell:
I can't find reference to the Nike using a fiberglass casing, and certainly its PMF implies it did not. As for it being a hollow-casing solid, I haven't been able to find out either way, still looking (and asking around).

lpgagnon:
> we need these ECMs to make a proper progression and capture the cost of investment in solids

It seems to me the main effect of the current setup is that nobody ever unlocks Nike until ready to also unlock 1956 solids. Nike is just not worth investing into on its own

NathanKell:
So, actually a bunch of solid entry costs are busted. GCRC certainly wasn't fiberglass, that was what made Altair special.  Closing this in favor of a larger rework here (including probably lowering the cost of solidsHollow a bunch and applying it earlier).

lpgagnon:
Entry costs were only one element of this PR; what about the rest? (moving 18KS7800 one node, Nike part cost)

NathanKell:
Yes, that stuff too we need to deal with. It's just that it's ballooning enough that I wanted to make an issue to track it. Add whatever you find that's off to the issue please.

PR #1491, opened 2021-05-27
Repackaged Parts Browser app as exe file

mattwrobel:
How to do this:

Uses pyinstaller to create the exe from the application: https://www.pyinstaller.org/

On WIndows machine that has python 3:  pip install pyinstaller
cd Source/Tech Tree/Parts Browser
pyinstaller --onefile app.py
mv dist/app.exe ./


Capkirk123:
I cannot get these steps to work. I followed the instructions as best I could, but when creating a new .exe, I get the error `Traceback (most recent call last):
  File "Parts Browser\app.py", line 5, in <module>
ModuleNotFoundError: No module named 'part_data'
[18572] Failed to execute script 'app' due to unhandled exception!`

NathanKell:
Updated with what is now required: `pyinstaller --onefile --path=. --add-data "templates;templates" --add-data "static;static" --distpath . app.py`
Path has to be extended to current dir to find the local modules; add-data needs to be used for flask to find the html etc data.

PR #1492, opened 2021-05-29
Don't allow stacking of avionics units' masses

NathanKell:
Now that we have procedural avionics, use only biggest avionics -- don't allow multiple avionics units to stack.


marsh1832:
Uh oh. Gamers are going to riot. Is it possible to unlock multi avionics with a certain tech level?

NathanKell:
I'm confused. This isn't about "you can only have 1 avionics part per craft". This is about "you can't get 100t from your booster and 20t from your upper and 5t from your payload and control a 125t stack" -- you need 125t of avionics somewhere. E.g. when Saturn V was launching to orbit, the CSM wasn't helping control it, the IU on the S-IVB was doing that alone.

marsh1832:
That's what I had thought you meant, and it makes sense. I must have misheard on one of your streams mentioning that splitting of the avionics between stages was something that happened at some point, but not on the early rockets.

NathanKell:
Oh! Yeah, other way around. Early on each stage had its own guidance (which in our terms means booster had guidance for the whole stack, upper had guidance for itself + payload, etc.). Due to how "guidance" meant "tuned for one set of rates". It took until the Saturn IU, basically, for there to be modern (i.e. generic) guidance--the IU could control the whole stack from launch through TLI, and would support addition of boosters / drop engines / whatever.

marsh1832:
Ohhhh okay I see what you're saying. It started with each stage having to have avionics to control the entire remaining mass of the rocket, but with Saturn, it changed so a single avionics unit could control the entire rocket. This is a cool update! And it makes building a rocket a little easier... no more forgetting to check the total avionics control... now just have to make sure there's a core somewhere that can control the entire mass of the rocket... whether it's on the top or bottom depends on how you're roleplaying.

NathanKell:
Yup! And hopefully Soon (tm) we'll have better stage detection and actually be able to make that tech distinction matter ingame. :)

siimav:
We probably also want to mention how the controllable mass works somewhere in the avionics window.

NathanKell:
Note to self: Fix the UI too

PR #1494, opened 2021-06-01
Rework the FirstSolarSat contract parameters

siimav:
Idea is for the contract to have meaningful design and mission considerations rather than have the player slap the smallest solar panel they can find onto the probe.


siimav:
@jwvanderbeck said that the "Have positive energy balance" parameter doesn't complete when it should 😕

siimav:
I remember someone mentioning that it completed correctly for them. Needs more testing, I guess?

PR #1497, opened 2021-06-04
Clarify that avionics is needed for first 50s

jwvanderbeck:


RCrockford:
This should probably include the 3km downrange milestone contract as well.

leudaimon:
Looks good! I somehow thought Sounding Bio and Sounding Film also required avionics, but it doesn't look like, so this should cover all the contracts with this requirement already.

lpgagnon:
bioSR goes straight up, needing avionics would be odd. But film goes downrange and would make sense to need it. I have a change to do it in my install, but it would probably deserve a rewards tweak, which I'm not interested in doing

jwvanderbeck:
> Looks good! I somehow thought Sounding Bio and Sounding Film also required avionics, but it doesn't look like, so this should cover all the contracts with this requirement already.

Haha yeah I had thought the same but when I looked they didn't appear to :)

PR #1499, opened 2021-06-04
Change science at Earth to remove flying low

leudaimon:
Removes the flying low multiplier, adds a situation mask at flying high for bio and photo experiments, and lowers flying high to 40km. This depends also on a PR to ROKerbalism https://github.com/KSP-RO/ROKerbalism/pull/99 that adds a crewed supersonic series of experiments

### DIFF
--- a/GameData/RP-0/Science/BodyScienceParams.cfg
+++ b/GameData/RP-0/Science/BodyScienceParams.cfg
@@ -12,8 +12,8 @@
 				
 				%landedDataValue = 0.00001
 				%splashedDataValue = 0.00001
-				%flyingLowDataValue = 0.25
-				%flyingHighDataValue = 0.5
+				%flyingLowDataValue = 0.001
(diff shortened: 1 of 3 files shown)


Bornholio:
Have to disagree with this.  seems like .noplanes hate and compared to one bio canister to space it does not affect balance that much.  with combined VAB SPH build rates low altitude planes are a big drain on a program in terms of speed.

Elouda:
Its just tedium removal, and makes launchsite choices less unbalanced in terms of biomes.

There is still plenty of reason to do planes to get the supersonic and near Mach 2, High Altitude and Hypersonic science which more than compensates imho, and is not launchsite specific.

Bornholio:
this is also NASA  [List_of_NASA_aircraft](https://en.wikipedia.org/wiki/List_of_NASA_aircraft) 
why take it away,  nothing is forcing this play style , don't have to play tedium by removing others fun

PR #1501, opened 2021-06-05
Restrict Far Side Lunar Lander angle to 80 deg

StonesmileGit:
Added notes in contract parameter about landing within 80 deg, as well as contract description, with reference to Lunar Libration


StonesmileGit:
Added change to how max completions are displayed, current contract description:
image

Issue #1495, opened 2021-06-02
Solid entrycosts are busted


NathanKell:
From a quick check: Nike is not fiber, nor is GCRC; that was what made Altair special and such a high performance revelation. Also other fiberglass case motors should get the solidsFiber ECM. And basically everything past the Tim should require solidsHollow (except sep motors I guess?).

That should mean solidsHollow becomes cheaper and solidsFiber more expensive. Also it's super weird that the Castor (XM-20) is cheap to unlock but the T17-E2 requires it but _then_ adds another 10k funds on top. So there's a bunch of stuff that needs a sanity check.

<EDIT> removed ref to NGNC, Kirk set me straight>

lpgagnon:
In addition to entry costs:
- Nike part cost is 360, vs 100 for Castor 1 (XM-20). 3x 2.5ks18000 gets roughly the same performance as 1 Nike for ~1/3 the cost, and unlocks 1 node earlier. Even if Nike's entry cost comes down, it's still a trap to use it
- 18KS7800 is weirdly placed in the 1956 node; it's a sounding rocket upper stage that unlocks right when the sounding rocket era ends

(from #1490, originally reported by Marsh)

marsh1832:
I agree about the 18KS7800, it comes too late to be of any use to sounding rockets, and it’s inferior to the rest of the solids in 1956.

The Nike has two problems, it’s ECMs/unlock cost, and the build cost. It’s vastly inferior to a castor, and beyond a small boost off the pad it can’t really be used for an LV, which means it’s only application is a sounding rocket. Critically, it needs to be priced appropriately for an aerobee rocket.

lpgagnon:
This one may be fine, but since we're here, maybe worth taking a fresh look at how justified this cost imbalance is:
1 GCRC: 170 funds
A full stack of Juno 1-style Baby Sergeants, T17-E2: 15*60=900 funds (plus decouplers)
Upgraded to JPL-532A: 1350 funds

(Castor's 150 funds is starting to seem like the outlier)

lpgagnon:
Another thing worth looking at is how Fuel cost factors into part cost. For the early solids, fuel is a small fraction of the cost (1/170 for gcrc, 19/150 for castor), but some of the later ones are just broken. AJ-260 costs -16000 funds when empty, UA1204 -800, etc (both PBAN, maybe a clue)

NathanKell:
[...]

PR #1503, opened 2021-06-07
Rebalance the solid motors (US mostly), esp. early

NathanKell:
Re: #1495 

The entry cost modifiers for many (most early) solids were very wrong. In addition the costs were often incoherent. These changes seek to make solids sensible, the ECMs consistent, etc.

High-level changes
* Pretty much everything except sep motors requires Hollow. Hollow SRMs were a WWII invention and come stock on just about everything. For that reason, this is pretty cheap.
* Fiber was a new advance for the Altair (and later ABL motors).
* Both HTPB and PBAN were post-aluminumizing. Add a shared Aluminum ECM.
* Fix up some specific issues with the Antares, and with basically all STAR motors sharing a single ECM (???)
* Fix out-of-whack Baby Sergeant costs and entrycosts. Make GCRC slightly cheaper and Altair slightly more expensive. Make Nike competitive.
* Make 56 solids no longer depend on Satellite materials science.
* Move some solids back to Basic from 56.


NathanKell:
- [ ] Add Polyurethane binder; HTPB is modern. Use that for Polaris and other early Aerojet solids.
- [ ] Remove HTPB from all uses before 1969 (earliest citation I can find for its use) - @Capkirk123 do you know of an earlier case?

NathanKell:
This depends on https://github.com/KSP-RO/RealismOverhaul/pull/2473

PR #1506, opened 2021-06-14
Spaceplane passenger cabins should require training

nova-ad-astra:
They used to default to 0 training time, which is not the desired behavior. The training times will now instead equal the equivalent capsule from the same tech node with the same reentry rating and bulkhead profile.

### DIFF
--- a/GameData/RP-0/CrewTrainingTimes.cfg
+++ b/GameData/RP-0/CrewTrainingTimes.cfg
@@ -73,7 +73,7 @@ TRAININGTIMES
 	ProtoSpaceplane = 100, EVA, Docking // a bit more than SecondGenCapsule because it needs actual piloting
 		RO-Mk1Cockpit = ProtoSpaceplane
 		RO-Mk1CockpitInline = ProtoSpaceplane
-		RO-MK1CrewModule = ProtoSpaceplane
+		RO-Mk1CrewModule = ProtoSpaceplane
 		MK1CrewCabin = ProtoSpaceplane
 	ProtoSpaceplane-Mission = 120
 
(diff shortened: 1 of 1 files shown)


ec429:
Idk, isn't half the point of spaceplanes that passengers can ride in them with less vehicle-specific training?  The passengers don't need to be trained to pilot the thing...

I think they should have much lower proficiency training times than cockpits/capsules, but the same mission training time.

nova-ad-astra:
@ec429 Well in real life, certainly. This could possibly be modeled, but from a gameplay point of view I believe players are more likely to train the entire crew in batch (I certainly do). If you split the training times, you either have to start training at two separate times to sync up the mission training expiry for your pilots/passengers (fiddly) or you'd run the risk of the passengers losing their mission readiness before your pilots, potentially making you start the training process all over again (frustrating).

Keeping it balanced means you have to initiate mission training once, the crew will all be ready to go at the same time and everything just works. You lose out on certain edge cases such as being able to hire and train extra crew after starting your pilot training in case you redecide on your desired crew count, but other than that this seems like the least painful solution.

ec429:
That's why I said the _mission_ training would match the cockpit.  It's only the _proficiency_ training that would be shorter for the cabin.

lpgagnon:
2 things to keep in mind (sort of repeating things from discord)
- There's not really a use case for "passengers" in rp-1; seems like the concept would need to be better-defined before any effort at fine-tuning training times. 'same as capsules' seems like a better, less-abusable default until that point
- For those who do have a "legitimate" use case for passengers, there's plenty of options for getting around training times (e.g. training in the ROEngines Launch Tower), so this change won't actually hard-block anything

lpgagnon:
... never mind https://github.com/KSP-RO/RP-0/commit/4850b31675a56af11cb6639d5c8d1f926bad2d96

NathanKell:
Yeah, I tend to think the best option here is very quick (a week?) or nonexistent proficiency training, and reasonable mission training.

DRVeyl:
Did 1636 actually implement this, and can it be closed?

siimav:
No, it did not

PR #1511, opened 2021-06-18
Let space station contracts select the same station

lpgagnon:
Because the "randomness" we get from SelectUnique is very poor
(took ~50 Declines to get NewCrew instead of PowerModule), and
because there doesn't seem to be much harm in allowing them to
be combined


lpgagnon:
This has been sitting in my install for a long time; had been sitting on it because someone had started working on a real overhaul of station contracts. No sense in keeping it to myself, though

PR #1518, opened 2021-06-29
Add missing config to ROE-RangerRetro

Capkirk123:
Add missing config to ROE-RangerRetro


NathanKell:
gonna ship this whenever we next ship, so merging down.

PR #1520, opened 2021-06-30
(text) Change text of the suborbital return contract

adouge:
Added three commas in the suborbital return contract description. Certain about the last one ("Although ..., ... "), but not sure about the first and second ones ("As early as 1947, ...", "... trajectories, but ..."). I think it doesn't hurt, though.


marsh1832:
I don't think the first comma is required. There's no need for a pause there, although the phrasing on the first sentence is a little odd. Might I recommend:  "As early as 1947 the U.S. was launching fruit flies into space on suborbital trajectories and recovering them, but it wasn't until 1951 that a monkey named Yorick and eleven mice were successfully recovered from an Aerobee flight.

adouge:
Yeah, that sounds good, thanks

PR #1521, opened 2021-07-01
When simming with TF enabled, set data rate to 0

jwvanderbeck:


jwvanderbeck:
do NOT under any circumstances merge this PR... I posted it only to get feedback

NathanKell:
Feedback: I think it is useful to have data enabled in a sim, so you can see how fast you gain data.

PR #1522, opened 2021-07-03
Give soft cooldowns to 'Experimental rocketplanes' and make it unavailable after either XPH is completed

siimav:
Being able to complete all 4 in a quick succession with basically the same plane doesn't make much sense. Same goes for coming back to this contract after having done an XPH or 2. Now the player has a choice of waiting and completing all 4 or jumping to XPH after 2 have been completed.


marsh1832:
Awesome. I was considering adding a soft cool down here. It was always kind of annoying to have to do 4 of these flights in a row after building an X1.

PR #1527, opened 2021-07-06
Add CI to RP-0

jwvanderbeck:
How to create releases in a repository once Github CI is enabled
Create the new release in Github as you normally would, with the following guidelines

1. The tag **must** be in the format `v[Major].[Minor].[Path].[Build]`.  All four elements are required.  Use 0 if you don't need them.  All elements **must** be integers. EG: `v1.10.0.0`
2. Be sure to point the release to the appropriate branch as desired
3. Title and Description as normal
4. Mark the release as **pre-release** initially.  This is so that CKAN won't pick it up until the release is finished building
5. Publish the release

After publishing, the CI should kick in if you did everything properly.  Once finished, it will automatically attach a ZIP file to the release, EG RP-1-v1.10.0.0.zip

Once that zip appears, you can go into an edit the release to uncheck pre-release and let CKAN pick it up

### DIFF
--- a/.github/workflows/attachReleaseArtifacts.yml
+++ b/.github/workflows/attachReleaseArtifacts.yml
@@ -33,7 +33,7 @@ jobs:
                     
       - name: Build mod solution
         run: |
-          rm -f -r ${GITHUB_WORKSPACE}/GameData/RP-0/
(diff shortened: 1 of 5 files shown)


jwvanderbeck:
Tested in my fork but I: would suggest doing a test release in this repo after merging just to be sure everything translates over correctly. (Testing it before the merge won't work due to the way CI works)

PR #1533, opened 2021-07-22
Update LoadingScreenTips.txt

norcalplanner:
Added more loading screen tips to reflect recent changes


norcalplanner:
I have no idea why this says that it failed checks - I used the same procedure that I did previously.

NathanKell:
@jwvanderbeck it's complaining the password for aws is incorrect?

NathanKell:
@norcalplanner to be clear you're fine, it's not you.

jwvanderbeck:
It is because it is a fork. Any PR from a fork will fail to build due to Github security restrictions that prevent "secrets" from being used.  There is no way around that.

I've said this before but the only way to handle these is:
1. Encourage core developers with the permissions to use branches inside the repo for work and not forks
2. For external collaborators using forks, if it must be verified, merge the fork into an internal branch for testing once the collaborator indicates they are finished.

This is not ideal, but there is simply no way around it unless you want the KSP assemblies out in the open and I for one don't feel like getting sued :)

norcalplanner:
So is there anything I can do to help move this forward at this point?

NathanKell:
Nope, I just didn't get to this last night. :)

PR #1560, opened 2021-09-19
Adjust research rate per-node based on how much the node is ahead or behind the current date

siimav:
Implement the following from #1370:
*  HOI-like system where the research rate of nodes is adjusted based on whether the player is ahead or behind the curve

Currently set the curves as following (negative year values are ahead and positive behind):
* -15 -> 0.1x
* -5 -> 0.5x
* [anything between -1 and 0] -> 1x
* 5 -> 1.8x
* 15 -> 2.5x


MikeOnTea:
I'd probably prefer a smooth curve but apart from that i like the concept, nice! :)

Issue #1481, opened 2021-05-10
Make link clearer between training item and parts


NathanKell:
Discord user dunbaratu mentions that there can be a disjoint between what a part is called (1.25m cockpit) and the training for it (X-15). We should look into ways of standardizing and/or making the linkage clearer.

Dunbaratu:
One way to present it that would have helped me would be on the RP-1 window under that "Courses" tab when it lists a lot of very wide horizontal buttons with the training program names.  There's a lot of unused horizontal space in those button labels that could be used, since they're stretched the full width of the window.  Something like this as the button label might work:

"Proficiency: Mercury. (For Mercury, Mk1 Pod,... ,... ,..)"

siimav:
With #1545 we could add tooltips that list all the parts associated with that training course.

PR #1562, opened 2021-10-04
Add tooltips with part titles to courses

siimav:
Caveats;
* Only parts that have been unlocked or are under research, are added to tooltips
* Needed to create courses for parts that are researched but not yet purchased. This includes mission courses.

img


siimav:
Closes #1481

PR #1570, opened 2021-10-17
Tweak space center visuals

lpgagnon:
- Since we can't normally upgrade the SPH, make it default to its
  fully-upgraded appearance like the runway. A fully-upgraded SPH in 1951
  looks less silly than a lvl1 SPH when everything else is fully upgraded

- Drop admin building upgrade cost, and tie visual level to upgrade
  level. So player can have the lvl1 shack in 1951, and can upgrade it
  to better fit the rest of the buildings later

n.b. This does mean that the SPH will look upgraded from the start even if the Common Build Line option is turned off and upgrades _are_ allowed. Seems a better tradeoff than the current state.

My preference would actually be to give the SPH token upgrade costs (and [allow upgrading the SPH](https://github.com/KSP-RO/RP-0/blob/0d6f93ab353f981df83124ed67dda10e4e3f9583/Source/KerbalConstructionTime/KSCContextMenuOverrider.cs#L57-L58)) like admin building; the SPH interior seems to be stuck on lvl1 even with this change. But that would really break separate-lines mode and would need separate presets with different costs.


siimav:
With #1573 only the admin building part is relevant now?

GameData/RP-0/CustomBarnKit.cfg
```diff
@@ -90,11 +90,11 @@
...
 	@ADMINISTRATION
 	{
-		@levels = 2
-		@upgradesVisual = 3, 3
-		@upgrades = 38000, 38000
-		@activeStrategyLimit = 1, 1
-		@strategyCommitRange = 0, 0
+		@levels = 3
+		@upgradesVisual = 1, 2, 3
+		@upgrades = 1, 1 // purely visual, so basically free
```
siimav, line 95: Pretty sure this should have 3 values. The cost of 1st level is still used for determining facility repairs.

PR #1574, opened 2021-10-20
Change the ScanSat contracts to require technology instead of scanner part

TwistedGiraffe:
Two major consequences: 
1) Now low-resolution contracts will show up only after finishing research human spaceflight science, whereas previously the contract is available once the tech is in the queue. 
2) Since neither SCANsat_Scanner24 nor SCANsat_Scanner2 has RP-1 configs, this pull request will actually allow biome and high-resolution scan contracts to be accepted.


lpgagnon:
PR seems to work as intended: I have access to contracts for LowResAlt, HiResAlt and Biome contracts, with required nodes unlocked. Looking at a 1951 save, the 'not met' criteria show the right tech names

PR #1579, opened 2021-11-02
Add kOS disk space to upgrade part description

Amselfass:
Siimav mentioned in Discord that a PR would be welcome to add **kOS disk space** to the descriptions of the procedural avionics upgrade parts in the tech tree.
Since it was me who complained about missing this information, it seems only fair that I take this upon myself. This is my very first PR ever, so please let me know if I'm not following the correct workflow.


siimav:
Looks good to me

PR #1583, opened 2021-11-06
Add kerbal chute and eva pack selection to KCT crew assignment dialog

siimav:
To support KSP 1.11 extracting chute and jetpack to separate inventory items that need to be assigned to the inventory of every kerbal. The stock logic of configuring inventories inside the VAB doesn't work too well since KCT assigns nauts to crewable parts in space center scene after the vessel has been built and rolled to the pad.

Depends on https://github.com/KSP-RO/RealismOverhaul/pull/2537 and https://github.com/KSP-RO/ROKerbalism/commit/27a696fde4daca919c4f50a43a91d3428b1db261.

image


siimav:
The DLLs are now included in the commit since the bot isn't yet set up to compile against 1.11 or 1.12

PR #1584, opened 2021-11-10
Adding Reputation and Hype to career logging

bongotastic:
Tested on RP1 install with and without Headlines. The first field, `Reputation` is KSP's reputation. The field `hype` is unique to Headlines and thus 0 when the mod isn't installed. We want one plot showing Reputation as one polygonal curve, and another curve for Reputation+hype. 

Not sure if the target of the PR should be master.

### DIFF
--- a/Source/CareerLog/CareerLog.cs
+++ b/Source/CareerLog/CareerLog.cs
@@ -482,7 +482,7 @@ namespace RP0
                 otherFees = logPeriod.OtherFees - constructionFees,
                 fundsGainMult = logPeriod.FundsGainMult,
                 numNautsKilled = logPeriod.NumNautsKilled,
-                Reputation = logPeriod.Reputation,
+                reputation = logPeriod.Reputation,
                 headlinesHype = logPeriod.HeadlinesHype
             };
         }
--- a/Source/CareerLog/CareerLog.cs
+++ b/Source/CareerLog/CareerLog.cs
@@ -315,7 +315,7 @@ namespace RP0
                     p.EntryCosts.ToString("F0"),
                     constructionFees.ToString("F0"),
                     (p.OtherFees - constructionFees).ToString("F0"),
-                    p.KSPReputation.ToString("F1"),
+                    p.Reputation.ToString("F1"),
                     p.HeadlinesHype.ToString("F1"),
                     string.Join(", ", _launchedVessels.Where(l => l.IsInPeriod(p))
                                                       .Select(l => l.VesselName)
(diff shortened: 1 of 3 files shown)


bongotastic:
One of my commit was just pulling from master since I last worked on my fork. Not sure how to make this go away, or whether it needs to.

Source/CareerLog/CareerLogDto.cs
```diff
@@ -26,6 +26,8 @@ internal class CareerLogDto
         public double constructionFees;
         public double otherFees;
         public double fundsGainMult;
+        public double KSPReputation;
+        public double HeadlinesReputation;
```
siimav, line 30: Inconsistent naming

bongotastic: Renamed to HeadlinesHype.

siimav: No, I meant Camel case vs Pascal case

bongotastic: Thanks for the clarification. Fixed.

Source/CareerLog/CareerLog.cs
```diff
@@ -474,7 +481,9 @@ private CareerLogDto CreateLogDto(LogPeriod logPeriod)
                 constructionFees = logPeriod.OtherFees,
                 otherFees = logPeriod.OtherFees - constructionFees,
                 fundsGainMult = logPeriod.FundsGainMult,
-                numNautsKilled = logPeriod.NumNautsKilled
+                numNautsKilled = logPeriod.NumNautsKilled,
+                kspReputation = logPeriod.KSPReputation,
```
siimav, line 485: Since you named the Headlines field to `headlinesHype` then it's better to use just `reputation` here

bongotastic: ksp isnot necessary. Renamed in both CareerLog and LogPeriod

Source/CareerLog/CareerLogDto.cs
```diff
@@ -26,6 +26,8 @@ internal class CareerLogDto
         public double constructionFees;
         public double otherFees;
         public double fundsGainMult;
+        public double Reputation;
```
siimav, line 29: Uhh, inconsistent casing strikes yet again

bongotastic: Alright. One class used camelCase, the other Pascal case as far as I can see. I think that things are consistent. Can you please suggest alt names if not.

siimav (approved):

bongotastic:
Thanks for the review!

PR #1589, opened 2021-11-17
Reconfigure naut G limits

siimav:
* Pilot, engineer and scientist now have the same G multipliers
* Each experience level adds 0.025 to the multiplier
* Badass flag adds 0.05 to the multiplier
* Courage no longer has any effect on G mult
* Set max sustained at 7.5G
* As an unrelated change - make chute available from experience lvl 0 for all naut classes

### DIFF
--- a/GameData/RP-0/CrewSettings.cfg
+++ b/GameData/RP-0/CrewSettings.cfg
@@ -1,15 +1,17 @@
 @PHYSICSGLOBALS:FOR[RP-0]
 {
-	@kerbalGBraveMult = 1
-	@kerbalGBadMult = 1.05
-	@kerbalGOffset = 3165
+	@kerbalGBraveMult = 1           // Courage stat no longer has any effect on G tolerances
+	@kerbalGBadMult = 1.025         // Down significantly from the 1.5 mult that stock uses
+	@kerbalGOffset = 1786           // Gives 6.5 sustained G's for Engineers and Scientists. Pilots have their own multiplier on top that brings the value to 7.5.
+	@kerbalGThresholdLOC = 90000    // A slight increase from the stock 60 000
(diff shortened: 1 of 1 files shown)


NathanKell:
I am slightly skeptical of pilots and engy/scientists having the same multiplier. I guess this depends on what you think roles mean, but to my mind basically every Apollo astronaut with the exception of Jack Schmitt was a Pilot, since every astronaut until Schmitt was a former test pilot.

Further, from a gameplay perspective--clearly scientists have use (labs), but pilots don't--MJ and avionics handle attitude control so SAS ability has no meaning.

I would thus be inclined to lower the multipliers for engineers and scientists such that the max sustained G is something like 6 for them--i.e. they'll be able to deal with Apollo reentries, even lunar, but not like other ones.

siimav:
Apparently there's a bug in how KSP scales sustained G's. It should be linear with `GToleranceMult` but is currently exponential instead (^1/4 with the default settings). Got said that the next version of Community Fixes will patch that.

NathanKell:
Per discord this is purely visual, under the hood they are indeed linear in practice, just what's shown in the naut complex is wrong.

siimav:
Made quite a lot of changes to the values:
* Engineer and scientist start out at 6.5 sustained Gees; pilot at 7.5
* First experience levels add more to the multiplier compared to later ones
* Reduced Badass effect from 0.05 to 0.025
* Increased `kerbalGThresholdLOC` by 50% which means that the G bar will fill up more slowly after going above the sutained G limit

siimav:
For reference this is what the new G mult and sustained G progression looks like for naut levels.

image

NathanKell (approved):

PR #1595, opened 2021-11-27
SSS and ReDirect Configs

ZombieZilla:
SSSs ET is now split into three separate parts, from [#2553](https://github.com/KSP-RO/RealismOverhaul/pull/2553). 1.25M entry cost on the SWT is from the RSB Shuttle ET entry cost.

These are my first RP-1 configs, so please point out anything wrong with them.


siimav:
This needs to use entry cost modifiers. Currently you have to pay 1.25M for the basic tank but LWT can be obtained for just 30k without ever paying for the standard tank. ECMs would allow paying 1.25M + 30k if you want to go straight for the LWT.

PR #1596, opened 2021-11-28
External Tank rework

ZombieZilla:
Space Shuttle External Tank rework. Entry costs now require the previous tank and tank technology, and are more accurate to the real developmet cost. Everything here should be right.


siimav:
Is the cost of the parts supposed to come entirely from whatever RF calculates them to be based on tank definitions?

ZombieZilla:
> Is the cost of the parts supposed to come entirely from whatever RF calculates them to be based on tank definitions?
I was looking at using real costs, but I couldn't find enough sources, and the ones I did varied a lot (62-116million for the LWT, 150-170million for the SWLT). 
Currently the tanks are 40k funds ingame, so do you think I should switch to a rough approximation of the real costs?

StonesmileGit:
Shouldn't all entryCosts be 0, since the entry cost is the ECMs?

StonesmileGit:
Looking at [this source](https://web.archive.org/web/20110629192620/http://science.ksc.nasa.gov/shuttle/nexgen/Nexgen_Downloads/NASA_Shuttle_Zero_Base_Cost_pt1.pdf) (page 10) gives a unit cost for the ET of 10.2k funds looking at the 10 flights/year numbers (converted from 1995 USD)

Capkirk123:
Just gonna merge this and fix the entry cost/ECM issues myself. Regarding the ET prices, that can probably be a separate issue, I'm pretty sure the current tank prices are completely arbitrary.

PR #1597, opened 2021-11-30
Add ECM, tech config for high performance prototype engines

Capkirk123:
Add ECM and tech config for ROE LR-129, RD-270, RD-270M and RD-301


Capkirk123:
Conflicts resolved

PR #1614, opened 2022-01-12
Configure NFP Engines

Capkirk123:
Configure and price NFP engines as best as possible. Costs calculated off the only good data points I could find (NSTAR and P5) and extrapolated from there based on type and design power consumption. Also removed RP-0 tags from obviously not configured stock NFP parts.


siimav:
Depends on https://github.com/KSP-RO/RealismOverhaul/pull/2574

Capkirk123:
Conflicts resolved

PR #1621, opened 2022-01-24
Add support for FWC-SRM engine config

StonesmileGit:
Places the FWC-SRM config one node after the redesigned (Post Challenger) RSRM, even though it was developed before. This is for game-play reasons, since having two configs in the same node would be odd, and placing the better config later seems like the better choise, especially since it never flew.


siimav:
Depends on https://github.com/KSP-RO/RealismOverhaul/pull/2587

PR #1623, opened 2022-01-27
Fix RF basecost patch not getting applied

StonesmileGit:
The patch to set RealFuels tanks cost to 0 when the mass is config-defined instead of code driven didn't apply to any parts due to being in a too early pass.

**This change can affect a lot of parts, since over 1200 parts in RO set the mass of RF tanks in this way.  Discussion is needed reagarding if this is the correct way to fix it!**


NathanKell:
YOLO.

PR #1625, opened 2022-01-31
Move D2 Block II interstages to matureCapsules

arrowmaster:
The Block II 1-engine Skirt is in matureCapsules and is needed for the Block II interstages


NathanKell:
Seems fine to do the reverse, move the others back?

PR #1644, opened 2022-03-29
Heat Control configs, better ECMs

Capkirk123:
Add configs for HeatControl radiators. Add some better ECMs for nuclear reactors and generators.


siimav:
Radiator parts depend on https://github.com/KSP-RO/RealismOverhaul/pull/2613

PR #1649, opened 2022-03-31
Split X-405H configs

Capkirk123:
Split X-405H proposals into one config without restart, and one with restart. Also adjust mass a bit based on Convair documents.

RO PR: https://github.com/KSP-RO/RealismOverhaul/pull/2619


NathanKell:
If the base config can no longer restart, why does it still require the PumpReignition ECM?

PR #1659, opened 2022-04-23
Make Basic Solids depend on Post-War Mat sci

lpgagnon:
Because gating it behind the 7sci of Early Mat Sci basically
made it useless.

per https://discord.com/channels/319857228905447436/620690446540341261/964606272617996298

Also shifted the solid nodes a bit to make the tree a bit clearer.

Screen Shot 2022-04-23 at 13 09 45


siimav:
LGTM

PR #1661, opened 2022-04-25
Support new HG-5 antenna model

TwistedGiraffe:
Fix the issue mentioned in https://discord.com/channels/319857228905447436/512556137380315139/964727589581258802.
Add RP-1 configs for `HighGainAntenna5_v2`.
Requires [RO-#2639](https://github.com/KSP-RO/RealismOverhaul/pull/2639).


siimav:
Since RO no longer configures the legacy part, we might as well remove all configs for it here as well.

PR #1667, opened 2022-05-08
Larch configs for RP-1

DeltaV1ms:
Adds RP-1 configs for the new Larch configs I PR'd to RO

Currently set to unlock in the OR1966 node since that's when Lark historically finished development but that's probably too late for them to be useful so if it's reasonable to move them forward that'd be nice


NathanKell:
Yeah I'd move them back to a node or two after the existing Gamma configs. I'll do that post merge. Cheers!

NathanKell:
Done.

NathanKell:
(moved to 1961)

PR #1670, opened 2022-05-14
Improve RnD UX for experimental parts and under-research tech nodes

siimav:
* Remove *Purchase all* button from under-research nodes
* Remove *Purchase* button from individual parts that are in nodes that are under-research
* Add *Purchase* button to experimental parts that have completed research
* KSP itself and hopefully also contracts no longer consider experimental parts as being available


NathanKell:
Nice! :D

PR #1691, opened 2022-06-16
Move Juno 45K to 1960

Capkirk123:
Move Juno 45K to 1960 rocketry with the Juno 6K engine. It's somewhat underpowered by 1963, and fits in better with it's X405H competitor.

( could've sworn I did this years ago)


NathanKell:
You moved the 6K back, but not the 45K back I think.

Might it be better in orbitalRocketry1961 ? Then there's separation between the X-405H and it, especially since you're removing multiple ignitions from the first X-405H.

NathanKell:
Oh! Huh. ok.

PR #1693, opened 2022-06-19
Programs (and LC) PR - DO NOT MERGE

NathanKell:
Programs is a new way of playing RP-1, where contracts are grouped into overarching Programs. Use the Administration Building to select your starting Program(s) and go from there. You will only have access to the contracts appropriate to your current Program(s), and those contracts will only reward rep. Income comes from the your programs' budgets, though you also get a large subsidy (towards salary and other maintenance costs). This subsidy increases year-over-year until the mid-60s, and varies between a floor and 2x the floor based on your level of rep (1 unit of rep = 100 funds).

This is in HEAVY WIP but is ready for preliminary testing. It is *not* backwards compatible, so START A NEW SAVE.

Instructions: You can replace your existing LC install (or start fresh from an RP-1 Express install). In addition to replacing your RP-0 folder with the one from the comment below (which includes bins, note the repository itself does not), see the following notes:

NOTE: REQUIRES THIS VERSION OF CONTRACT CONFIGURATOR. Replace your CC folder with the one from the Archive. https://github.com/KSP-RO/ContractConfigurator/releases/tag/2.0.0

NOTE: Requires the LATEST ModuleManager (4.2.2). Update it through CKAN!


GameData/RP-0/Contracts/Programs.cfg
```diff
@@ -0,0 +1,679 @@
...
+		{
+			complete_contract = orbitJupiter
+			ATLEAST
+			{
+				count = 2
+				complete_contract = flybyIo
+				complete_contract = flybyEuropa
+				complete_contract = flybyCallisto
+				complete_contract = flybyGanymede
+		}
```
HebaruSan, line 444:
```suggestion
			}
		}
```

There's a missing close brace.

NathanKell: Thanks! Committed separately. Cheers for being awesome, as always. <3

NathanKell:
Closing so artifacts regenerate

PR #1711, opened 2022-07-03
Remove DBS from conflicts in netkan

StonesmileGit:
Dynamic Battery Storage disables itself when Kerbalism is installed, so it doesn't cause `incoherent behaviour`


NathanKell:
Does Kerbalism no longer complain? I conflicted it because, despite DBS having that, Kerbalism still complained.

Capkirk123:
I can confirm that the current release of DBS fully disables itself, and Kerbalism does not complain.

PR #1713, opened 2022-07-06
Only discount used engine EC when in editor scene

ryanc55:
Resolve conflict with EC use in recovery formula.


NathanKell:
Gotcha. Thanks!

PR #1720, opened 2022-07-15
Suborbital Research program requirements description

aeroeng91:
PR to address [KSP-RO/ProgramsIssueTracker#61](https://github.com/KSP-RO/ProgramsIssueTracker/issues/61).

Allow for program requirements description to explicitly state that a contract only needs to be completed once.


NathanKell:
These are getting reworked per https://github.com/KSP-RO/ProgramsIssueTracker/issues/56 but I'm merging meanwhile. Cheers!

PR #1723, opened 2022-07-16
Contract Splits, Optionals and Exclusion

Elouda:
-Split X-Plane contracts into separate contracts for required, program capstones and optionals.
-Remove cooldown from non-optionals.
-Improve desciptions for contracts


pap1723:
Tested, works great!

PR #1725, opened 2022-07-19
Unlock cost and tech list on KCT build window

dgeastman:
Display the cost to unlock any unpurchased parts/upgrades. If any technologies are missing, list those.

### DIFF
--- a/Source/KerbalConstructionTime/GUI/GUI_Editor.cs
+++ b/Source/KerbalConstructionTime/GUI/GUI_Editor.cs
@@ -88,11 +88,28 @@ namespace KerbalConstructionTime
             }
 
             if (KCTGameStates.EditorIntegrationCosts > 0)
-                GUILayout.Label($"Integration Cost: {Math.Round(KCTGameStates.EditorIntegrationCosts, 1)}");
+                GUILayout.Label($"Integration Cost: √{Math.Round(KCTGameStates.EditorIntegrationCosts, 1):N1}");
 
             if (KCTGameStates.EditorRolloutCosts > 0)
-                GUILayout.Label($"Rollout Cost: {Math.Round(KCTGameStates.EditorRolloutCosts, 1)}");
+                GUILayout.Label($"Rollout Cost: √{Math.Round(KCTGameStates.EditorRolloutCosts, 1):N1}");
 
+            if (KCTGameStates.EditorUnlockCosts > 0)
+                GUILayout.Label($"Unlock Cost: √{Math.Round(KCTGameStates.EditorUnlockCosts, 1):N1}");
+
+            if (KCTGameStates.EditorRequiredTechs.Any())
+            {
+                string techLabel = string.Empty;
+                foreach (string techId in KCTGameStates.EditorRequiredTechs)
+                {
+                    string techName = ResearchAndDevelopment.GetTechnologyTitle(techId);
+
+                    if (string.IsNullOrEmpty(techLabel))
+                        techLabel = $"Needs: {techName}";
+                    else
+                        techLabel += $"\n\r       {techName}";
+                }
+                GUILayout.Label(techLabel);
+            }
             GUILayout.BeginHorizontal();
             if (GUILayout.Button("Simulate"))
             {
--- a/Source/KerbalConstructionTime/Utilities/Utilities.cs
+++ b/Source/KerbalConstructionTime/Utilities/Utilities.cs
@@ -1459,6 +1459,10 @@ namespace KerbalConstructionTime
                 KCTGameStates.EditorRolloutCosts = 0;
                 KCTGameStates.EditorRolloutTime = 0;
             }
+
+            var unlockInfo = GetVesselUnlockInfo(ship);
+            KCTGameStates.EditorUnlockCosts = unlockInfo.Item1;
+            KCTGameStates.EditorRequiredTechs = unlockInfo.Item2;
         }
 
         public static bool IsApproximatelyEqual(double d1, double d2, double error = 0.01)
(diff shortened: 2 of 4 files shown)


Source/KerbalConstructionTime/Utilities/Utilities.cs
```diff
@@ -2122,6 +2126,131 @@ public static void CleanupDebris(string launchSiteName)
...
+                    // this part is not in the ECM, take the raw cost
+                    runningCost += rawCost;
+                }
+            }
+
+            // filter down further to those parts that can't be unlocked with our current tech and get the tech names needed
+            var lockedParts = purchasableParts.Where(kvp => kvp.Value.Status == PurchasabilityStatus.Unavailable).Select(kvp => kvp.Key).ToList();
+            List<string> pendingTech = new List<string>();
+
+            foreach (AvailablePart p in lockedParts)
```
siimav, line 2162: Could simplify this entire loop to `lockedParts.Select(p => p.TechRequired).ToHashSet()`

Source/KerbalConstructionTime/Utilities/Utilities.cs
```diff
@@ -2122,6 +2126,131 @@ public static void CleanupDebris(string launchSiteName)
...
+
+        public static List<string> RemoveParentsFromTechList(List<string> input)
+        {
+            List<string> blacklist = new List<string>();
+            SortedList<string,string> slist = new SortedList<string, string>();
+            foreach(string s in input)
+            {
+                foreach (string parent in KerbalConstructionTimeData.techNameToParents[s])
+                {
+                    if (!blacklist.Contains(parent)) blacklist.Add(parent);
```
siimav, line 2208: HashSet would allow straight add without even having to check that this entry already exists.

Source/KerbalConstructionTime/Utilities/Utilities.cs
```diff
@@ -2122,6 +2126,131 @@ public static void CleanupDebris(string launchSiteName)
...
+            var techList = RemoveParentsFromTechList(pendingTech);
+            float totalCost = runningCost + Convert.ToSingle(ecmCost);
+            KCTDebug.Log($"Vessel parts unlock cost check. Total: {totalCost}, Raw cost: {runningCost}, ECM cost: {ecmCost}");
+            return new Tuple<float, List<string>>(totalCost, techList);
+        }
+
+        public static List<string> RemoveParentsFromTechList(List<string> input)
+        {
+            List<string> blacklist = new List<string>();
+            SortedList<string,string> slist = new SortedList<string, string>();
```
siimav, line 2203: Missing space

Source/KerbalConstructionTime/Utilities/Utilities.cs
```diff
@@ -2122,6 +2126,131 @@ public static void CleanupDebris(string launchSiteName)
...
+                foreach (string parent in KerbalConstructionTimeData.techNameToParents[s])
+                {
+                    if (!blacklist.Contains(parent)) blacklist.Add(parent);
+                }
+            }
+            foreach (string s in input)
+            {
+                if (!blacklist.Contains(s))
+                {
+                    // sort our result, depth into the tree then alpha
```
siimav, line 2215: Method name should also imply that it does some kind of sorting. Or maybe split it into 2 separate methods?

Source/KerbalConstructionTime/Utilities/Utilities.cs
```diff
@@ -2122,6 +2126,131 @@ public static void CleanupDebris(string launchSiteName)
...
+                    int depth = KerbalConstructionTimeData.techNameToParents[s].Count();
+                    string skey = $"{depth:d2}{s}";
```
siimav, line 2217: Since we're already using LINQ pretty much everywhere, maybe `OrderBy().ThenBy()`?

siimav (changes requested):

Source/KerbalConstructionTime/Utilities/Utilities.cs
dgeastman: I could make it a list of Tuple<int,string>, put those in a list, and then return that with the OrderBy().ThenBy(), but I don't see that it would be any better. The int for depth is not going to be unique. We could just do a pure alpha order rather than trying to use depth as a proxy for lateness in the tree.

dgeastman: I mean, if you deliberately tried to maximize it, you might get as many as a dozen entries in this list, and the sort is only going to need to get a few characters deep into most names to figure out position. Optimizing that seems kind of pointless.

Source/KerbalConstructionTime/GUI/GUI_Editor.cs
```diff
@@ -88,11 +88,28 @@ private static void RenderBuildMode()
             }
 
             if (KCTGameStates.EditorIntegrationCosts > 0)
-                GUILayout.Label($"Integration Cost: {Math.Round(KCTGameStates.EditorIntegrationCosts, 1)}");
+                GUILayout.Label($"Integration Cost: √{Math.Round(KCTGameStates.EditorIntegrationCosts, 1):N1}");
```
NathanKell, line 91: This and following don't need the Math.Round if you're doing :N1.

Source/KerbalConstructionTime/Utilities/Utilities.cs
```diff
@@ -2122,6 +2126,131 @@ public static void CleanupDebris(string launchSiteName)
...
+
+            // compare the part specified entry cost to the ECM database
+            foreach (AvailablePart part in purchasableParts.Select(kvp => kvp.Key))
+            {
+                int rawCost = part.entryCost;
+                double ecmEstCost = RealFuels.EntryCostManager.Instance.ConfigEntryCost(part.name);
+                if (rawCost == ecmEstCost)
+                {
+                    // this part is managed by the ECM, save its name later for a batch request
+                    string sanitizedName = RealFuels.Utilities.SanitizeName(part.name);
```
NathanKell, line 2147: In looking through the RF ECM code, turns out it sanitizes for you so locally sanitizing is unnecessary.

Source/KerbalConstructionTime/Utilities/Utilities.cs
```diff
@@ -2122,6 +2126,131 @@ public static void CleanupDebris(string launchSiteName)
...
+            List<AvailablePart> lockedParts = purchasableParts.Where(kvp => kvp.Value.Status == PurchasabilityStatus.Unavailable).Select(kvp => kvp.Key).ToList();
+            HashSet<string> pendingTech = new HashSet<string>(lockedParts.Select(ap => ap.TechRequired));
+
+            // now back through the list looking for upgrades to add to our batch list
+            foreach (Part p in ship.Parts)
+            {
+                var mecs = p.FindModulesImplementing<RealFuels.ModuleEngineConfigsBase>();
+                foreach (RealFuels.ModuleEngineConfigsBase mec in mecs)
+                {
+                    string sanitizedName = RealFuels.Utilities.SanitizeName(mec.configuration);
```
NathanKell, line 2168: ditto

Source/KerbalConstructionTime/Utilities/Utilities.cs
```diff
@@ -2122,6 +2126,131 @@ public static void CleanupDebris(string launchSiteName)
...
+                        continue;
+                    
+                    ecmPartsList.Add(sanitizedName);
+                    ConfigNode n = mec.configs.Find(c => c.GetValue("name") == mec.configuration);
+                    if (n != null)
+                    {
+                        string techRequired = n.GetValue("techRequired");
+                        if (techRequired == null) continue;
+
+                        ProtoTechNode techState = ResearchAndDevelopment.Instance.GetTechState(techRequired);
```
NathanKell, line 2179: You might want to consider just adding the techRequired for this (and for upgrades) to a hashset, and only at the end checking techstate for them all, to avoid duplication? Otherwise. Same with having ecmPartsList as a hashset. Otherwise you'll be doing a bunch of contains checks for every symmetry counterpart engine, etc.

NathanKell (changes requested):
Marked as "request changes" even though a couple are suggestions. :)

Source/KerbalConstructionTime/Utilities/Utilities.cs
dgeastman: I think we saw that of the various RF calls I make, not every one of them sanitizes. And I have to make sure I have a matchable name in my local collections, so I sanitize on this side.

Source/KerbalConstructionTime/Utilities/Utilities.cs
dgeastman: The contains check at the top of this block was to avoid doing the rest of the logic for duplicate parts. I think the contains there is cheaper than finding the config and then pulling it's tech repeatedly. I originally had it this way as I needed to know if it was a duplicate before checking costs, but even with the cost code no longer in this block, I think it's the faster approach.

NathanKell:
Uh, I just noticed--this doesn't handle tanks?? I sent you that dll so you could :)

PR #1729, opened 2022-07-28
Keep both RD-200 and RD-103 as ECMs for the R7 engines.

leudaimon:
I'm not happy with the current name of the RD-200 ECM as RD200-TP, given it was the combustion chamber that was used, but I don't wanna change everything in the browser just for that.


arrowmaster:
I don't think the RD-200 ECM should go in R7-Engines as that effects not only the RD-107/108 but also the RD-0105 and all other R7 upper stage engines after it. While the RD-103 was a key part of early soviet rockets and would impact all open chamber designs. That is why I chose to place it with R7-Engines in PR #1728.

While it is the RD-200 multichamber design and not the turbopump that is used, it could be hand waved as 'multichamber design and turbopump *piping* to multiple chambers' as an excuse to not have to rename everything.

As it was @Capkirk123's PR that changed the ECM from using the RD103-TP to the RD200-TP, I think their input would be valuable.

Capkirk123:
Yes, that was poorly named (I don't think the RD200 even has a turbopump). The RD-200 was the testbed for the unitary showerhead injector used by the RD-107/108, while the turbopump was mostly just a scaled up RD-10x TP.

NathanKell:
Where are we at with this?
(Also it has merge conflicts)

NathanKell:
Manually making the change in master.

PR #1732, opened 2022-08-08
Charlie Bossart

badgermasher:
Add Charlie Bossart as a leader for the programs branch of RP-0

Gives +15% on Balloon tank speed (maybe egregrious) as well as a +10% research rate to Materials nodes.


NathanKell:
That seems...rather OP? Also has no downside. I'd suggest leaving the +15% balloon tank upside and removing the materials science buff, and picking out some small thematic downside?

Also should only be in Engineering, the fact that many other leaders are duplicated is an issue that's getting fixed. :)

badgermasher:
I have no strong feelings one way or the other when it comes to leader balance so I have gutted the bonuses and added a -5% rep penalty to reflect the fact that Bossart was not a very public figure. Additionally I deleted the administrator version, I thought it was probably not intended at the time but everyone else was duplicated.

NathanKell:
Shoot, shoulda merged this before the force push. Pap has you covered though with the leader rework! Hopefully it's to your liking :)

PR #1733, opened 2022-08-09
Improved grammar and clarity

NovemberOrWhatever:
Using a common noun as a proper noun ("Leader" in this case) can be good for clarity, but can make it hard to be consistent. I've updated all uses to be uppercase.


NathanKell:
Merged manually.

PR #1739, opened 2022-08-12
Consider resources in LC closeness check

dgeastman:


NathanKell:
Gonna have to think about this a bit because this change makes it no longer cheap to run the method and I didn't consider that. Probably have to cache closeness between LCEs now, since it runs a *lot*.

PR #1745, opened 2022-08-13
Fix dirty enumerator issue

dgeastman:


NathanKell:
I am deeply confused by the purpose of this PR--you're replacing a stack-alloc struct enumerator with a heap-alloc list?

dgeastman:
For some reason the enumeration was getting modified during this process so it would throw an error and you wouldn't be able to modify an LC. So this pulls a static list of the keys so any modification becomes irrelevant.

NathanKell:
Sounds like the case where this is true the source and dest dict pointers are identical, so seems better to test for that.

NathanKell:
Unrelated to the actual bug, per discord.

PR #1753, opened 2022-08-15
Improved Leader descriptions, grammar of contracts.

NovemberOrWhatever:


NathanKell:
Not just named after him! He invented it! As the altitude at which the speed necessary for lift to cancel gravity, given air density, matches orbital velocity--he rounded it off to 100km. (That means it exists for any body with an atmosphere, fwiw :D )

NathanKell:
Merging this as I need to do a force push at some point and that'll screw history.

PR #1756, opened 2022-08-18
Split cost of stock rd855

lpgagnon:
similar to #1346: 350 funds for each of 4 verniers doesn't seem to make
sense, especially when paired with 3x 2-nozzle rd250 (which would cost
1140, based on 760 for ROE's 4-nozzle rd250) for a tsyklon first stage.

so assume 350 was meant to be the cost of the full 4-vernier engine,
and divide by 4. RO already divides mass & thrust by 4.


Capkirk123:
This seems reasonable. Looks good to go!

PR #1758, opened 2022-08-21
Fix minor errors in program descriptions, improve objectivesPrettyText, and add Von Braun Leader description

NovemberOrWhatever:


ZombieZilla:
This needs to be targeted towards the NotBudgets branch

PR #1761, opened 2022-08-22
FirstPR! add Arthur "Val Cleaver", British Aerospace Engineer as a leader

Beasty-One:
I tried to give a little bonus in both areas he developed well during his career (despite Sprite not being in-game *shrug*), and Hydrolox will suffer. The RZ.20, as good as it was, never really took off (pun intended). I used a few of the effects from other leaders as templates so those should be fine, I dont know all the keywords.


NathanKell:
Great start!
Will have some inline comments.

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
```diff
@@ -389,3 +389,59 @@ STRATEGY
 		}
 	}
 }
+
+STRATEGY
+{
+	name = leaderValCleaverEngineer
+	RP0conf = True
+	title = Arthur Valentine Cleaver
+	desc = Arthur "Val Cleaver" is a British Aerospace Engineer working for de Havilland in the early 1950s, responsible for overseeing the development of the early British rocket plane engines, "Spectre" and "Sprite". Later he worked to develop upgraded variants of the "Spectre" engine, that went on to power the Blue Steel missile.  In 1956, Val Cleaver hands in his notice to de Havilland, and takes a job with Rolls-Royce as their Chief Rocket Propulsion Engineer. Under his guidance the development of the RZ.2 and RZ.20 takes place, built to power the Blue Streak missile. Considered highly advanced engines at the time, it later goes on to power the first stage of Europa 1. Arthur Val Cleaver is awarded an Order of the British Empire, a prestigious medal for prominent national achievements.
```
NathanKell, line 398: The other leader descriptions are in the past tense.

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
```diff
@@ -389,3 +389,59 @@ STRATEGY
...
+	removeOnDeactivateTag = leaderValCleaver
+  
+	minDuration = 2592000 // 30 days
+	maxDuration = 315576000 // 10 years
+	reactivateCooldown = 31557600 // 1 year
+	
+	EFFECT
+	{
+		name = ResearchRateModifier
+		effectTitle = Rolls-RoyceDidRocketsGood!
```
NathanKell, line 416: effectTitle is the title shown ingame. It should be a human-readable phrase, yours seem to lack spaces.

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
```diff
@@ -389,3 +389,59 @@ STRATEGY
...
+		multiplier = 1.05
+		effectDescription = Early and Orbital Rocket Engine tech
```
NathanKell, line 418: I would suggest not adding another engineer with a bonus to rockets and a penalty to hydrolox. We've got so many! Why not a bonus integration time for launch vehicles that use HTP? You can look at Mary Sherman Morgan for an example of how to do that.

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
```diff
@@ -389,3 +389,59 @@ STRATEGY
...
+			typeNode = RocketEngines
+		}
+	}
+
+	EFFECT
+	{
+		name = ResearchRateModifier
+		effectTitle = BritishRocketPlanes
+		multiplier = 1.05
+		effectDescription = Flight tech
```
NathanKell, line 430: I'm a little skeptical of this one as well, as we have so many other things that give a bonus to flight research. Can you think of any other advantages?

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
```diff
@@ -389,3 +389,59 @@ STRATEGY
...
+			typeNode = Flight
+		}
+	}
+
+	EFFECT
+	{
+		name = ResearchRateModifier
+		effectTitle = RZ.20 never took off
+		multiplier = 0.9
+		effectDescription = Flight tech
```
NathanKell, line 442: This effect description is copypasta from the previous Effect.

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
```diff
@@ -389,3 +389,59 @@ STRATEGY
...
+		nodeTypes
+		{
+			typeNode = Hydrolox
```
NathanKell, line 445: As above re: hydrolox penalties, it feels like Pap made half the leaders have those!

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
```diff
@@ -389,3 +389,59 @@ STRATEGY
...
+		}
+	}
```
NathanKell, line 447: You are missing a final closing }

NathanKell (changes requested):

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
Beasty-One: I was hoping a splash of British was unique enough, similar enough in rocketry and planes research that it would only be used for A Very British Space Program type deal. Players self made house rules preventing use of American or Soviet nationals ect. (I can't be the only one who does it! aha)
Perhaps Kerolox bonuses as an alternative? I will try figure out the keywords for that

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
Beasty-One: re above yourself!:), its just a British guy doing the same thing as everyone else. I Can think an alternative negative (if one at All?) being tankage, Blue Streak was a corrogated stainless-steel conventional tank (https://www.globalsecurity.org/wmd/world/uk/blue-streak-design.htm) if thats more prefferable!

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
Beasty-One: alright thanks!

pap1723:
I am all for flavor inspired people from different countries. I think eventually down the line I will add an option to have the player restrict to certain ones if they want to.

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
NathanKell: We've tried to make sure no leaders duplicate other leaders, and it would be really good for that to continue. Otherwise adding this leader adds no gameplay and could be better served by some cosmetic mod that just changes names and faces. There's a ton of keywords (in technodedata and in KCTTags.cfg and in CurrencyModifierRP0.cs) so there should be plenty of options

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
NathanKell: For some reason I recalled Blue Streak being balloon, with some stiffening stringers on the outside of the after tank?

NathanKell:
I've made a couple changes to the code that apply here:
* PartIntegrationRateModifier is now just IntegrationRateModifier
* the `resources` field there now takes name/value pairs, instead of `item = <name>` pairs. The values are the minimum of that resource for the bonus (or penalty) to take effect. I would suggest doing something like having an integration rate bonus for the vessel when `HTP = 10000` (which is to say, enough of it that it's probably a primary propellant, not just a turbopump-powerer) and as a penalty you could do something like a 0.95 multiplier when there's more than 2000 units of (all the solid resources).

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
Beasty-One: Might well be, I dont know the legitimacy of the source, what caught my eye was "The skin was supported on a riveted structure of internal frames and stringers" and "~was further stiffened by spotwelding longitudinal top-hat stringers around its outer circumference." though the thickness of the sheets is what you would see with an Atlas Balloon tank, a bit of a mashup of sorts perhaps

NathanKell: _A Vertical Empire_ might have the fullest details?

Beasty-One:
Thanks I did notice from your prompt to Zephyr, I have made those changes without commit yet. I will steal your idea of rate penalty when lacking those resources instead of the straight Solids nerf, thanks.
I have also made an Alan Bond as I envisioned, and a Australian born British physicist Sir Harrie Massey, will make a new PR for those once I get Val done and good enough

NathanKell:
Ah great you saw that and are making the changes!
Not sure what you mean by "rate penalty when lacking those resources" -- the system supports having a multiplier that is active when, for each resource listed in the `resources` block and for each resource on the vessel, the listed amount is <= the vessel's amount.
Example:
multiplier = 1.1
HTP = 10000
=> will apply a 10% bonus when there's >=10k liters of HTP on the vessel.

multiplier = 0.95
PUPE = 4000
PSPC = 4000
HTPB = 4000
... etc

will apply a 5% penalty if at least one of those resources is present in the given or greater amount on the vessel. A vessel with 2k PUPE and 5000 HTPB and 0 PSPC would not trigger it, whereas one with 5000 PUPE but no HTPB and only 100 PSPC would trigger it.

There isn't support for "unless present" checks, just "present" checks.

Beasty-One:
Ok I think I got it right.
I've set every solid prop (if I didnt miss any) to 500 with a multiplier of 0.9
So alot of earlier kickstages up to the Waxwing can slip by, as it only having 175 units of CTPB wouldnt get penalized if the only solid source on there? Or a good amount of Medium separation motors wont trigger the PSPC penalty.

GameData/RP-0/Strategies/Leaders/LeadersEngineers.cfg
Beasty-One: I think you're right, seems a good source, it reads there were only external stringers on the rear tank that would support the weight of the front tank to be supported without pressuring the rear tank. and the upper tank Was pressurized

NathanKell:
Not sure you want Kerosene in there--that means you'll get a bonus to every LV that has >=10k liters of Kerosene *or* 10k liters of HTP. Or is that the intent? (I.e. it'll give a bonus to all the kerolox and hydrolox stuff too--the HTP is basically superfluous at that point)

NathanKell:
Ah, nm, you say and/or, so I guess so?

The solids thing looks correct to me! :)

Beasty-One:
Yeah I cant really Force people to use anything with HTP, but it also gets a Gamma bonus, as no Gamma on its own at full burn will cross that kerosene margin. other big dumb Kerolox boosters are still a win in my books!

NathanKell:
This had errors that broke RP-1 until I fixed it just now. Please actually test your stuff before submitting the PR (or leave it on Draft until you *have* tested your stuff.

NathanKell:
The wording is also busted.
image


I'm...really pretty disappointed that you clearly didn't test this ingame, despite mentioning that you were going to need to in the channel.

PR #1778, opened 2022-09-06
Add RP-1 support for ROC BDB Apollo

Capkirk123:
Add support for ROC BDB Apollo, and adjust the values of existing Apollo parts


LuneoLucy:
It seems that the CM and LEM lack avionic modules, they are uncontrollable without adding proc avionics

LuneoLucy:
BDB's apollo rcs are also unconfigured, but I do not know if it's on purpose or not

PR #1780, opened 2022-09-06
Use reusable workflows from KSP-RO/BuildTools

StonesmileGit:


siimav (approved):
Most of those got tested with CC and they appear to work as advertised.

PR #1785, opened 2022-09-15
Reset X-plane contracts when vessel gets destroyed

siimav:
Leverage the new CC resetChildrenWhenVesselDestroyed option to fix X-plane contracts staying locked to a destroyed vessel.

**NEEDS https://github.com/KSP-RO/ContractConfigurator/releases/tag/v99.9.9.3 TO WORK**


Test-Account-KSP:
Tested with a copy of an actual career
Accepted X-Planes High contract, rejected it, accepted again
Airlaunched the plane, reached the contract altitude 
screenshot9
Crashed
Built a new plane and flied the mission again, without rejecting and reaccepting the contract. 
Nothing changed, the contract is not completed

screenshot12

PR #1793, opened 2022-09-25
Add color images for every currently existing program

Falki-git:
Added color images for programs that didn't have any; replaced existing grayscale images with new colored ones.


NathanKell:
It would be good to keep the existing clean style; many of these will be hard to make out as small icons, and are often not easily distinguished from other similar icons. But thank you for working on this!! :)

Falki-git:
They work quite well for me. Added realism and color make the programs more vibrant, more joyful to pick. But I understand if you want to keep the simple style. Thought I could chip in! :)

KSP_admin_programs

PR #1808, opened 2022-10-09
Configure some FireSpitter cockpits

lpgagnon:
A start-node biplane cockpit; essentially useless, but cheap and fun.

A cockpit to sit in-between start-node and F-100 cockpits. Not placed in supersonicDev to give X-1 one node to shine alone; maybe overly cautious, given the ease of going supersonic in a Stuka anyway.

Could instead embrace the 'x-1's niche is altitude records' meta, but it just feels wrong.


Capkirk123 (approved):
A little silly, but why not. Looks good.

PR #1821, opened 2022-11-05
Add speclevel to engine CONFIG PARTUPGRADEs

lpgagnon:
Mostly to give a clue why a config isn't actually available despite
the tree showing a PARTUPGRADE for it

(also includes a partupgrade for xlr35, forgot I had that in my tree)

Modifies the parts browser, which possibly means we don't want to merge this without also updating the .exe, and I'm not going to do that.

Screen Shot 2022-11-05 at 14 20 14

### DIFF
--- a/GameData/RP-0/Tree/TREE-Engines.cfg
+++ b/GameData/RP-0/Tree/TREE-Engines.cfg
@@ -4348,6 +4348,32 @@ PARTUPGRADE
         description = The RD58 Engine now supports the 11D33 configuration for increased performance. Unlock it in the VAB/SPH through the engine configs interface.\n\nDeveloped version of S1.5400 engine for Block L stage of Molniya launch vehicle. Slightly better performance.
 }
 
+PART
+{
+    name = RFUpgrade_engineConfigSource_11D33
+    engineType = RD58
+}
+
+@PART[RFUpgrade_engineConfigSource_11D33]:AFTER[RealismOverhaulEngines]
+{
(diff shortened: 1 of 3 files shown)


siimav:
The part browser exe would also need to be rebuilt before it can be merged.

Capkirk123:
Instructions on how to create a new .exe are detailed here https://github.com/KSP-RO/RP-0/pull/1491.
Not sure how to best go about this, but we can always regenerate the .exe after merging if lpg doesn't have Python 3 installed?

lpgagnon:
If it's that simple, is there a reason the exe couldn't be generated as another build artifact?

siimav:
> If it's that simple, is there a reason the exe couldn't be generated as another build artifact?

Problem is that build artifacts are only available in the resulting zip packages. This tool however is pointless to include in those.
It might be possible to make it commit exe changes back to the repo as a separate commit but honestly I'm not liking this idea for various reasons.

lpgagnon:
I meant as a _separate_ build artifact, i.e. its own zip. So long as it's downloadable from somewhere, that seems just as good (better?) than living in the repo.

StonesmileGit:
Took a go at trying to compile the part browser, with no luck. Maybe I did it in some weird way, so might be worth for others to try too.

StonesmileGit:
Merging this without re-building, since another PR also modifies the part browser, so will re-compile there only

PR #1824, opened 2022-11-14
Add ROT-CMES-Tank-* support

lpgagnon:
needs https://github.com/KSP-RO/ROTanks/pull/81 and https://github.com/KSP-RO/RealismOverhaul/pull/2734


Capkirk123:
Looks good, tooling works.

PR #1835, opened 2022-11-24
Add to the Earth Space Station Program

Teykn:
Adds a set of 3 contracts based on the Skylab station. 1st contract requires launching a more advanced station and crewing it with 3 crew for 30d. 2nd contract involves crewing for 60d, and 3rd contract involves crewing for 90d. Both 2nd & 3rd contract can be completed twice, with the former intended to be required and the latter optional.

Also allows `Duration` timers to count down while outside the vessel scene, allowing for players to focus on other contracts. (Should a note/popup be added to let players know they can focus on other contracts?)

### DIFF
--- a/GameData/RP-0/Contracts/Space Stations/Improved Space Station.cfg
+++ b/GameData/RP-0/Contracts/Space Stations/Improved Space Station.cfg
@@ -2,7 +2,7 @@ CONTRACT_TYPE
 {
 	name = improved_spaceStation
 	title = Improved Space Station
-	group = SpaceStations
+	group = EarthSpaceStation
 	agent = Stations
 	
 	description = Design, build, and launch an improved space station with power generation, a science lab, and science experiments into Earth orbit. Once this contract is complete, missions to crew this station will be unlocked. This contract is based off of the Skylab and Second Generation Salyut space stations. 
@@ -18,7 +18,7 @@ CONTRACT_TYPE
 	autoAccept = false
 	minExpiry = 0
 	maxExpiry = 0
-	maxCompletions = 0 //Note: Despite maxCompletions being infinite, the advance/reward funds can only be obtained once.
+	maxCompletions = 0 
 	maxSimultaneous = 1
 	deadline = 730 * RP1DeadlineMult() // 2 Years
 	
@@ -55,7 +55,6 @@ CONTRACT_TYPE
 		minCount = 2
 		invertRequirement = true
 	}
-	
 	REQUIREMENT
 	{
 		name = noStation
@@ -64,13 +63,13 @@ CONTRACT_TYPE
 		vessel = @/craft
 		invertRequirement = true
 	}
-	
 	DATA
 	{
 		type = string
 		craft = "improvedSpaceStation"
 		hidden = true
 	}
+	
 	// ************ PARAMETERS ************
 	
 	PARAMETER
@@ -81,6 +80,8 @@ CONTRACT_TYPE
 		notes = The station must include at least 1 Docking Port, have support for at least 3 Crew, a Science Lab, and can generate power.
 		define = @/craft
 		defineList = basicEarthStations
+		
+		disableOnStateChange = true
 		
 		PARAMETER
 		{
@@ -134,7 +135,6 @@ CONTRACT_TYPE
 			title = Reach Orbit with a Perigee greater than 400 km and an Apogee less than 500 km 
 			hideChildren = true
 		}
-		disableOnStateChange = true
 	}
 	
 	PARAMETER
@@ -174,6 +174,7 @@ CONTRACT_TYPE
 		notes = You can work on other contracts while your crew are working. You are safe to return to Mission Control
 
 		completeInSequence = true
+		disableOnStateChange = true
 		
 		PARAMETER 
 		{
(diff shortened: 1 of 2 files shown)


siimav:
It appears that the cfg file validator found some errors in your files.

Teykn:
Fixed. Had a typo

StonesmileGit:
Regarding a note about being able to leave the station while the timer ticks down:
If this is not the case for all contracts converted to Programs, then yes, some note in the description of the contract or in the parameter in-flight would be nice

siimav:
Have you tested what happens if the crew capsule or the station itself gets destroyed? Does the contract get locked to vessels that no longer exist? If that does indeed happen then take a look at the new VPG parameters that were added for the X-plane contracts.

Teykn:
I've tested it with the station but not the crew capsule. When station is deleted, current contract fails and you start over from the 1st contract. I haven't done that for the crew capsule, but that shouldn't be too hard to implement.

siimav:
Some notes:
* Some of those contracts still have deadlines. This doesn't really make sense with programs and in fact I recently nuked those from all existing ones.
* In some cases disableOnStateChange is at the very end of VPG which is somewhat hard to read
* Noticed a couple of places where the indenting is off a bit

Otherwise if you have validated that these work then I imagine it's good to merge.

GameData/RP-0/Contracts/Space Stations/Improved Space Station.cfg
```diff
@@ -0,0 +1,606 @@
...
+		program = EarthSpaceStation
+	}
+	REQUIREMENT
+	{
+		name = CompleteContract
+		type = CompleteContract
+		contractType = first_ISScrew
+		minCount = 2
+	}
+	RREQUIREMENT
```
Falki-git, line 479: typo

Falki-git: Also, in order for these contracts to show up, [this contract group](https://github.com/KSP-RO/RP-0/blob/NotBudgets/GameData/RP-0/Contracts/Groups.cfg#L124) needs to be renamed to SpaceStations

Teykn: Thanks for pointing out the typo, will get to fixing it.

Wasn’t the contract group and program renamed from SpaceStations to EarthSpaceStation, presumably with a Lunar, Martian, etc programs planned in the future? I noticed that and decided to go along with it.

Falki-git: Sure, but your changes in this PR renamed the group of all contracts back to SpaceStations. So either rename them back to EarthSpaceStation or rename the entire group to SpaceStations, preferably the former :)

GameData/RP-0/Contracts/Space Stations/Improved Space Station.cfg
```diff
@@ -243,7 +243,7 @@ CONTRACT_TYPE
 {
 	name = first_ISScrew
 	title = Crew your Improved Space Station for 60 days and collect science
-	group = SpaceStations
+	group = EarthSpaceStation
 	agent = Stations
 	
 	description = You have succeeded in crewing your station for a month. Now it's time for you to crew it for longer. Crew your station for 60 days and then return safely. You'll be staying for longer intervals in the future. This mission can be completed 2 times. &br;<b><color=red>Do NOT deorbit the station while you are completing this contract. Otherwise, this contract will fail and you will be forced to redo the Improved Space Station contract</color></b> &br;&br;<b>Number of Contracts Completed: $first_ISScrew_Count</b>
```
Falki-git, line 249: Line 274. Rep reward of 25 for first ISS seems a bit low, no?

GameData/RP-0/Contracts/Space Stations/Improved Space Station.cfg
```diff
@@ -447,7 +447,7 @@ CONTRACT_TYPE
 	maxExpiry = 0
 	maxCompletions = 2
 	maxSimultaneous = 1
-	deadline = 365 * RP1DeadlineMult() // 1 year
+	deadline = 0
 	
 	targetBody = HomeWorld()
```
Falki-git, line 453: Line 459. Rep reward here also seems low

Teykn:
@Falki-git Regarding rep rewards, what would be reasonable values? I’m not good at balancing, so tried to set values that weren’t too high

Falki-git:
Neither am I, but looking at other contracts that are at the same "tier", I'd maybe go with:
- first_spaceStation = 300
- improved_spaceStation = 200
- first_ISScrew = 200
- second_ISScrew = 300

For reference, these are some of other contracts currently:
- crewed first orbit = 200
- first EVA = 150
- orbit3 = 250
- lunar crewed flyby = 300 (commented out)
- lunar crewed orbit = 200 (commented out)
- first crewed landing = 400

siimav:
Has anyone got the chance to test the contracts ingame?
I would need to merge this before I can pull in the changes from latest release into NotBudgets.

Teykn:
> Has anyone got the chance to test the contracts ingame? I would need to merge this before I can pull in the changes from latest release into NotBudgets.

Tested it. Parameters involving the crew capsule are reset when it's destroyed, so working as intended. Parameter where the crew stay for X duration isn't reset if it's complete though. Should it be reset if the crew capsule is destroyed? If not, this can be merged.

PR #1838, opened 2022-11-27
Fix KCT not allowing crewing of parts with crew capacity other than the first one found

Falki-git:
Old code filled only the first crewable part found which made the Apollo CM + LEM combo not being able to crew the CM if the LEM was the first part found.

Tested with Apollo CM + LEM, Gemini, Mercury, X-15 and non crewable vessels.

Apollo CM + LEM

### DIFF
--- a/Source/KerbalConstructionTime/GUI/GUI_CrewSelect.cs
+++ b/Source/KerbalConstructionTime/GUI/GUI_CrewSelect.cs
@@ -77,7 +77,7 @@ namespace KerbalConstructionTime
                 Part p = _parts[j];
                 if (p.CrewCapacity == 0) continue;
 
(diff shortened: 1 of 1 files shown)


siimav (changes requested):
Awesome!
Please remove all unused code instead of just commenting it out. It also appears that the `CrewFirstAvailable` method is completely unused now.

siimav:
Would you be willing to backport this to the master branch as well? I suspect that the code is identical other than just being located in a different folder.

Falki-git:
> Would you be willing to backport this to the master branch as well? I suspect that the code is identical other than just being located in a different folder.

Sure, I'll give it a a go.

siimav (approved):

PR #1844, opened 2022-12-04
Rushing builds should increase failiure chances

strandedkerbal:
for realism


siimav:
Well, you aren't the first one who has thought about it.
https://github.com/KSP-RO/TestFlight/issues/180
Also, it might be a good idea to learn the difference between Issues and Pull Requests.

PR #1847, opened 2022-12-06
Add reputation to Lunar Crewed contracts

Falki-git:
Fix for https://github.com/KSP-RO/ProgramsIssueTracker/issues/107 (thanks @Teykn !)

### DIFF
--- a/GameData/RP-0/Contracts/Lunar Crewed/CrewedLunarFlyby.cfg
+++ b/GameData/RP-0/Contracts/Lunar Crewed/CrewedLunarFlyby.cfg
@@ -31,8 +31,8 @@ CONTRACT_TYPE
 	rewardScience = 0
 	rewardFunds = 0
 	failureFunds = 0
-	rewardReputation = 1	// was 300
-	failureReputation = 1	// was 300
+	rewardReputation = 300
+	failureReputation = @rewardReputation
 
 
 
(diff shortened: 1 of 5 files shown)


Falki-git:
As @Teykn already fixed the waypoint issue, renaming this PR

siimav:
Looks like we've so far set the prestige value to Trivial for all contracts

Falki-git:
What is the reason for using only Trivial prestige?

In RP-1 contracts can be accepted regardless of their prestige. Since we're using only reputation for the rewards, prestige multiplier only effects reputation gains. Significant = x1.25, Exceptional = x1.5 so it would be easy to maintain and balance contracts. I might be wrong here somewhere?

Reason for using other prestige levels:
- adds to the immersion - designating a contract for landing first people on the Moon as Trivial feels wrong
- adds sense of fulfillment - completing Significant (2 stars) & Exceptional (3 stars) contracts feels more rewarding
- signals players that contract is harder to achieve
- (optional) in P&LC it could be used to add progress to programs - programs start with Trivial contracts, progress to Significant and finish off with Exceptional

siimav:
Mostly because it's annoying that the reward number you put inside the contract isn't the actual reward. Body multipliers are even worse offenders here though. Maybe we can bring the rep back once that *feature* has been patched out. The prestige values in RP-1 contracts weren't particularly consistent anyway so starting from a blank sheet might be worth it.

Falki-git:
I see your point. Although I think my point regarding user experience and immersion trump the "value is not the actual reward" fact, I'm offering an interim solution -> print out the actual reward in comment.

siimav:
Best to leave them all as trivial right now and tackle this issue separately later on.

Clayell:
> Best to leave them all as trivial right now and tackle this issue separately later on.

Done with https://github.com/KSP-RO/RP-1/pull/2776. Prestige can now be set freely without changing any of the reward values.

Issue #1722, opened 2022-07-15
Crewed records HasCrew is sticky


lpgagnon:
Crewed records can get completed by visiting an uncrewed vessl that satisfies the altitude/speed target right after a crewed vessl that doesn't.

Example: place satellite in orbit; put crewed plane on runway, recover it right away. Go to satellite via tracking station. As soon as satellite is loaded, the next crewed Altitude and Speed records (if less than satellite's state) immediately complete.

Looking in the .sfs right after recovering plane, one can see the next crewed records' HasCrew PARAM has state=Complete, presumably preserved from while the plane was on the runway. 

Don't know if CC bug or misconfigured contract. .cfg just has type=hascrew, mincrew=1; .sfs shows a possibly-fishy `disableOnStateChange = False`

NathanKell:
From what I understand this is a CC bug.

`disableOnStateChange = False` is correct -- if it were true then the parameter, once it went true, would never go to false again.

pap1723:
I think this goes back to some conversations that we ran into when we had the docking and Space Stations contracts written a little differently. What would be great is if a specific craft was "attached" to a specific contract. For example, before a launch, the player would say, this craft is for "x,y,z" contract and CC would read that parameter and then only allow that craft to complete that contract (until the player "removed" the link due to failure).

I don't know how big of a rabbit hole that gets into though.

NathanKell:
I don't think that's really related? The issue is that CC is reading some of a vessel's state (alt, speed) without reading the *rest* of the vessel's state (crewcount)

pap1723:
Ah, looking at how the contract is designed, I bet if we wrap the HasCrew inside of the ReachState it would work properly.

**CURRENT CONTRACT**
PARAMETER
	{
		name = ReachState
		type = ReachState

		minAltitude = @/crewedTargetAltitude
	}
	PARAMETER
	{
		name = HasCrew
		type = HasCrew
		minCrew = 1
	}

**PROPOSED FIX**
PARAMTER
{
	name = ReachState
	type = ReachState
	
	minAltitude = @/crewedTargetAltitude

	PARAMETER
	{
		name = HasCrew
		type = HasCrew
		minCrew = 1
	}
}

NathanKell:
Oooh, good call! It's weird that the param isn't reset on load of vessel, but that should probably fix it?

pap1723:
[...]

PR #1849, opened 2022-12-08
Fix HasCrew being sticky in some contracts

Teykn:
Addresses issue in #1722, which was present in both Human Records and X-Plane Contracts. 

Ideally, a change to CC's code in how it deals with in-flight state changes should be implemented, since that is what's causing erroneous parameter completions (which afaik only seems to affect HasCrew). Using `DATA` to store the vessel fixes this without touching CC's code, but I am not sure why it works.

### DIFF
--- a/GameData/RP-0/Contracts/Human Records/CrewedAltRecord.cfg
+++ b/GameData/RP-0/Contracts/Human Records/CrewedAltRecord.cfg
@@ -54,19 +54,33 @@ CONTRACT_TYPE
 
         crewedTargetAltitudeKM = @crewedTargetAltitude * 0.001
     }
-
-    PARAMETER
-    {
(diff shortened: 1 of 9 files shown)


lpgagnon:
from #1722 
> Example: place satellite in orbit; put crewed plane on runway, recover it right away. Go to satellite via tracking station. As soon as satellite is loaded, the next crewed Altitude and Speed records (if less than satellite's state) immediately complete.

Teykn:
From when I was testing station, I noticed that when I entered space or reached orbit in flight with an uncrewed craft, the parameter would complete

Capkirk123:
Any progress on this?

siimav:
Superseded by #1869

PR #1867, opened 2023-01-11
KF 2 Rp1 configs

xFluffyDemon:
Rp1 configs to go along with the RO config, again more balance is needed most likely, but everything seems to work fine


siimav:
The cost and tech values look a bit low effort. At least try to divide them between tech nodes by size. There should also be at least some scaling for cost based on that.

Also if `entry_cost_mods` is assigned then `entry_cost` is not used and just becomes noise. Most parts just assign a value of 0 in such cases.

Capkirk123:
Any progress on this?

PR #1869, opened 2023-01-14
Fix crewed records getting completed when switching to an uncrewed vessel

siimav:
Rework all crewed record contracts such that the parameters are wrapped in a VPG.

Depends on [CC v2.2.2.0](https://github.com/KSP-RO/ContractConfigurator/releases/tag/v2.2.2.0) or newer.


siimav:
Fixes #1722

liamdennehy:
Thank You! No more undeserved achievements, this was happening way to much :)

PR #1877, opened 2023-01-27
Unlock ROE-Viking5 part at the same time as the Viking-5C engine config

arrowmaster:


arrowmaster:
Part POE-Viking5 is currently unlocked in 1965. Two engine configs are used by the part, Viking-5C and Viking-6. Viking-5C is unlocked in 1962 and Viking-6 is unlocked in 1965.

The Viking engines are locked in RP-1 far earlier than their IRL flight dates, ruling out using launch dates to determine when to unlock them.

The engine part should unlock at the same time as the first engine config. The two options to correct this is to either unlock ROE-Viking5 in 1962 or change the Viking-5C engine config to unlock in 1965. If doing the later then moving the Viking-6 engine config to later than 1965 needs to be considered. The first option is implemented in this PR as it is simply the easiest. If it is decided to reject this I can do a PR for the second option.

Capkirk123:
Looks good to me

PR #1879, opened 2023-02-03
Hide default engine configs

arrowmaster:
New ROTank parts were added to the start node instead of unlockParts and have been moved.

Hide engine configs from the tech tree if they are the default config for an engine part from an RO mod. An inconsistency with the Viking-5C is addressed in another open PR. The RD-170 and RD-191 still have some inconsistencies. 

The rest appeared from running the Part Browser. I guess it wasn't properly synced in before?

### DIFF
--- a/GameData/RP-0/Tree/TREE-Engines.cfg
+++ b/GameData/RP-0/Tree/TREE-Engines.cfg
@@ -221,6 +221,7 @@
             {
                 %techRequired = orbitalRocketry1986
                 %cost = 100
+                *@PARTUPGRADE[RFUpgrade_AJ10-118K]/deleteme -= 1
             }
 
             @CONFIG[AJ10-133-LH]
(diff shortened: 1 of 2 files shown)


Capkirk123:
I would discourage untagging upgrades, just because we have parts that have different default configs than the global setting doesn't mean all installs will have that part (non-express installs are still nominally supported). Anything that isn't the first config should retain its tag.
Try to keep PRs limited to just one subject for the sake of readability. Separate the ROTanks changes and the engine config changes if possible.

arrowmaster:
The first commit has been split out into #1885. The third commit has already been corrected on master. I've rebased to only have the commit that changes the engine upgrade flag from true to false starting configs.

I'm going to review the changes I made again as there are a few strange circumstances such as the Viking engines were there is only a single engineType Viking but multiple parts that are pulling specific configs from that engineType.

Capkirk123:
This is just removing upgrade tags from configs that shouldn't've had them in the first place, should be good to merge.

PR #1887, opened 2023-02-20
Allow early experiments to run when shrouded

arrowmaster:
Early suborbital rockets had cutouts for their cameras and other sensors. Lets just pretend we can also add cutouts to fairings for the early experiments. This would open up more possibilities for suborbital rocket design beyond either having a 1.25m section or have fairings that need to be jettisoned in the atmosphere before they can start running experiments.


siimav:
After some consults I think the consensus is that it's fine to remove the check from thermometer experiment but keep it on film return.

marsh1832:
Is there a game balance or anti cheese reason for doing so? It seems pretty reasonable for the film camera to have a cutout or periscope attachment

siimav:
Re-added the check to early film experiment.

PR #1888, opened 2023-02-21
New Science Parts from BDB

Capkirk123:
Replace science parts using BDB models.
New parts include:
CR02 - Pioneer 10 Cosmic Ray Telescope
IR01 - Mariner 2 Radiometer
IR02 - Nimbus-ITHIR Scanning Infrared Radiometer
IR03 - Mariner 10 Electrostatic Science & IR Radiometer
IS01 - Nimbus-SIS Satellite Infrared Spectrometer
IS02 - Mariner 10 Extreme Ultraviolet Spectrometer
MG01 - Transit 5E Extending Magnetometer Experiment
MG02 - Pioneer 10 Helium Vector Magnetometer
MM01 - Mariner 3/4 Cosmic Dust Detector
MS02 - Mariner 3-5 Solar Plasma Probe
PH02 - KH-4 Corona Camera System
PH03 - KH-9 Hexagon Reconnaissance System
RPWS01 - Pioneer 6 Top Antenna
RPWS02 - Helios RPWS Antenna
VI01 - Ranger Block 2 Television Camera
VI02 - Mariner 3/4 TV Camera Scan Platform
VI03 - Mariner 10 Television Camera System

This is slightly save-breaking due to some parts having their models replaced with ones that are of a significantly different size/shape.


arrowmaster:
As much as I love trimming out old parts, shouldn't the current parts keep RP0conf = True and just have deprecated = True set in their cfg files to keep them from being trimmed off of in flight crafts on Express Install.

Capkirk123:
Should I just remove the old parts and commit to PLC instead? It would certainly simplify things.

Capkirk123:
Retargeted to PLC, and did a straight replacement of old parts rather than deprecation. Might still need some cleanup of old textures.

PR #1892, opened 2023-03-05
Tweak air intake costs

lpgagnon:
Mostly because having all starting intakes cost 20 regardless of area was weird

miniIntake: less area than pitot intake, lighter; match it for lowest cost, 5
LRadialAirIntake: double mini's area, so 10
IntakeRadialLong: 15 to be closer to the area/cost curve of other supersonic intakes

Full analysis for the skeptical:

Starting with a look at intakes with a cost other than 20, which presumably have had more thought go into them:

RO-IntakeRadialLongPitot: .25, 86kg, 5
IntakeRadialLong: .25, supersonic, 114kg, 10
RO-shockCOneIntakeMig: .35, supersonic, 66kg, 24
RO-shockConeIntake125: .75, supersonic, 137kg, 35
ramAirIntake: .75, supersonic, 47kg, 40
shockConeIntake: 2.2, supersonic, 400kg, 60

All the supersonics with non-20 costs have a reasonable progression, except IntakeRadialLong seems too cheap. Bump to 15.

Pitot is the only non-20 starting/non-supersonic part; doesn't give a lot of guidance, but it's a start

Then look at those costing 20 and see which ones need tweaking

miniIntake: .16, 5kg, 20   -> 5 (less area than pitot but lighter)
LRadialAirIntake: 0.3 10kg 20 -> 10 (double mini's area)
airScoop: 0.45 20kg 20 -> ok (not quite 4x mini, but highest area radial in starter parts commands premium)
CircularIntake: .63, 22kg, 20 -> ok (4x mini, double LRadial, works)
SXTinlineAirIntake: 0.6 11kg 20 -> ok (ditto)
SXTradialAirIntakeShockCone: 0.3, supersonic, 13gk, 20 -> ok (fits in the supersonic progression)

So just nudging 3 parts gives something vaguely reasonable without creating new outliers


siimav:
The subsonic intake tweaks look good but I think it makes sense that the supersonic ones scale up in cost beyond just the area.

lpgagnon:
Updated as discussed: both adjustable intakes bumped up in cost to be more expensive than equal-area fixed cones.

IntakeRadialLong: 22; there's no 0.25 fixed cone, so based on those we have (20 for 0.3, 24 for 0.5)
ramAirIntake: 50, a ~50% premium over RO-shockConeIntake125 (same area, fixed cone)

Discussion brought up the possibility that some of the existing shock cones should actually be using a different intake preset (DSI), which would possibly mean lowering their cost; that can be a separate PR, they'd all be parts unimpacted here.

PR #1901, opened 2023-03-18
Remove unused Offset cone tooling

StonesmileGit:
The offset cone in ProcParts was merged into the normal smooth cone, so the tooling for that version isn't needed now.

(See https://github.com/KSP-RO/ProceduralParts/pull/328)


siimav:
Needs to wait until the ProcParts change gets released.

StonesmileGit:
It does not. This was added as a temp solution during testing, but since it was never used on anything released, it could be merged now

PR #1906, opened 2023-03-31
Remove TestLite from loading screen tips

Test-Account-KSP:


siimav:
I wonder if @ec429 has anything to say about it 🤔

ec429:
Is there any specific reason to un-recommend TestLite?  Any missing feature that's now essential for RP-1?  Have to admit I haven't been keeping up with the discourse lately, so maybe I missed something.

siimav:
I think you haven't officially released the missing features that you were working on not too long ago?

Test-Account-KSP:
It's not in the express install so I thought it's unsupported. If it's still supported I will close the PR.

ec429:
@siimav yeah I'm terrible about that.  Wanted to test it all before releasing but didn't have time to play much so there might be some lurking badness.  I've cut a release anyway because I'm not likely to have much KSP-ing time in the _next_ few months either.

siimav:
In that case I'm fine with the loading screen tip staying since the people preferring TL can now manually switch over if they'd like.

jwvanderbeck:
" yeah I'm terrible about that."
Aren't we all?

- John Vanderbeck
- http://www.johnvanderbeck.com

NovemberOrWhatever:
I'll also note that the Programs and Launch Complexes branch has some heavily rewritten, cleaned up, and better organized loading screen tips. We could just not deal with this until PLC releases, or try to backport some of the tip changes. More an additional note than a suggestion though.

PR #1909, opened 2023-04-02
X-15 altitude limits

lpgagnon:
125km to begin with, roughly matching history.
1Gm (because the partmodule has no off switch) in the mercury node, to more or less represent the extra r&d assumed to be necessary to "orbit-rate" a x-15.

Would be nicer if the partmodule did support getting turned off; as is, the in-game info will still say "unpressurized cockpit, crew will die above 1Gm" which is pretty silly.

extra notes:
- x-15 has no 'human-rated' tag, which maybe makes orbital-x15 too much cheaper than mercury; but adding it would also punish "real" x-15.
- x-15 should maybe also get the no-reentry-rated tag to discourage cockpit+shield contraptions; but it'd be a placebo (tested long ago: even with the tag, the shield was enough to let the cockpit survive, probably just thanks to the extra drag)

### DIFF
--- a/GameData/RP-0/Tree/CockpitUpgrades.cfg
+++ b/GameData/RP-0/Tree/CockpitUpgrades.cfg
@@ -10,10 +10,29 @@ PARTUPGRADE
 	description = Upgrades the X-1 cockpit to X-2 specification. Service ceiling is increased to 75km.\nAfter paying the entry cost of this upgrade, all freshly built and KCT-edited vessels will acquire it automatically.
 }
 
+PARTUPGRADE
+{
+	name = X15CockpitUpgrade
+	partIcon = Mark1Cockpit
+	// mercury and vostok's node: all the prereqs make sense (life support, flight control, heat shielding,
+	// material science) plus "whatever extra is needed to build an orbit-capable spacecraft".
+	// Could instead add an extra node between hypersonic flight and prototype spaceplanes,
+	// but the idea isn't to encourage orbital X-15 as a viable alternate path; it's to acknowledge
+	// that after unlocking the capsule node AND the x-15 node, an orbital x-15 cockpit won't hurt
+	// game balance that much.
+	techRequired = basicCapsules
+	entryCost = 40000
+	cost = 0
+	title = X-15 cockpit upgrade
+	manufacturer = Generic
+	description = Upgrades the X-15 cockpit for space operation. Service ceiling is (mostly) removed.\nAfter paying the entry cost of this upgrade, all freshly built and KCT-edited vessels will acquire it automatically.
+}
+
 @PART[RP0Nose-Cockpit|X1_Crew|RO-X1Cockpit]:FOR[zRP-0]
 {
 	@MODULE[ModuleUnpressurizedCockpit],*
 	{
+		// X-1A altitude record 27km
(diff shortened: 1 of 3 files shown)


NathanKell:
What if the upgrade were conditioned on human-rated EDL instead? I kinda don't want to block orbit-rated X-15 behind the very same node that you need for capsules...

lpgagnon:
current total capsule-node cost: 125
current total x15 r&d cost: 36
+EDL: 51 (crewed-orbit contract needs it anyway)
+HEDL: 66 -> barely an extra speed bump

My plan B was to put it in the matsci node, total cost 55+edl=70, basically same as HEDL. Either way yields an orbital x15 that doesn't need life support or flight control nodes, which is even more bizarre than x15 not needing those nodes in the first place.

The "good" plan is to add a even-more-prototype-spaceplanes node with orbital x15, and move the spaceplane wings there (or maybe a new tier of wings, but ew); but that's going a bit far in the direction of "we approve of orbital x15"

NathanKell:
IMO the problem here is that there's a huge difference between a 180km suborbital, and an orbital flight. I would not want to prevent doing suborbital flights with the X-15.
Adding non-reentry-rated is probably the solution here?

lpgagnon:
> there's a huge difference between a 180km suborbital, and an orbital flight. I would not want to prevent doing suborbital flights with the X-15.

There's also a pretty big difference between the historical 108km and 180km in terms of reentry challenges; it's not clear to me that preventing the latter is much of a loss.

But if we want to go in that direction, maybe we can play with the scrubber instead of altitude limits? Give it a 20-minute scrubber or such, possibly with later upgrade if we really really want to give it orbit access. Wouldn't make a whole lot of sense realism-wise, but I don't see a path that doesn't involve some level of arbitrariness.


[Amusingly, the current 'good enough for orbit, not as good as Mercury' scrubber was my doing.](https://github.com/KSP-RO/ROKerbalism/pull/27)


> Adding non-reentry-rated is probably the solution here?

quoth the PR:
> x-15 should maybe also get the no-reentry-rated tag to discourage cockpit+shield contraptions; but it'd be a placebo (tested long ago: even with the tag, the shield was enough to let the cockpit survive, probably just thanks to the extra drag)

Even if it did work, there's always the 'add a big braking stage' strategy (~4km/s lets a bare cockpit survive, I think).

In any case, it's the "why does mercury need so much r&d and x15 needs so little for the same end result" problem I'd like to see solved.

pap1723:
Why not just create a new part?? Create an Orbital Rated X-15 part that unlocks later.

NathanKell:
The core of the question is how to _non_ orbital rate the X-15, no?

lpgagnon:
New part or partupgrade doesn't really make much difference; either way, the main point of contention is _where_ to put the new unlock.

Just _what_ the difference should be between the two seems comparatively unimportant. Temp rating seems obvious but doesn't really work; altitude or scrubber limits would both feel arbitrary but would both work fine

NathanKell:
For now I've added the non-reentry-rated tags. We'll see if that's enough, since the other options are more involved.

lpgagnon:
does not appear to be enough https://discord.com/channels/319857228905447436/845456949331623985/1192424856701124618

PR #1913, opened 2023-04-16
Make Leaders Unlock Gradually

NovemberOrWhatever:
Chief designers unlock at Karman line,
Flight Directors at crewed 50 km,
Scientists at First sci sat
And the first batch of contractors and subcontractors unlocks at Post-War MatSci, while the rest unlock at Human MatSci. The choice of which node to unlock at was based on what each one offered in terms of pros and cons, and there isn't a perfect match with actual dates or even between main and sub contractors. Still, it should be a decent start. Dummy leaders will also appear to tell you when the next thing will unlock


NovemberOrWhatever:
That is way too many commits. I think I did something wrong, though I'm not sure what exactly

PR #1919, opened 2023-04-26
Add descriptions to ОКБs

NovemberOrWhatever:
And change OKB to ОКБ because I tested and it renders fine


Test-Account-KSP:
One day Chinese leaders and contractors will be added like that
王希季

PR #1927, opened 2023-05-02
XLR35-RM-1 subconfig price reduction

VeselyiKot:
Any reason why XLR35-RM-1 XLR11 subconfig should add +200f to the engine price, which already is quite high, sitting at 300 thanks to a human-rated, spaceplane natue of the original config? The combined cost is, like, **more than even RD108 costs**, for a second. If anything, it should be exactly the opposite: disposable variant of the engine, speculatively optimized for mass production.
Looking at similar-ish engines in game RN, like RD-200 or Gamma 4, pricepoint of ~200 cr for a disposable four-chambered gimballed turbopump motor looks plausible. No historical sources exist on real unit price AFAIK, since historically it was a one-off prototype, and they wouldn't account for speculative serial production anyway, so balance takes the point.
And from the point of balance, 500cr disposable motor with less than mediocre parameters looks useless. Completely and utterly. Meanwhile, 200cr engine could be useful in at least some theoretical applications, like building a rocket for downrange milestone?

### DIFF
--- a/GameData/RP-0/Tree/TREE-Engines.cfg
+++ b/GameData/RP-0/Tree/TREE-Engines.cfg
@@ -5096,7 +5096,7 @@
             @CONFIG[XLR35-RM-1]
             {
                 %techRequired = earlyRocketry
-                %cost = -100
+                %cost = 100
                 %description = Modified XLR11 used on RTV-A-2 Hiroc missile in Project MX-774. Higher performance, single use. Includes gimbal assembly.
                 *@PARTUPGRADE[RFUpgrade_XLR35-RM-1]/deleteme -= 1
             }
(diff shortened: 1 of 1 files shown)


lpgagnon:
the xlr11 lacks a turbopump multiplier because its first config is pressure-fed; the cost of other configs is inflated to compensate

VeselyiKot:
> s

Understood. Then it could cost, like, +75 to +100, less or equal than RM-5? I mean, when compared to RM-5 it is +100f on top, lacks the ability to throttle, ability to restart, ability to be reused, burn time and even +0.05t of mass. All that for gimbal and a little bit of extra throttle.
Gimbal is worth nothing in credits, looking at NK nogimbal configs, and peformance increase is only achieved due to the worktime sacrification and single-use nature of an engine. Also worth assuming that the assembly would be produced on scale thanks to the single-use, missile nature, if it was real enough to justify use in rocket program.

Also looking at S2.253 pumpfed conversion price, S3.42T adds only 75 exta cost on top. So it results in 225kr vs, again, 500kr, which looks stange.

PR #1935, opened 2023-05-13
First draft of spaceplane program and contracts

ec429:
I've tested that the program and contracts are present in-game but that's about all.  And I have no idea how to balance things for NotBudgets so I just took the capsule equivalent contract rewards and added 20%.

Note that the way I've done this, the program isn't an alternative to `CrewedOrbit`, because it requires completing the orbital re-entry contract.  So instead it's either a follow-on or a parallel development.  And I've blocked taking both the capsule and spaceplane version of a given contract simultaneously, so you can't get paid twice for one mission, but there's no reason you can't use your spaceplane to complete `CrewedOrbit` contracts if it's ready in time.

### DIFF
--- a/GameData/RP-0/Contracts/Groups.cfg
+++ b/GameData/RP-0/Contracts/Groups.cfg
@@ -113,10 +113,18 @@ CONTRACT_GROUP
 	}
 	CONTRACT_GROUP
 	{
+		name = EarlySpacePlanes
+		displayName = Spaceplane Development
+		minVersion = 1.12.2
+		sortKey = 23
+		agent = Exploration
+	}
+	CONTRACT_GROUP
+	{
 		name = CrewedLunar
 		displayName = Crewed Lunar Exploration
 		minVersion = 1.12.2
-		sortKey = 23
+		sortKey = 24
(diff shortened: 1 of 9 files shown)


NathanKell:
This is pretty cool. Merging for now, though I think it's underfunded (probably needs total funding of about 700-750k to be worthwhile). Also probably need to make Crewed Lunar and Stations (if we have a stations) accept this _or_ crewed orbit as a prereq.

NathanKell:
Ugh I misread the above about it being parallel. In that case it maybe needs _lower_ funding, since otherwise you probably want to just pick both crewed orbit and this.

ec429:
Rough consensus from follow-up discussion on Discord:
1. make mk1 cockpit single seat (if possible add a KCT-recognised tag for "must be only crewable part on vehicle" to prevent stacking) until upgraded by effectiveSpaceplanes
1. three-seater contract becomes optional (so '2 crew + manœuvres' is the capstone), with funding and duration for the program reduced accordingly
1. add optional objective on existing station crew rotation contracts to increase payout if runway landing

Issue #1964, opened 2023-05-31
Can no longer add oxygen to service modules for some reason


smunstu:
In the middle of a campaign, something occurred that has rendered me unable to add oxygen to my service modules.  I can still add water and food.

I do have Crew Survivability researched.

I believe this occurred after CKAN pushed some mod updates to my install. I did attempt to reinstall the entire mod pack in case it was caused by an improper installation.

Poking around in my Persistent.SFS file, it looks as though that ability has been transferred to the research node, 'standardDockingPorts'.

Help!!!

TwistedGiraffe:
Are you on both the latest release of RO and RP-1? Did this save start earlier than those releases? 
Please post your ModuleManager.ConfigCache (or at least the relevant lines).

smunstu:
Yes I am on the latest version to my knowledge. No updates come up when I refresh CKAN. To give this a date, I believe it started occurring almost two days ago. This campaign began some weeks before this.

File uploaded because it is too large to paste directly.

I deeply appreciate the prompt assistance.

https://drive.google.com/file/d/1-3l8sOMwcq50DBsMCBpzyHYz7YyTgqkj/view?usp=sharing

TwistedGiraffe:
Speaking of prompt, you can try https://github.com/KSP-RO/RealismOverhaul/pull/2857 and https://github.com/KSP-RO/RP-0/pull/1965. These two PRs should solve your problem.

smunstu:
Just to clarify, I should manually make the changes to the files indicated in files changed tab, correct?

TwistedGiraffe:
Yes. Though if you have GitHub Desktop, you can fetch the PRs and copy the changed files, so you don't have to do it manually.

smunstu:
Thanks so much! I was quite worried I might have to pause this campaign for a while

NathanKell:
Use the official release of the hot fix. The pull request was against
master and therefore had a ton of save breaking changes.

bofhgr:
[...]

PR #1965, opened 2023-06-01
Explicitly distinguish HP and SM tank on oxygen gas unlock tech

TwistedGiraffe:
So that patching order is not an issue. Uses the tags from Realism Overhaul, before they are removed by zzzTagCleanup.
Requires https://github.com/KSP-RO/RealismOverhaul/pull/2857 and should solve #1964. 
Thank @Capkirk123 and @arrowmaster for the idea.


NathanKell:
This PR isn't actually needed, AFAICT. The RO PR alone should be sufficient.  The TACLS fix is also unneeded (though we should delete that case regardless) because RP-0 requires Kerbalism and thus conflicts TACLS.

PR #1969, opened 2023-06-03
Contract completion checking for CC_RP0

ryanc55:
- Add RP1ContractTracker scenario module. Tracks contract completions by KCTVesselID.  Contract tag used if available, otherwise name.
- Add HasCompleted vessel parameter. Checks RP1ContractTracker for contracts already completed by KCTVesselID, either current or in tracked/unloaded VPG.  Invertible.  Multiple  contract tags allowed, results ORed.
- Sounding Rocket contracts reworked to remove contract exclusion requirements and add completion exclusions using HasCompleted.


ryanc55:
Since the plan is to remove penalties for canceling contracts, this is no longer needed.

NathanKell:
This is still useful for e.g. stations, no?

PR #1989, opened 2023-06-08
Add "science-based" KCT resource cost mults

Capkirk123:
Add NFPA-based resource cost mults instead of whatever we had before.

Multipliers are calculated by summing the numbers on the NFPA fire triangle + any special symbols, adding +1 for cryo, +2 for deep cryo, and +2 for subcooled, and then performing some math to get an exponential curve.

Some of these multipliers are much larger than they were before, but propellant cost is a relatively small fraction of vessel cost, so the difference isn't as large as it may seem. Although cost in VAB is not effected, rollout cost is. Based on some experimentation, an R-7 stuffed full of Pentaborane (the highest multiplier resource) experiences twice the rollout cost and time, which seems reasonable?


NathanKell:
I want to try an experiment where instead of applying to a part's BP, these multipliers just multiply the resource cost itself. You might need a rather larger exponential function though for things to work out balance-wise.

Capkirk123:
This doesn't have a huge effect on balance and can be easily changed later, so merging

PR #1997, opened 2023-06-10
Add simplified Chinese translation

mbnull:
Chinses localization.
Not fully translated, as some usages in Chinese may require specific contexts.For example "Can Remove In/On /-". It might be same in Chinese,but diff in eng.


mbnull (commented):
Add simplified Chinese translation

PR #2000, opened 2023-06-11
Rework the altitude SR contracts

siimav:
* Remove 10% base variation
* Fix payload amount to 75 units, for all levels
* Reduce the altitude increase per level from ~100km to 60km
* Cap levels to 15
* Completing the difficult variant will make the next intermediate have the exact same parameters


NathanKell:
I like the fixed payload, it makes smaller-than-V-2 rockets way more viable 👍

PR #2013, opened 2023-06-15
Reduce upgrade cost of admin building

NovemberOrWhatever:
I also reduced the cost of the level one astronaut complex which will halve the maintenance cost from ~7200/year to ~3600/year, which is more in line with other buildings. As this is the level 1 version, it won't effect upgrade prices.


siimav:
Sounds good. I noticed the same discrepancy as well.
Nauts themselves already have a high maintenance cost so doesn't make sense for the facility to have as well. Especially considering that :noplanes: careers do not touch crew until like 10 years in.

PR #2019, opened 2023-06-15
Update DistanceSoundingDifficult.cfg

ppboyle:
This SR distance contract needed to be patched as well.


siimav:
Merged this PR with #2018

PR #2026, opened 2023-06-17
Reduce cost of first admin building upgrade

NovemberOrWhatever:
I also reduced the cost of the level one astronaut complex which will halve the maintenance cost from ~7200/year to ~3600/year, which is more in line with other buildings. As this is the level 1 version, it won't effect upgrade prices.


siimav:
I changed AC upkeep separately. No need for that to wait until we decide what to do with AB upgrade cost.

NathanKell:
Closing this in favor of #2074

PR #2030, opened 2023-06-17
Rebalance RP-1 contractors and subcontractors

NovemberOrWhatever:
Make locked leader disappear properly for subcontractors
Buff a couple contractors slightly
Buff a lot of subcontractors

More to do but this should be a good start


NathanKell:
The icon paths are pointing to RP-0. Please make sure you actually have the updated files, a rename alone isn't enough. I think your best bet is to copy your work out, do a reset hard to origin/master, then create a fresh branch and PR and put your changes in there only on the lines you want to change.

PR #2062, opened 2023-06-24
fix eye-piercing red texts in contract details and crewed earth orbit contract descriptions

byt3c:
Fix description in contract to enable triggering success criteria.


TwistedGiraffe:
Instead of explicitly requiring a part, how about requiring the science a la #1942?

byt3c:
it might be a nice twist to get people to place actual sensor on a craft :) i have nothing against requiring a device present as long as it is in the description. It gives a slight challenge to place items correctly :)

siimav:
Looks like you added some unrelated changes to this PR now.

byt3c:
welp - i think i pushed manned orbit contract description fix to the same branch. Now there are two fixes :)

siimav:
Probably not the best wording on those two crewed orbital flight contracts. @NovemberOrWhatever thoughts?

NovemberOrWhatever:
Once you've completed this contract, Orbital Flight (One Crew) contracts will become unavailable. If you have both this and the Orbital Flight (One Crew) contract accepted, completing this will cause that contract to fail.

Is there any penalty for failing though? Or was it just decline that we changed?

siimav:
> Is there any penalty for failing though? Or was it just decline that we changed?

Nope, both are 0 now.

byt3c:
I received a lot of negative rep (like failing contract not cancelling it) when reporting - not sure how it is now.

PR #2064, opened 2023-06-25
Fix WarpToFundTarget with high incomes

FioraAeterna:
The algorithm used to detect a fund target failed when daily income became relatively large, which caused it to error out towards the mid to late game.

With the new algorithm, it will warp to the soonest time that it determines the listed amount of money will be available, even if it's very slightly above the desired fund target.


NathanKell:
Thanks!! :)

PR #2069, opened 2023-06-26
Film Rework (Do not merge yet)

zephyrspace:
A rework to the way film science is done to make it less of a pain to run + make it better reflect the reality of early satellite technology.

* Photo-0 (incomplete):
  * An experiment in the style of the ones run on V2s and Aerobees, granting science equivalent to 1.5 biomes of the current photo-1, but not biome specific, for 12 minutes of total runtime
  * The part would be a suborbital 35mm film camera, reinforced to withstand impacts as hard as 100m/s, and would have a runtime measured in the 30 second mark, this would mean you could run a single camera on an aerobee for a possible 24 launches with the intent that you get less science total, than if you run a ring of 8 of them in the historic V2 config.

* Photo-1 (Complete): 
  * A KH-1 style experiement in satellite era science, very similar to current photo 1, but only operable in space low. biome specific to soft incentivise polar orbits, with 4 science per biome and an operating run time of 2hrs per biome leading to needing 20hrs of orbital lifetime total. 
  * The KH-1 part would have 20kgs of onboard film which would complete photo 1 in 4 launches, as ~10 science per launch fits way better with the expected return on early sats than the current 40, making them less lucrative than lunar impactors but also for less funding requirements. 
  * EC requirement should drain agena b (70000 units of ec but nondeepspace for game balance) in a day 
  * Exact run time numbers will need to be changed based on minimum biome size and how much of a pain it is to hit the smallest ones

* Photo-2 (KH-4 and Zenit Complete, no KH-4a/b yet): 
  * A KH-4 style experiment in early human spaceflight era science, that follows on from photo-1 in that it stays biome specific, but extends the time needed in each biome to 4 days and the science per biome to 16. For a total of 40 days of on orbit lifetime and 160 science over the course of the experiment
  * The KH-4 part would carry 1/8th the sample at 40kg, needing 8 launches total for 20 science per launch and would drain the (deep space) agena of ec in 5 days.
  * We could then add the KH-4A in a close follow on node (like how improved avionics is set up), and it and zenit can could carry 1/4th (at 80kg) the sample to cut down on the launches and reflect the longer on orbit lifetime.
  * Also unless i'm missing something there's no way to limit sample capacity by mass is there, otherwise we could put a limit on the amount of film carried by the return capsule


NathanKell:
@zephyrspace please reopen when you're able to work on this again.

PR #2074, opened 2023-06-27
Implement major/minor program support.

NathanKell:
* Programs now have a 'slots' field. *Admin Building now has a slot limit rather than program limit
* Apply slots field to all existing programs
* Add upgrade script to handle transition from pre-2.4 Admin.


norcalplanner:
Love it!  Thanks for implementing it in an even better format than I was thinking.  One suggestion - space stations should probably also be 4 slots, based purely on the funding provided (15 million over 10 years with no curve, compared to 18 million over 9 years with a backloaded curve for crewed lunar).

NathanKell:
Hmm, think it depends on what form that program takes. but fair point!

PR #2080, opened 2023-06-29
Log individual maintenance components & many more

siimav:
Somewhat tested this.


NathanKell:
LGTM except the log line about totalupkeep is busted now (you need to divide by timeFactor, since you multiplied it into totalUpkeep) but I think we can just kill that log line and its associated error checking since there hasn't been any reported errors around maintenance and subsidy since I, err, Finally Fixed It (tm).

PR #2111, opened 2023-07-15
Allow Flexibility in OrbitalFlight2 Contract

Teykn:
Allows the player to start in either orbit 1 or orbit 2, rather than just orbit 1, in the "Orbital Flight with Maneuvers and 2+ Crew" contract


NathanKell:
@Test-Account-KSP

Issue #2106, opened 2023-07-14
Orbital Test Flight (uncrewed) contract is bugged


Test-Account-KSP:
Visiting another (uncrewed) spacecraft when the contract is accepted results in red "failed" checkmark for the condition "Have space for at least 1 crewmember on board".  This doesn't go away during the actual mission with an empty crewed capsule.

![image](https://github.com/KSP-RO/RP-1/assets/110846612/cefb6257-0bab-4ad6-bd55-f1827209a11c)

After cancelling and reaccepting the contract, there is no issue in the sim

![image](https://github.com/KSP-RO/RP-1/assets/110846612/608652b1-29e2-4b3e-a4fa-3940cb86f470)

PR #2113, opened 2023-07-16
Unstick HasCrewCapacity in contracts

Teykn:
Addresses #2106. Added `disableOnStateChange = false` to `HasCrewCapacity` parameters since it appears that once failed, they cannot return to being uncompleted.

### DIFF
--- a/GameData/RP-1/Contracts/Earth Crewed/OrbitalTestFlight.cfg
+++ b/GameData/RP-1/Contracts/Earth Crewed/OrbitalTestFlight.cfg
@@ -83,6 +83,7 @@ CONTRACT_TYPE
 			minCapacity = 1
 			title = Have space for at least 1 crewmember on board
 			hideChildren = true
+   			disableOnStateChange = false
 		}
 		
 		PARAMETER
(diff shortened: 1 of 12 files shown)


siimav:
Interesting. Not entirely sure why this helps but I assume you verified it experimentally?
There certainly is this ominous message in CC source:
https://github.com/KSP-RO/ContractConfigurator/blob/36fd07845b25ff20421245f1fccc86cb9bfcfaaf/source/ContractConfigurator/Parameter/VesselParameter/HasCrewCapacity.cs#L27-L29

Teykn:
Verified it with the contract that testaccount was experiencing the issue with. 

> Not entirely sure why this helps
My understanding is that, based on testaccount's issue, that hasCrewCapacity implicitly has `disableOnStateChange = true`, so if it fails, it'll stay that way (Ominous message in CC source also seems to align with this). Changing `true` to `false` ensures that it doesn't get stuck on "failed", and instead it'll only be between "incomplete" and "completed". This also shouldn't impact vessels with crew capacity, since the only case where the crew capacity would change is when the crew part is destroyed or undocked

PR #2123, opened 2023-07-18
Add RP-1 cost multiplier for new fairing side petal option

ryanc55:
Add RP-1 cost multiplier for new fairing side petal option


siimav:
For https://github.com/KSP-RO/ProceduralFairings/pull/47

PR #2124, opened 2023-07-18
Reduce astro3 cost and crewed lunar payout

NovemberOrWhatever:
It'd actually still have a bit of an impact on balance in the form of maintenance costs, build time, and the fact that program payout depends on difficulty, while construction cost does not, but this could at least help a bit if too many construction costs is an issue.

### DIFF
--- a/GameData/RP-1/CustomBarnKit.cfg
+++ b/GameData/RP-1/CustomBarnKit.cfg
@@ -47,7 +47,7 @@
 	}
 	@ASTRONAUTS
 	{
-		@upgrades = 25000, 500000, 1500000
+		@upgrades = 25000, 500000, 1000000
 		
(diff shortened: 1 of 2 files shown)


NathanKell:
Is this still needed after the construction time change?

NathanKell:
(And that our funds balance is rather too generous atm--between the unlock credit changes, the rep->subsidy change, and the increases in rep, funds are waaaay more generous than they were)

NovemberOrWhatever:
I don't know, but I figured I can just do it now while I remember it and we can either merge it or close it depending on if that's the case. But I'd say more likely than not that it'd still be useful, especially when people are also complaining about the portion of their budget that goes to construction.

NathanKell:
Decreased program funding by 2m instead of 500k

PR #2131, opened 2023-07-21
First Crewed Orbital Flight: Update contract description for P&LC

Amselfass:
The warning about hard deadlines and steep failure penalties is no longer required.


NathanKell:
Thanks!

PR #2134, opened 2023-07-22
Fix sounding and satellite payload rounding errors issue

### LINKED ISSUE #2133: Sounding and satellite payload rounding errors
Sounding rocket contracts, 4500+ km downrange contracts and many satellite contracts ask for a fixed amount of sounding/comsat/navsat/weather sat payload.  However, there is rounding in the game that can result in 499.99 shown as 500 and the contract not completing, even though the player was sure that the rocket has enough payload.

The proposal is make all these contracts check for (contract amount - 0.1) payload.  So for example, a sounding rocket contract would ask for 75 So, but in fact check that the rocket has 74.9 So.

Test-Account-KSP:
Trying to fix #2133:
"Sounding rocket contracts, 4500+ km downrange contracts and many satellite contracts ask for a fixed amount of sounding/comsat/navsat/weather sat payload. However, there is rounding in the game that can result in 499.99 shown as 500 and the contract not completing, even though the player was sure that the rocket has enough payload."

All the contract now have a margin of 0.01, which is apparently not enough. This PR changes 0.01 to 0.1.

### DIFF
--- a/GameData/RP-1/Contracts/Early Satellites (Heavy)/Downrange6000-Heavy.cfg
+++ b/GameData/RP-1/Contracts/Early Satellites (Heavy)/Downrange6000-Heavy.cfg
@@ -161,7 +161,7 @@ CONTRACT_TYPE
 			name = HasSoundingPayload
 			type = HasResource
 			resource = SoundingPayload
-			minQuantity = @/missionPayload - 0.01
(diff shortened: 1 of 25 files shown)


Capkirk123:
Seems reasonable, seeing as RF generally rounds to one decimal place except with very small numbers.

PR #2136, opened 2023-07-25
Rebalance later tanks

NovemberOrWhatever:
Increase the max utilization of advanced composite tanks to match standard composite tanks
Make Starship tanks cheaper to balance their high weight
Change modern balloon tanks to only a bit more expensive than their refined counterparts (25% instead of 40%) and increase their tooling cost to 7% more than refined balloon tanks. 

Please merge at same time as RO PR


siimav:
Associated RO PR: https://github.com/KSP-RO/RealismOverhaul/pull/2884

PR #2137, opened 2023-07-25
Allow newspaper screenshots to be taken on contract parameter completion

siimav:


NathanKell:
Neat! Probably should apply this to crewed milestones too since it'd be better to have the pic when the plane breaks the sound barrier rather than when it lands?

PR #2139, opened 2023-07-26
Better tracking station descriptions

Capkirk123:
I don't understand how facility descriptions work.

This should replace the existing TS descriptions with a new hardcoded text block that contains data more relevant to RA comms. Some code work will still be required to actually put these locs in the correct place.


NathanKell:
Merging this and making the code change in master

PR #2154, opened 2023-08-01
Add more localization support

TwistedGiraffe:
This PR currently adds localization support for:
- RP-1 Craft Files (e.g., the Jet Trainer)
- Tech Node names and descriptions
- Leader names, biography, and effects. 
- Milestone newspaper articles. 

Some quality standards I have, which are negotiable:
- All localization tags in RP-1 begin with `#rp1` and then the filename they are in. They follow CamelCase. 
- Split localizations into several files based on their origins. The base `en-us.cfg` should be reserved for general strings. How general I am not sure. 

Future additions include but are not limited to:
- RP-1 Partupgrades
- Science Parts
- Science Experiments and Results
- Programs and Contracts

<s>This PR is totally not me trying to bump my lines of contribution. </s>


siimav:
Did you actually test that craft file loc works?

TwistedGiraffe:
> Did you actually test that craft file loc works?

Yes. That's how Squad localized their stock vessels. See KSP\Ships.

NathanKell:
Thank you so much for taking this on! This is AWESOME.

Couple things:
* I'm fine with switching from #rp0 to #rp1 but please make the change to all loc tokens (and to their references, which may be in .cs files as well as .cfg files), I really don't want half the loc to start #rp0 and the other half to start #rp1
* Please keep the underscores in the tokens. It makes them far more readable IMO.

Also, I'm a bit skeptical of the change to multiple files: what are the advantages here? I tend to like the single file for three reasons:
1. It's always clear where to put the token. You don't have to think about whether a token fits somewhere else or needs a new file.
2. It's very easy to search a single file (and you always know where to look for text, even if you don't know the token name).
3. It's easier to manage for translators (I think)--you only have to check changes to a single file, you can find new lines easily (rather than having to check many files), and only a single translated file needs to be maintained.

TwistedGiraffe:
I was mostly following Kirk's approach to adding localization support in RO. He used multiple files and CamelCase. I agree with your opinions on single vs. multiple files, but we may need to reach a consensus with Kirk about style. 
I intend to leave the `rp0` prefixes until the last commit in this PR, as I don't know how to build RP-1 on my machine.

TwistedGiraffe:
Is there any update on what style I should use? I can open another PR over the weekend with a single file and `lower_with_under` if that's the decision.

NathanKell:
@Capkirk123 can you speak to the multifile approach? My preferences are clear as above, but if you have compelling arguments for that I'm fine with RP-1 going that way too

NathanKell:
(I'm _very_ strongly in favor of using underscores however, as I find it hard to follow loc strings without them)

Capkirk123:
Uh for the multifile stuff? I mostly just did it for engine configs, since that is a very large amount of relatively standardized tags.

As for camelcase, I had no particular reason.

NathanKell:
Ok! Then I'd suggest keeping to things as they are here already, i.e. single file and underscores (although changing #rp0 to #rp1 good catch) -- and we can add the underscores to RO at our leisure.

NathanKell:
@TwistedGiraffe think we're resolved now :)

TwistedGiraffe:
Okay. I will close this PR draft and open a new one with a single file and underscores. Localization of RP-1 needs only a handful of strings from RealismOverhaul, e.g., `#roMfrGeneric`, so I will temporarily put those strings in RP-1's localization file before we roll out the underscores for the RO localization file.

NathanKell:
Pinging here--what's the status of this? Would love to ship something. Thanks!

TwistedGiraffe:
I was busy working on a few other projects recently and could not continue working on this PR. Now the new semester is starting, I will have more free time to look at this.

NathanKell:
Sweet!

PR #2155, opened 2023-08-02
Add some landing contracts

NovemberOrWhatever:
Still need to balance rewards, but all but two are going into disabled programs anyway

### DIFF
--- a/GameData/RP-1/Programs/Programs.cfg
+++ b/GameData/RP-1/Programs/Programs.cfg
@@ -877,6 +877,8 @@ RP0_PROGRAM
 	{
 		orbitCeres_Rep = true
 		orbitVesta_Rep = true
+		landingCeres = true
+		landingVesta = true
 	}
 }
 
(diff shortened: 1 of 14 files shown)


GameData/RP-1/Programs/Programs.cfg
```diff
@@ -928,10 +930,10 @@ RP0_PROGRAM
 	}
 }
 
-disabled_RP0_PROGRAM
+RP0_PROGRAM
 {
 	name = JupiterMoonLandings
-	title = Jovian Moon Landings (No Content Yet)
+	title = Jovian Moon Landings (WIP)
```
NathanKell, line 936: I don't think we should be enabling programs with TODO descriptions? :D
(Also I thought Max gave feedback on the contracts, not the program length & funding etc?)

NovemberOrWhatever: We already have some with TODO descriptions enabled, IIRC, but I can redisable them in a bit (or you can do it yourself if you want to release soon)

NathanKell: Oh jeez we do? I thought that was part of your mass disabling pass before initial release. Whoops! And no, not in a hurry, I need to do the unlock credit curve stuff and fix the bug siimav reported with credit.

PR #2159, opened 2023-08-04
Update RP-1 description on ckan to make it more clear

Clayell:
I'm sorta half-joking with this, (and it's like 2 am) but I just can't believe how many people we get in the server who don't install the express install. If they can't follow these instructions then... I don't know
I did do a quick check and this has less characters than the express install's description so it should fit into 2 lines.

### DIFF
--- a/RP-1.netkan
+++ b/RP-1.netkan
@@ -2,7 +2,7 @@
 	"spec_version" : "v1.26",
 	"identifier" : "RP-1",
 	"name"       : "Realistic Progression One (RP-1)",
-	"abstract"   : "Career for Realism Overhaul. Don't install this directly without the Express Install unless you know what you're doing.",
+	"abstract"   : "Career for Realism Overhaul. Use the Express Install and leave this as it is unless you know what you're doing.",
 	"license"    : "CC-BY-4.0",
 	"author"     : "RP-1 Group",
 	"$kref"      : "#/ckan/github/KSP-RO/RP-1",
--- a/RP-1.netkan
+++ b/RP-1.netkan
@@ -2,7 +2,7 @@
 	"spec_version" : "v1.26",
 	"identifier" : "RP-1",
 	"name"       : "Realistic Progression One (RP-1)",
-	"abstract"   : "Realistic Progression One - Career for Realism Overhaul",
+	"abstract"   : "Realistic Progression One - Career for Realism Overhaul. Do not install this without the Express Install unless you know what you're doing.",
 	"license"    : "CC-BY-4.0",
 	"author"     : "RP-1 Group",
 	"$kref"      : "#/ckan/github/KSP-RO/RP-1",
--- a/RP-1.netkan
+++ b/RP-1.netkan
@@ -2,7 +2,7 @@
 	"spec_version" : "v1.26",
 	"identifier" : "RP-1",
 	"name"       : "Realistic Progression One (RP-1)",
-	"abstract"   : "Realistic Progression One - Career for Realism Overhaul. Do not install this without the Express Install unless you know what you're doing.",
+	"abstract"   : "Career for Realism Overhaul. Don't install this directly without the Express Install unless you know what you're doing.",
 	"license"    : "CC-BY-4.0",
 	"author"     : "RP-1 Group",
 	"$kref"      : "#/ckan/github/KSP-RO/RP-1",
--- a/RP-1.netkan
+++ b/RP-1.netkan
@@ -2,7 +2,7 @@
 	"spec_version" : "v1.26",
 	"identifier" : "RP-1",
 	"name"       : "Realistic Progression One (RP-1)",
-	"abstract"   : "Career for Realism Overhaul. Use the Express Install and leave this as it is unless you know what you're doing.",
+	"abstract"   : "Career for Realism Overhaul. Do not select/install this directly, instead select the RP-1 Express Install and let it install this automatically unless you know what you're doing and can fix it yourself.",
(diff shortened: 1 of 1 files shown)


NovemberOrWhatever:
Looks good, thank you, though the "unless you know what you're doing" part might actually not be needed

Clayell:
> Looks good, thank you, though the "unless you know what you're doing" part might actually not be needed

I did think about that, and as someone who doesn't use the express install anymore I think there are some very tiny merits to avoiding it. (mainly not having to remove everything if I want to change something) I also think that the text would go on to the 2nd description line anyway without that added comment so it's not taking up unneeded space. If you really want to remove it then I won't be bothered.

RP-1.netkan
```diff
@@ -2,7 +2,7 @@
 	"spec_version" : "v1.26",
 	"identifier" : "RP-1",
 	"name"       : "Realistic Progression One (RP-1)",
-	"abstract"   : "Realistic Progression One - Career for Realism Overhaul",
+	"abstract"   : "Realistic Progression One - Career for Realism Overhaul. Do not install this without the Express Install unless you know what you're doing.",
```
NathanKell, line 5: I'd probably say something more like "Don't install this directly, instead install the RP-1 Express Install except in rare circumstances."

NathanKell: And I'd also remove the repeat of RP-1, so it's just "Career for Realism Overhaul. Don't..." because you _do_ indeed want the "Don't" bit to show up on the first line.

Clayell: > I'd probably say something more like "Don't install this directly, instead install the RP-1 Express Install except in rare circumstances."

This is actually a fair bit longer, and I'd argue that "unless you know what you're doing" is a better descriptor than "rare circumstances". I'll put directly in there though.

> And I'd also remove the repeat of RP-1, so it's just "Career for Realism Overhaul. Don't..." because you do indeed want the "Don't" bit to show up on the first line.

Good point, I'll go ahead and do that.

NathanKell:
So, the thing about this is--you _don't_ want to encourage people to check both the express install and this. You want them to check _only_ the express install, and have it select all subsequent mods. Having this and the express install both checked together will lead to bad results.

Clayell:
> So, the thing about this is--you _don't_ want to encourage people to check both the express install and this. You want them to check _only_ the express install, and have it select all subsequent mods. Having this and the express install both checked together will lead to bad results.

Hm...didn't think of that. How about "Use the Express Install and don't install this unless you know what you're doing."?
Or, "Use the Express Install and avoid installing this unless you know what you're doing."

lpgagnon:
I feel the emphasis should be along the lines of "don't manually select this". A "don't install" advice is sure to lead to people trying to _uninstall_ it after the express install installs it, "because ckan told me to"

Clayell:
> I feel the emphasis should be along the lines of "don't manually select this". A "don't install" advice is sure to lead to people trying to _uninstall_ it after the express install installs it, "because ckan told me to"

I see the problem that you're pointing out, and I'm having a hard time thinking of a solution to it. Saying to "not manually select/install it" could be interpreted as "don't install this through the github" or "don't click on this while installing the express install".

Maybe if I just say "Use the Express Install and leave this alone unless you know what you're doing."
Idiot-proofing is surprisingly hard! Perhaps we can't cover every single avenue without writing a full instruction manual in the description.

RP-1.netkan
Clayell, line 1: This seems pretty long, and it kinda goes against my work on making it shorter. It is considerably longer than the express install's description, so it would need to be 3 lines of description instead of two. I'd also argue that it doesn't add that much more info than my version, so I'm slightly annoyed that you went ahead with this.

NathanKell: "Leave this as it is" is not clear language--does it mean leave this checked? Unchecked? Something else? Need to explicitly say "unselect this", to make clear to the player that this mod is not to be directly installed unless they know what they're doing. It also needs to be clarified IMO that the express install _will_ install this, so don't worry about it. Yes, the description is longer (and I'd be happy to have it tighter) but the reason I changed it is because the version in the PR was simply not clear enough for users.

NathanKell: To add, I'm sorry that came across as stepping on toes, that wasn't my intention. I didn't think my edit was that big or controversial a change, and it seemed more straightforward than going round again a few times on wording, but I'm happy to continue that discussion!

Clayell: > "Leave this as it is" is not clear language--does it mean leave this checked? Unchecked? Something else? Need to explicitly say "unselect this", to make clear to the player that this mod is not to be directly installed unless they know what they're doing.

I think "leave this as it is" is quite clear, it pretty much just means don't touch it in as few words as possible. As lpg pointed out, saying "don't install this" could lead to people uninstalling it. (saying that the express will do it automatically may fix this, but its a lot of extra words, and it could be still interpreted as it would be fine to unselect this after the express install is selected)

How about "Select the RP-1 Express Install and do not touch this unless you know what you're doing."
I think "fix it yourself" is already included in "know what you're doing". Also, if I say to "not select it", then people could unselect after the express selects it, and if I say to "not unselect it" then people could select it along with the express install. "do not touch it" takes the best of both worlds by just telling people to leave it alone. I do not believe it can be made perfect, (as at least one person will think of it in an unpredicted way) but in my opinion this is a better solution.

>  I didn't think my edit was that big or controversial a change, and it seemed more straightforward than going round again a few times on wording, but I'm happy to continue that discussion!

I'm always happy to continue discussion! I want to make sure everyone is as happy as possible before we are done here.

PR #2160, opened 2023-08-04
Revise TargetedSats and CommNetwork balancing

siimav:
Give both programs the same per-year payout so there doesn't exist a perverse incentive to pick commnet over targeted sats.
Also, at least with the contracts we have right now in those programs I really can't justify commnet paying more than targetedSats.


NathanKell:
I'm confused, I thought you were changing it so the comm network *program* required completing Target Sat's First Commsat contract. If you make just the contracts require that, then you can lock yourself by taking comm network without taking Targeted Sats, and not be able to have any contracts.

Similarly, GEO Comm Network should require First GEO.

siimav:
Turns out that the First CommSat contract was already available in both programs. It was just weird that the commnet contracts didn't require it to be completed so I fixed that.

PR #2166, opened 2023-08-07
Formulaic engine pricing

Capkirk123:
At the moment, the vast majority of engines in RP-1 are balanced on gut feelings alone. Reliable sources for engine cost are very difficult to come by, and are further confounded by inflation correction, production rate, and politics. In an attempt to correct this, I have taken a few engine configs whos prices I do feel are reliable, and attempted to extract that into a usable parametric engine price formula.

I am relatively satisfied with the results, although I had to use a separate curve for hydrolox engines, which is somewhat less refined and doesn't really work with engines smaller than the RL10.

I have applied the output to all existing (and a few new) engine configs for testing and evaluation.

Issues include:

- ~~In an effort to preserve the early-game, I have tried to avoid making significant changes to pre-orbital engines.~~ Early game engines (well, mostly A-4 and derivatives) are not being spared. Testing still needed, but balance changes may not be as significant as first expected.
- The price curve strongly encourages bigger engines. This effect may be a little strong and could use adjusting.
- I have made no attempt to consider man-rating or other external factors in pricing. Some engines may need manual adjustment.


GameData/RP-1/Tree/ECM-Engines.cfg
```diff
@@ -10,10 +10,13 @@
 
 @ENTRYCOSTMODS:FOR[xxxRP-0]
 {
+    11D23 = 10000,RD-0216
     11D33 = 5000, S1-5400
     11D33M = 5000, S1-5400
     11D417 = 10000, throttlingTP
     11D417B = 11D417
+    11D423 =
```
NathanKell, line 18: uh, that's blank. Same for next one.

NathanKell (changes requested):
pretty sure these need to not be blank

NathanKell:
I think reliability kinda has to be taken into account. Other than that, neat plan!

isturdy:
I think this is a great idea, but I have a few concerns over the details. I think most stem from the use of mass directly--it's going to be correlated with cost due to the strong correlations between thrust and both mass and cost (And to some extent vacuum ISP via nozzle size), but after controlling for thrust I'd expect the reverse correlation--the design/manufacturing challenges to making a lighter engine usually exceed the cost of the materials saved. I think that either mass should be dropped or it should use the mass/thrust ratio.

To highlight a place I think this is producing questionable results, consider the '58 upper-stage engine situation. At present, I think the Gamma-2 and 0105 are fairly well priced--overall rocket build costs are fairly close, with the expense of the 0105 offset by its ability to use a smaller and cheaper booster. But there's no question it's the better engine--significantly lighter and higher ISP, with slightly lower thrust more than offset by a much longer burn time. But comparing engines by mass, the Gamma-2's high mass "offsets" its low ISP and they wind up priced quite closely--there's no longer any reason other than ECM optimization to use anything other than the 0105.

I think it may be best to try to make separate formulae for booster and vacuum engines--boosters care about thrust significantly more than anything else, while vacuum engines care primarily about ISP and total impulse.

Capkirk123:
Alright, patched up some obvious issues. I haven't seen any other significant balance problems, but it'll probably need more eyes on it for that.

Blothorn:
After a more comprehensive look, I see this much more favorably. In particular, previously engines had a fairly clear upward trend in both performance and price, which often left earlier engines/configs with a cost efficiency advantage, the fact that developments of an engine (and, less consistently, new engines) tend to reduce mass and improve performance means greater incentive to actually move along the tree. Two things I'll call attention to, though:
- It looks like the pump-Scud engines lost their adjustment for having pressure-fed integration modifiers.
- The identifiable category where I see the biggest change is broadly "human-mission-specific" engines--rocketplane engines and the lunar landing/ascent engines. I'm not opposed to this, but I think it does highlight the importance of attributes that this formula ignores--primarily restartability, throttle, and reliability. I expect to see a lot more use of the S-155 and XLR-35 under this.
- I wonder if it's worth adding a general discount for HTP-gas-generator engines. They are realistically easier to design and build than combustion-gas-gen engines, and this may avoid the need to special-case the Gammas while also helping the still-overpriced NAA-75-110, XLR10, and RD-200 series compete with the XLR41. (Granted, the anomalous cost effectiveness of the direct V-2 derivatives is even worse without these changes, but this seems the time to address it and I don't like the effects on the early game of addressing it by increasing the V-2 families' costs to prediction.)

Capkirk123:
Whoops, I adjusted the XLR11, but forgot the Scud. I'll get on that

Throttlability, especially deep throttle, might be a good factor to consider. It nicely hits stuff like the LMDE, XLR25 and XLR11.

The V-2 and a lot of it's direct derivatives took advantage of captured parts and tooling. I'd almost be tempted to ask for some sort of system where the player gets A-4s at the current low price, but then has to pay full price after a certain number of them are used. The XLR41 being extremely cost effective is probably my fault, I took a few liberties on that config due to the lack of sources. I should probably drop it's TWR to match the RD-100, rather than RD-101. Otherwise the RD-100 and XLR10 do offer a better $/kN ratio than the A-4, while the RD-200 is very similar. Only the Veronique is noticeably worse in that department, but it was a pretty awful engine IRL too.

Capkirk123:
Actually Veronique is pressurefed and doesn't get pump mults, so it's about the same as the RD-100

NathanKell:
Hmm, can't remember where/if I replied, but I think rather than dropping the current XLR41 config's perf, it'd be better to introduce an earlier model at the above pricepoint with lower perf, so there's a starting US engine? And/or the XLR10.

Per commit comment, A-9 config being so cheap seems strange to me, and also I think rather bad for gameplay.

As others have mentioned I strongly favor including reliability for the engine into the costing, both because it will make unreliable engines more compelling (and be representative of the cheapness involved in making them), but also because it will make properly expensive carefully human-rated engines.

NathanKell:
(That said I concur with the no-longer-undercosting-A-4s direction 👍 )

Capkirk123:
There's nothing that really exists between the A-4 and XLR41, the US got the lion's share of the leftover A-4s and so never built a direct replacement (even the XLR41 is fairly speculative). I suppose I could dig up the Hermes A-3 stuff, but it wasn't really better either from what I remember.

I was thinking of adding Wasserfall. Plenty were built, and it got around quite a bit (Scud and Veronique were derived from it).

I can mess with the A-9 a little. Maybe even move it, it's pretty darn optimistic considering people were still designing rockets to avoid air-starting in 1955.

As for reliability, that's a bit harder. You've got gold-plated stuff like the LMDE and SSME alongside mass-produced stuff like the RD-107/108 and AJ10-118K that were just simplified and refined forever until there was nothing left to fail. I could certainly discount really unreliable stuff, and increase the prices on stuff like the Titan GLV engines where only the most consistently performing engines were pulled off the assembly line and modified into GLV spec.

Capkirk123:
After having slept on it, maybe fixed multipliers for particularly awful engines (say, anything with an end reliability worse than 95%?) and specialized engines (LMDE/LMAE, XLR99, low production run high reliability stuff).

NathanKell (dismissed):

NathanKell:
Looked a bit more and while I still applaud the effort, and think it's useful for weird outliers or for engines whose cost we don't know at all, I'm skeptical of the changes on engines where we do have data. The LR79, LR89, and H-1 are a great example here: we know that the H-1 was a simplified LR79/89 and appreciably cheaper, but this is making it the same price. It's also extremely strange to me that XLR81 is only a hair more expensive than the early AJ10s, and the AJ10-190 is so much cheaper than the AJ10-137. On the Russian side, it's strange to see the RD-0105 being only a bit cheaper than the staged-combustion S1.5400 (and both of them being only 2/3rds the cost of the AJ10-137).

If what we're looking for is historical pricing, then per above I think we need to give more weight to actual historical pricing; if what we want is gameplay pricing, then I think a model that's more based off Thrust x (TWR / nominal TWR) x Isp x (PF or pump) x reliability x relights, modified by tech (so newer engines are uniformly cheaper _at the same performance_) is perhaps a better way to go?

Capkirk123:
The H-1 may be priced similarly to the LR79/89, but it makes much more thrust, resulting in a noticeably superior fund/kN ratio, not to mention its generally better Isp and reliability. And fwiw, I haven’t really seen any direct prices for any of those engines, just the “H-1 is cheaper” claim. XLR81 is also surprisingly evenly matched with the AJ10 for most of its run, its worse Isp and shorter burn time offsets the better mass ratio.

AJ10-137 is huge, it’s an AJ10 in name only and makes twice as much thrust as the next most powerful AJ10, and is relatively heavy on account of having almost every part duplicated. The -190 is one of the weakest AJ10s, and redundancy was provided by a second AJ10. The -190 being the significantly cheaper engine doesn’t seem that unusual.

On the Soviet side, the S1.5400 is several years newer and in a different research line than the RD-0105, so it’s performance being better makes sense.

I mean, as is, this is about as heavily based on historical pricing as I can, which also seems to have produced a fairly balanced pricing model.

Rather than using thrust and TWR, I just used weight directly, the formula already uses ISP, and integration multipliers quite heavily penalize pump-fed engines (further leveling the AJ10-Agena playing field). Adding restart capability generally also increases mass (see the restart kits for Agena or J2), and I am looking into reliability modifiers.

NathanKell:
My point re: TWR is the opposite of yours: an engine with a higher TWR, all else being equal, should cost _more_ (higher machining tolerances, more advanced materials required, etc), but on your model it costs _less_. As to the rest while you are correct about pump-fed multipliers making such engines take longer to integrate, it's still the case that they're more complex to _manufacture_, and that's what price represents (well, apart from all the weirdness around how aerospace companies price their products). Further, restart capability does _not_ generally increase mass appreciably on pressure-fed hypergolic engines, since it's just "open or shut the valves", no? Put differently, I would expect its effect on cost to be over-proportional to its effect on mass.

NathanKell:
I also tried to dig up historical pricing on the H-1, I'm pretty sure I remember finding something saying it was about $200k in serial production, but I recall RP-0 not using that number because it was too low compared to the LR79.

Capkirk123:
I already explained my logic above. Jet Engines tend to get better thrust/$ with higher TWR, and I strongly suspect rocket engines are the same. 1 kg of engine costs about the same as 1 kg of any other engine, so the engine with the higher TWR will be better.

ppboyle:
I'm going to have to go with NK here.  I'm not saying the current tuning is great - there are a lot of engines the player wouldn't touch if they're paying attention to price/perfomance balance.  But I think this tuning, if anything, is roughly the same, and may exacerbate the problems of only certain engines being at all worthwhile.

First, because it high TWR is already an advantageous situation, doubling down by making higher TWR engines cheaper makes already good engines just better.   Also, making heavier engines that were generally more expensive due to low production runs cheaper instead exacerbates the inherent advantage of less big engines > more small engines for reliability purposes.

Then ignoring other advantageous/disadvantageous factors - reliability, ignitions, throttability - things that genuinely were costly either in terms of development cost, or integration/testing/validation cost, just makes tuning scattershot.

If we're going to bite off re-costing everything and the inherent wailing and gnashing of teeth caused by that in the community ( I retuned the Wolfhound in KSP, I should know) - the tuning should at least improve the situation in terms of making engines more balanced on a cost vs performance basis.

I also seriously question the relevance of the paper, if this is the one you used - https://www.rand.org/pubs/monograph_reports/MR1596.html  

1. Only kerosene burning military engines released 1965 to 1996's are considered - a time period with many advances, but notably skipping the early days of jet engine development 
2. RAND balances against a production run of 1000 engines, which negates the production advantage a smaller rocket engine would have.  
3. The range of engine sizes (With one exception) is within a factor of 5x for all of RAND's data.  
4. All the RAND engines are built for the same purpose and using the same technology (turbojet) vs rocket engines that had different turbopump setups and cycles and fuel types. 
5.  All of them have relatively high flight hours and therefore comparable reliability.

 If the J58 had been on there - an engine with genuinely different technology - it might be a more useful comparison to see if the 'statistically significant' variables listed by RAND were true when a more esoteric engine was included in the sample.

Overall I would not be in favor of implementing the tuning changes here, unless more consideration were giving to balance factors that NK mentions in his post

Capkirk123:
I don't see any issues with the formula as it exists now. I experimented with using thrust as a factor, and the results were pretty much the same, with the exception that I couldn't get small engines and large engines to have sensible prices at the same time with the same formula.

I'm looking at the results of formula on cost/kN, and I'm pretty happy with it. Newer engines generally have better value than older engines, upgrades generally have better value than base configs, larger engines generally have better value than small ones. Sure, this results in the list of "best" engines being full of stuff from the 70s, but that's mostly because new engine development slowed significantly in the 70s. The few new projects that did happen, like Merlin or the SLI studies, pop up pretty close to the top of the list too. That seems to meet both gameplay and realistic balance requirements to me.

For upper stage engines, things are a little less clear-cut. Cost/kN doesn't have as clear of a trend, but it's also far from the whole story, since upper stage engines care a lot more about Isp, restarts, etc. Smaller engines see it stay fairly constant or increase within a family, while larger engines tend to see a small decrease. However, upgrading is still probably much more incentivised than with booster engines, as the increases in Isp and burn time are much more significant.

Capkirk123:
That's not to say I'm opposed to making changes, considering reliability makes sense as a gameplay factor, and although I'm not convinced on ignitions, it will almost exclusively apply to upper stages where balance is admittedly a little fuzzier.

Ultimately though, I don't think mass-based price balancing is an issue. It works as a good stand-in for multiple factors. It's tied pretty well to thrust, pressurefed vs pump-fed (a pump fed engine is generally heavier for the same thrust due to the extra machinery, up to a point), and tech level (engine TWR generally increases with time. Outliers like the XLR43 have unusually good TWR for their time, but also have poor burn time and reliability)

NathanKell:
Busy work day so can't dig in, but I did notice that the J-2S seems all kind of messed up? It was a simpler, cheaper J-2 and its config price went from "cheaper" to "much more expensive" ?

Capkirk123:
Most of the "simplified upgrades" (H-1, F-1A, J-2S) are actually more expensive than their predecessors, but provide a lot more thrust and performance so their overall value is still much better.

J-2S is a bit of an outlier, the J-2S as configured always has a restart kit installed, while it's optional on the normal J-2s. This means it's a lot more expensive, but if the restart kit is considered then the price difference disappears. I should probably make the restart kit optional on the J-2S as well, but even with it, it's still better value than the standard J-2 (similar cost/kN, and better Isp).

NathanKell:
A few more outliers:
* F-1 is I think quite underpriced. My recollection is that historically a single F-1 was considerably more expensive than a cluster of H-1s of equivalent thrust, and for gameplay purposes the F-1 has higher specific impulse and, more to the point, higher reliability (due to clusters = many dicerolls, regardless of _actual_ reliability stats).
* I'm surprised Gamma 2 and Gamma 8 are so close in price.
* RL10 costs nearly as much as a J-2? I mean, yes that's probably historic, but I'm surprised the formula is yielding that.
* Deep-throttling should definitely have a price penalty IMO.
* I continue to think that pressure-fed engines vs. pump-fed should have more of a distinction. I was surprised that you said that pump-fed engines tended to have worse TWR since that was not my understanding--I thought they were broadly equivalent in vacuum and obviously pressure-fed is considerably worse in atmosphere.
* LR91 costing about as much as an LR87 is, again, about what I'd expect from reality but not at all what I'd expect from your formula.
* Why is the LR89 so much less expensive than the LR79, despite the former being the 'performance' model and the latter the 'workhorse' model??
* I would expect vac Merlins to be more expensive than SL Merlins, given the nozzle extension.
* X-Plane engines should be more expensive IMO (and especially S-155 and XLR25)

I know the Soviet side less well but at a glance it seems workable to me, excepting general notes above.

Lastly, since this is changing the general costs for liquid engines, have you thought about what it does to the relative balance of BP for capsules, avionics, tankage, and engines? It will definitely shift that around some. And similarly for solid vs. liquid comparisons, particularly for upper stages.

Capkirk123:
The formula does favor larger engines. The F-1 price is pretty close to historical as far as I can tell. I can try to find info on how much an H-1 cost, it would be a very useful data point.

All the Gammas are a single turbopump feeding multiple chambers (part of the reason why I'm not a fan of making a standalone single gamma), which is why their price is pretty similar despite different sizes.

As is, RL10 costs a lot more than a J-2 with later upgrades.

Yup, I can add a factor for deep throttle

Pressure-fed engines vary. Sea level pressurefed engines actually tend to have very good TWR since they have no nozzle extension. Stuff like the aerobee, U-1250 and Valois have some of the highest TWR in the game (over 150).

LR87/91 costs are basically the same as their old costs. LR91 is smaller but has much better Isp. I did not put the vacuum LR87 in, but if I did I suspect it would be a lot more expensive from the nozzle extension.

LR79 and LR89 are basically the same cost. LR89 just looks cheaper because the first config is the Navaho config, and so the part cost is set to that. If you look at total cost with the production configs, LR89 is more expensive.

Merlin Vac configs are much more expensive?

X-Plane engines will probably get whacked by the deep throttle penalty, so that should take care of it hopefully.

Otherwise, engine costs aren't that different? Certainly for your workhorse boosters and upper stage engines (LR79/87/89/91, AJ10, RD107/108) your cost changes are less than 25%. Small engines change a lot, but they were fairly negligable in the first place.

NathanKell:
That all makes sense thanks. Regarding H-1, I *think* the cite I found was for a purchase agreement with Rocketdyne for some number of x engines for the last run of IBs, which gave marginal cost at that flight rate.

Capkirk123:
I found some sources on the H-1 cost. The average price paid for the H-1 by NASA was ~405k (undated $), while studies indicated costs as low as 240k (1965 $) assuming 30+ Saturn IB flights a year. While the latter might not be out of reach of the average player, the former matches the calculated value pretty well, which is a little reassuring.

NathanKell:
And I assume the F-1 cost is for similarly-historic flight rates? If so then huh, interesting. Thanks for digging!

NathanKell:
Ok then pending tweaking for deep-throttle, and maybe a fudge factor for reliability, I'm satisfied, although I do need to check and make sure early V-2s aren't too nuts (I do think we shouldn't _force_ players to start with sub-V-2 LCs).

@ppboyle anything else?

Capkirk123:
F-1 is a bit more expensive probably? The source includes the J-2 and F-1 prices together, but assuming our J-2 prices are relatively accurate, our F-1 prices are a little low but not too far off. An F-1 did appear to be cheaper than 8 H-1s

Capkirk123:
> Ok then pending tweaking for deep-throttle, and maybe a fudge factor for reliability, I'm satisfied, although I do need to check and make sure early V-2s aren't too nuts (I do think we shouldn't _force_ players to start with sub-V-2 LCs).

I played a short V-2 based career in rp1 speedrunners with new A-4 prices. Although I really should finish it, it didn't slow me down a terrible amount. I started with aerobees, but I was able to get V-2s out pretty quick and was only about 6 months behind the current pace.

NathanKell:
"I started with aerobees" -- yeah that's my point, then it's not actually a V-2 campaign. :P I want it to remain viable to _start_ at V-2s. Like, not optimal, and not speedrun-compliant, but a reasonable experience.

Blothorn:
I’m sure it’s viable—I think that on a single pad you’re looking at finishing your first SR program in ‘53, and the second in ‘54 with luck. (And things should be faster if you run unguided rockets for the contracts that allow it.

Whether it’s a reasonable experience is another question—it’s a long wait to a first guided V-2 launch with the changes. However, I think that if you turn down not only Bees but also unguided V-2s, Vikings, and X-planes (any of which can be launched by August ‘51) it’s somewhat self-inflicted.

Karhgath:
I will try out this rebalance soon, but I looked at the changes and the current strategy I am "speedrunning"(not a caveman so proper node usage) seems to not be affected much and could be better now, comparatively, since it is a full V2 alternative - Wait for XLR10 for first launch, upgrade to Vanguard on same LC.
  
I will try an unguided vs guided V2 start with the rebalance and see if it seems reasonable.

Capkirk123:
Deep throttle and overall reliability modifiers implemented. Might be a little rough around the edges, haven't double-checked anything.

NathanKell:
Thanks! Gonna look over the sheet. Honestly I think it's probably easier if I just write some stuff in excel to output cost corrections directly from the sheet, rather than having to type then in everywhere, until we end up with what we're shipping.

Karhgath:
The dual LR87 titan booster is priced the same as the single LR87 unit. Same with the RD-216/215. Might be some other dual/quad engines with the same issue.

NathanKell:
Per discord I'm going to assume this is a result of my tweaks not the original

NathanKell:
Merging in prep for cost tweaking

PR #2169, opened 2023-08-07
Fix MM issue by splitting ORed nodes

StonesmileGit:
Fix two places where nodes that are not top-level were attempted to be filtered by using `|` to match multiple nodes (this doesn't work)

This might affect balance, so testing is needed!


NathanKell:
I'm gonna have to check and see if I need to tweak these costs, yeah (or tweak resource costs) but merging for now.

PR #2180, opened 2023-08-17
Clarify description of Early Comms Network program

Amselfass:
Since many other programs give the player hints regarding the minimum research required to complete them, I have added a note that Comms TL2 is required for actually having a useful network (even though I think the contract might not enforce it). I have also clarified the unhelpful statement about „fairly high“ orbits.

### DIFF
--- a/GameData/RP-1/Programs/Programs.cfg
+++ b/GameData/RP-1/Programs/Programs.cfg
@@ -292,7 +292,7 @@ RP0_PROGRAM
 {
 	name = CommNetwork
 	title = Communication Network
-	description = Communicating with our satellites, and eventually our crewed craft, can be difficult at times. Unless we have line of sight to a tracking station, communication isn't possible. This program tasks you with creating a communications network of 3 or 4 satellites that will provide much better communications coverage. Each satellite will require 125 units of ComSatPayload, and the orbits are fairly high.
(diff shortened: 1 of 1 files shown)


Blothorn:
I think this just promotes the misunderstanding that this is supposed to be a stock-style relay constellation for player use, not a commercial network that would not normally have hardware for connecting to other spacecraft.

Amselfass:
I guess you can pencil me in as another victim of this misunderstanding then. :D

LuneoLucy:
I definitely agree on changing the program and even contracts definitions of the early com network I don't really understand what this wants the player to think of it, i'm not sure this is the change it needs but i hope your PR will contribute to the potential change :) 

As for discord reactions, don't mind it; it's discord

NathanKell:
These contracts (and this program) is getting replaced by a Commercial Applications program, which should hopefully clarify things.

It's worth pointing out that yes, indeed, this program is about creating a comms network _for terrestrial use_, not in the stock-KSP sense, so anything that encourages the player to think in stock-KSP-style-comm-network terms is most certainly a trap.

NathanKell:
The change re: orbital altitude is useful however (it's worth just saying it's a MEO constellation, rather than a specific number, IMO) and we'll bear that in mind for the rework. Thanks!

PR #2184, opened 2023-08-19
Crewed Interplanetary flybys Program

CXG-2827:
Program combining Legacy Crewed Flybys of Venus and Mars and added optional contract for 14 day duration high Earth orbit.


CXG-2827:
> Not that I want to discourage any new programs/contracts, but here's an amusing but great write-up about why crewed fly-bys were such a terrible idea:
> 
> https://www.projectrho.com/public_html/rocket/realdesigns.php#badidea

According to the citations linking to the original article, this was written by a free-lance writer. So this is really just an editorial opinion piece.

Although it has some valid points, the fact that there are still design studies for Crewed Flybys being conceptualized makes me think there is merit in these missions:

Boeings website (Click the "Hello, Mars" Tile):
https://www.boeing.com/space/space-launch-system/launch/future.html

And the presentation:
https://www.exploremars.org/wp-content/uploads/2023/08/0517_0840_MDuggan_Mars-2033.pdf


I would like to bring up some points regarding this flyby Program which I believe gives reason to include in RP-1

**Existing Contract Conditions and Spyware Metrics:**
- Crewed lunar flyby and Lunar Orbit contracts can be taken simultaneously, meaning you can take the course of flying both separately or knock both out in a single flight with an orbiter-rated craft.
- Early Inner Planet Probes contracts allow for the Orbit contract to be completed without launching a new vessel.
- Spyware activity for those who have reached either Venus, Mars, or both shows that a majority of people complete the flyby and orbit in the same flight.

**Program Proposal:**
- Assuming that most people would take Crewed Venus/Mars programs with an architecture that gets straight to orbit rather than send a precursor crewed flyby, then the Flyby contracts should be excluded from the Crewed Venus/Mars Programs. 
- Crewed Flyby Program can be viewed similar to the X-planes branch. People can opt to run the program, or skip this entirely.
- This program should not be a prerequisite to the Crewed Venus/Mars programs

SierraHotel84:
I just found it funny since I was researching a program I'm putting together for nuclear propulsion and had just read through this one.  Your proposal makes sense, and the two could work together very easily as well.  I would probably treat it just like the early lunar missions, first a fly-by, then an orbit.  Have the crewed landing missions be a wholly separate program (to Mars anyway, not sure we even have equipment that will allow a crew to survive Venus?).

My only concern is there's not a lot of meat with just 3 contracts.  

Though (light bulb goes off) with the projectrho website fresh on my mind, having the program require using the same ship the entire time, and add in some resupply and crew change contracts?  That could be interesting.

CXG-2827:
Yea, its a bit of a light program being standalone.

I know on discord it has been mentioned that the majority of players barely make it to Moon boots, and even fewer get crewed Venus orbit/mars landings. With this being standalone from the other Programs, I would imagine this is an easier stepping stone from Lunar landings. It could encourage folks to take that next step without being too overwhelmed with mission architecture and planning as it is not committing to Mars boots or Venus orbit.

From my experience with Kerbalism running interplanetary missions in JNSQ, crew stress and radiation buildup is a major factor that limits how long the round trip can take. Even watching some RSS/RP-1 playthroughs on YouTube, crew are coming back from Mars almost fully radiated and on the brink of mental breakdowns. I am not sure if adding crew-rotation contracts would make sense for this specific program, especially if the idea behind it is to keep it fairly "lightweight" for planning and execution.

Regarding using the same ship, this can be accomplished by allowing both the Mars and Venus Flyby contracts to be taken simultaneously (Like Crewed Lunar Flyby and Orbit). Giving the player flexibility to run 2 missions or try for a double flyby.

SierraHotel84:
> From my experience with Kerbalism running interplanetary missions in JNSQ, crew stress and radiation buildup is a major factor that limits how long the round trip can take. Even watching some RSS/RP-1 playthroughs on YouTube, crew are coming back from Mars almost fully radiated and on the brink of mental breakdowns. 

This is why I was thinking a crew rotation contract in the middle to put fresh crew and supplies on board.  Making it two separate trips with a new crew and fresh supplies seems like an easier combination, while making it all one long trip seems pretty difficult to pull off.   It wouldn't be difficult to give it a few different ways to play it:

a) Separate ships, separate crews, one for Venus, one for Mars
b) Same ship, separate crews (The Martian approach)
c) Same ship, same crew  (The revised rescue mission Martian approach)

The most dangerous but highest payout mission would be C, but the program could be completed doing A.  Something like that.

CXG-2827:
Most Flyby proposals involved skipping a propulsive capture back at Earth, and instead departing the craft in the command module for a direct re-entry (with a slowdown burn) while the flyby craft continues on to a heliocentric orbit. 

If the craft is designed for propulsive capture/crew/rotation/refit I am not sure if the program funding curve would balance correctly. Now you need the extra 3-4k/s for the capture, and this impacts your earth departure stage needing to be even larger. Additionally, you have to factor in rated burn time of the engines as well as ignitions available.

Options A and C would be manageable. C would not be as difficult as you imagine as this can be figured out using KSPTOT or the MGA planner. This was even a soviet proposal:
https://en.wikipedia.org/wiki/TMK

SierraHotel84:
Oh I definitely understand there were proposals for double fly-bys, I was just responding with regard to the bit about Kerbalism and long duration crews not playing nice.  I haven't done any long crewed missions so I don't know what a reasonable limit is and if Option C would fit within it.  

From a historical perspective, it in no way surprises me the Soviet Union wanted to put 3 dudes in a Space Winnebago for 3 years and fly them around the inner Solar System. No thanks.

Blothorn:
I think these should just be optionals in the associated programs. From a gameplay perspective, this is quite sparse for a program--if a program can be completed in one flight it's a contract, not a program. From a realistic perspective, this is a pure flag-showing exercise that neither accomplishes a unique scientific mission nor significantly furthers manned exploration. (It does increase experience on the effect of radiation during interplanetary flight, but given that there's no way to abort the mission it's just as dangerous for this to be the first mission as for orbit/landing.) I think programs should generally build up to sensible medium- or long-term ambitions; this feels like a preparatory step (and one of dubious worth, at that) toward something bigger.

CXG-2827:
For me, this came mostly from the perspective of career progression and flexibility in playthroughs. I have not played Legacy before, but from my understanding it was more A La Carte, where you can pick and choose contracts at your leisure as prerequisites were met.

Assuming no comments from NK in the next week or so I will just close this PR

SierraHotel84:
No reason to close it, there is definitely historical precedent for it. Just trying to help build it up some is all.  And to your point, having options in contracts to do is never a bad thing.
________________________________

PR #2197, opened 2023-08-26
Disable the Venus Rover contract

Vlood27:
As you wish.
image


basilatrue:
i did Venus rover, it's absolutely possible. well, at least in RO :)

NathanKell:
It's possible in KSP/RO (or was before Kirk's thermal rework). That doesn't mean it should be. :D

PR #2198, opened 2023-08-27
X-Planes overhaul

Blothorn:
As much as I like planes, I have a few frustrations with the X-Planes contracts/program--some going back to legacy, some new with PLC.
- It's somewhat tedious. Despite not having any unlimited repeatables, a full completion of the program involves about 30 flights--even ignoring the levels of SS High that require High-speed Flight or above. This is quite a lot of flight time, and much of it non-AFK game time for the many people who haven't fully automated these flights. (Also, the high number of cooldown contracts combined with crew training requirements can require a lot of bookkeeping.)
- There isn't a lot of variety. All missions are basically either "hold X speed with a jet" or "reach X altitude with a rocket plane"--the only deviation being the rocketplane development contracts that somewhat constrain the flight path to reach an altitude. It's possible, and in PLC quite attractive, to complete the program with two planes, each optimized for a single mission profile, and one of which repeats the exact same kOS script over a dozen times.
- The funding fits awkwardly within PLC. Now that crewed orbit is split, no program comes close to the time and tech range--the first handful of missions can be easily completed in the first year, while the last requires post-orbital tech. This means that from start to finish the opportunity cost of the program slots increase considerably, and thus its funding is only competitive for a fairly narrow range of start dates and speeds--most of which require painful compromises. (And I think it's actually more expensive to run than an SR program--almost all of the tech required is dead-end tech you could otherwise skip without issue, and hiring a pilot in the first year is a significant expense. Meanwhile, since you need to complete an SR program to advance and both overlap heavily in tech and LC requirements, you save very little from only taking one.

I've attempted to address this from a few directions. I think these are largely separable if you think some but not others are good ideas; I've lumped them in this PR so I don't have to also balance them separately.
- Split the program into an early program (roughly covering the X-1 through X-3 missions) and a late program (X-15). This allows people to start planes late while still seeing X-15 funding while it's still vaguely competitive. (Particularly attractive to double SR/heavy sat playthroughs where you need to wait for either two admin upgrades or finishing both SR programs to take early.) I considered further splitting early into 1-point jet and rocket programs to appease people who are bored of jets, but I think that would make taking at least one of the plane programs at the start to fill out the fifth program slot much too attractive.
- Reduce the number of missions--this cuts the Supersonic and 80km Altitude milestones, reduces the repetitions of most optional missions by 30-50% by increasing the gap between contracts, and bumps up some reputation rewards to partially make up for the reduced count.
- Add some new missions. The Stratospheric Research missions are modeled on the cover story for the U-2, and I think provide an interesting design challenge that doesn't consist of optimizing the wave-drag coefficient. The hypersonic research missions are modeled on the high-speed low-altitude X-15 flights, and probably don't change plane design much but at least provide something different to do.

Aside from contract testing, I've done a partial career taking X-planes at the start. I still think that's suboptimal; hiring an astronaut or two and adding at least one point of research that doesn't benefit sounding rockets can't compete with the synergy between the SR programs. However, the reduction in confidence for the smaller program and better first-year funding feels a lot better than the current slow start. I think the situation will be somewhat better for careers taking it alongside sounding rockets after an admin upgrade. I expect the X-15 program is somewhat undertuned--if you take it in late '52 it's respectable money through 53 and 54, but after FO it blocks the targeted/comms satellite programs that are even more lucrative despite being single-slot programs--but I haven't tested that stage enough to dare push it higher.


NathanKell:
Firth, thanks for taking this on!! Some comments:

I remain deeply skeptical of separating the early X-plane flights from the X-15. I think the reasoning is pretty clearly demonstrated by your X-15 program: it requires only 1 flight to complete. You could stretch it to 2 flights if the contracts couldn't be both accepted at once, but...that's really not much of a program. Even the single-contract Comm Net program would be four _launches_ except for the fact we don't check coverage (or a single launch on a chonky LV indeed).

Regarding flight count: you've ended up with the Early X-Planes program having only one fewer required flight than the existing, combined program. That there are a bunch of optionals is IMO neither here nor there--there are currently _unlimited_ optionals in the Crewed Orbital programs, as you mention. That the flights are more expensive in real-world time than orbital missions, however, I grant, which is an argument for making fewer of them with higher rep rewards IMO.

As to duration (that the program is so much longer than the other two starting programs) -- I think that's evidence that the other two are too easily rushable, rather than necessarily this one being too long. Certainly no starting program was intended to be completed in 1952 (!) or even 1953.

I love the additional variety you've added with the Stratospheric contract. I'd be happy to remove a required contract and replace it with that, and a short string of optionals. The hypersonic one, as you say, is not really a different design challenge, but is reasonable to include as an optional.

What about somewhat increasing the science rewards from cockpit experiments (I nerfed them hard right before ship), and increasing the funding a bunch more, with a high initial spike to cover naut hiring. Then removing, per above, a required contract or two ( `XPlanesSupersonic` can go away, for instance--if you want to fly optionals and later edit in a better jet engine that would take care of it, though I kinda hate having to only fly a jet a _required_ one time...) and removing the requirement on RPD so you just have to hit 40km, keeping the optional RPD contracts so you can fly missions on the X-1 (not X-2) cockpit? That would net-reduce requireds by 1 if we make the Stratospheric one required, increasing required plane builds to 3 (although likely 4, and 2 of them getting edited with better cockpits/engines).

ppboyle:
Glad someone is looking at this - and I definitely like adding more variety to contracts.  I had some ideas for doing more of the historical x-planes as contracts - particularly X-6,7,9,10, and 13/14, as additional optional.

However, I'd note that 'taking a long time to fly' is not a thing to balance against - in particular, because rewarding people for doing things that you feel are more tedious is a good way to get them to bore themselves.  Especially when many of those are optional contracts.  The x-planes programs should be for people who want to build & fly the planes, not for people to be dragged into kicking and screaming to do the planes. This isn't like an MMORPG where time=reward.  It would be the same as adding extra rewards for playing the game with Principia on.

Further - not sure that 'too many not-unlimited optionals' is a thing to be concerned about - I wish more programs had more optionals, and its good you're adding some.

I think there's overall been a continual reaction on the part of players to reduce programs till they can be done as close to immediately as possible - either by reducing contract count or limiting them to only requiring a narrow range of technology - that's not what programs are meant to be - they're meant to pay for doing the research & facilities to do them.  some programs fail to adhere to that, like commsats and geo sync, and those are the ones that most need to be adjusted.  I understand why players want this - they typically would like absolute minimum friction and the first place they look is the point of highest friction.  Sometimes when that friction is pure busywork or poor UX, that is what needs to be addressed, but sometimes it would just be a case of overly simplifying a deliberately complex endeavor.

Tsits11:
I just want to say that I agree about the contracts and the program duration , however the science rewards from supersonic flight and supersonic flight 2 had been needed hard and isn’t worth it. Also it takes forever( used to be less flight time) now I need to fly supersonic for a total of 40 min (thereabout) in game to earn only 20(???) science.

just bump up the science rewards of reduce the duration needed for a start

Blothorn:
@NathanKell --that makes sense, I hadn't thought much about the depth of the X-15 program. I'll update along those lines.

@ppboyle--note that I'm not trying to boost rewards to match the time investment, I'm trying to reduce the time investment to match the rewards (somewhat). As it stands, a rocketplane development contract gives the same reward as an altitude SR optional, and the highest rewards in the program were less than that of a downrange optional. It's also worth note that because X-planes is such a detour in terms of technology and doesn't go anywhere, I don't think it'll ever be worthwhile to do a minimal, milestones-only completion--the main attraction of X-planes is the ability to do a lot of optionals with modest build investment. (Although the fact that recovery/airlaunch prep require assigned engineers means that the advantage there is significantly reduced relative to legacy.)

As far as other X-planes go--I don't think the X-6 can be implemented without somewhat radically new parts, and nuclear jets don't have much other application to make it worth adding them. Same goes for the X-14, and I think the X-13 would be more annoying than interesting to build, at least for people who fly by hand. The X-10 is basically an unmanned version of an XPSS mid plane, and I think largely repetitive. The X-9 is novel and could be implemented with parts already in RP-1, but I'm not sure what a contract for it would look like. (I'm also not sure it belongs in a "plane" program?) The X-7 is finally a novel aircraft type with reasonable contracts, but the only existing ramjet is out of reach (adding an AJE dependency would over double the research cost of the program, and add five years to the nominal unlock date. There's a historical case for making ramjets available earlier, but I worry that it would upset the balance of the existing contracts.

NathanKell:
@Tsits11  That's a good point re: experiment duration, we could lower the duration some which will discourage grinding.

@Blothorn yeah I agree that having fewer contracts with larger lump rewards than Downrange or Alt SR makes sense here. Max is right that we don't want to encourage the grind by increasing rewards without changing other factors, but if we change other factors we can decrease the grind I think.

ppboyle:
For the x-6 - which never flew with a real reactor anyway - I was thinking of having the player fly a plane with a large quantity of lead ballast to simulate the actual test, and make the player build a cargo plane, essentially.  And I've built my own X-13, its doable.  Is it annoying?  Anything is annoying if you don't like doing it, hence making it an optional contract.

Blothorn:
@NathanKell I've updated to my understanding of your requests--let me know if that understanding was mistaken in places. (In particular, keeping the increment-between-contracts and reward increase for XPSS and XPA.) So the summary of changes from master becomes:
- Remove RPD and XPSS required contracts
- Add a stratospheric research requirement and two optionals
- Remove anti-grind cooldowns for optional series with <= 3 completions, as I understand is now the intended threshold
- Add a "low"-altitude hypersonic optional series
- Increase program funding to 800k and use a new funding curve with a small initial boost
- Modestly increase science rewards for cockpit science, and decrease durations for the supersonic/mach 2 experiments. (I think achieving the duration for the rocketplane ones is a worthwhile challenge to keep.)
- Slightly relax some experiment/contract speed ranges to increase the extent to which you can satisfy contract and science requirements at the same time

@ppboyle That's a good point--I, at least, wouldn't mind an X-13 optional-only program, although I'm not feeling inspired to write the contract for it.

NathanKell:
That looks quite good to me! The one thing that I somewhat miss is a required contract that necessitates an X-1 or X-2. By removing both the required RPD contract, and the 40km High contract, there's now no reason other than optionals to build such a craft. So instead I'd suggest making the High Altitude contract a 60km requirement, which is within the X-2 cockpit's capabilities.

Regarding the new curve, did you see what the funding curve looks like ingame? Since what's shown there is the derivative, it can end up looking pretty funky.

On the testing side, here's what I'd check if I were you:
* Loading a few savegames partway through the (existing) program to see what happens / how broken they become.
* Cheat-completing contracts through to make sure the various branches (no optionals, some optionals, all optionals) all work as expected.
* Trying out as many of the contracts as possible ingame (although not the various "steps" of the multi-step contracts, that'd take forever). Quickest thing would be accept contract, make sure the situation eventually goes green, and then rather than completing the duration and landing requirements, revert and cheat-complete the contract and go on to the next.

NathanKell:
@ppboyle thoughts on the numbers balance?

Blothorn:
The 40km Altitude requirement is still there (I think?), but not the 80km—this way there’s one requirement each for an XPSS jet, X-2, U-2, and X-15. (I’m happy to move it to 60km, but in my testing 40km is about as high as it’s easy to get a pressure-fed XLR-11, and I wanted to encourage building an early rocket plane rather than starting with jets.)

The finding curve looks fine in-game—I’ll try to screenshot it and re-test contracts for the latest changes tomorrow.

NathanKell:
That makes sense to me!

Blothorn:
Screenshot_26

NathanKell:
I think you might want a bit more of a bump early to pay for nauts and engineers to integrate your first plane? Starting at 70k and decreasing to 50k by 6mo in, perhaps? But that curve looks broadly fine too :)

Blothorn:
New curve: 
Screenshot_27

ppboyle:
I disgaree with removing much of the anti grind from both the required and optional contracts in supersonic and altitude.  You've essentially made the program completable & rewarding for someone who's done a few researches in the mid-to-late 1950's as fast as they can recover and relaunch a single jet and a single rocketplane.  Would make this program a cheat for someone picking it up and rapidly doing it for the reputation, early-completion reputation, and confidence.

Blothorn:
You can say the date of most programs—build all the rockets you need, accept the program, and then launch them as fast as pad refurbishment/rollout allows. And unless you have a large (and expensive) pilot corps, rest/training is the limiting factor—depending on which cockpits you use I think doing all the missions (even excluding >700m/s XPSS) with a single pilot would take around two years.

My motivation was actually the opposite—since a normal completion takes 6+ years, adding anti-grind with a reasonable interval allows spreading out missions and roughly doubling rewards with little downside.

NathanKell:
I think if anyone wants to take this program after they've already (nearly) researched the X-15 cockpit, and spend the engineers and pilot-hours to build and fly jets when they could have a more lucrative use of the slots, that's ok?

@Blothorn have you had a chance to test this in the ways I suggested above?

Blothorn:
Yes

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -5,8 +5,8 @@ RP0_PROGRAM
 	description = This program tasks the space agency with the development of experimental aircraft to test aerodynamic behavior at velocities and altitudes previously unreachable by crewed craft.
 	objectivesPrettyText = Complete X-Planes (Mach 2 Supersonic) and X-Planes (Karman Line) contracts.
 	nominalDurationYears = 9
-	baseFunding = 600000
-	fundingCurve = BackloadedFundingCurve
+	baseFunding = 800000
```
NathanKell, line 8: Using the spreadsheet, put this number / duration in the "average funds/year for normal" column and it should spit out the confidence costs and the rep (apparently we switched to symmetric rep, so use the higher number for both early-complete and late-complete).

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -373,7 +373,7 @@ RP0_PROGRAM
 {
 	name = EarlyLunarProbes
 	title = Early Lunar Probes
-	description = The Moon is the first stepping stone to exploring the solar system. More than a light-second away, operating a probe at that distance is a worthy first step. This program requires new technology and heavier launch vehicles to take your space program's first steps toward studying the Moon up close. Complete it by putting a probe into orbit around the Moon. This program will require Lunar Range Communications at a minimum, and will generally require upgrading Mission Control and the Tracking Station in order to plan your journey. Basic Avionics and Probes unlocks deep-space avionics, which will be essential if you need guided avionics at the Moon.
```
NathanKell, line 376: Looks like some unrelated changes got stuck in the rebase?

NathanKell: (Affects everything after this point in the file.)

GameData/RP-1/Science/Experiments/CrewScience/CockpitExperiments.cfg
```diff
@@ -63,8 +63,8 @@ EXPERIMENT_DEFINITION
 {
     id = RP0SupersonicLow2
     title = Mach 2 flight
-    baseValue = 10
-    scienceCap = 10
+    baseValue = 12
+    scienceCap = 12
```
NathanKell, line 67: I think this could even go up to 14 12 still feels a bit dinky.

NathanKell (changes requested):

NathanKell:
Thanks!! I agree on making hypersonic required, for sure.

GameData/RP-1/Programs/Programs.cfg
Blothorn: Thanks, reverted.

GameData/RP-1/Programs/Programs.cfg
Blothorn: Done. The resulting rep gain seemed high so I reset the multiplier back to 1 (hence 130/year rather than 195). I reduced the confidence multiplier from 0.8 to 0.7--I think cutting off the fast X-Planes/fast SR makes it inviable to take as a starting program with the present curve, since both normal X-planes/breakneck SR and fast X-planes/normal SR pay less over the first two years than breakneck/fast SRs even before accounting for astronaut costs. I don't know how strongly you want to avoid playing with those multipliers--I think a viable alternative would be increasing program funding and reducing how much it's backloaded until fast X-planes/normal SR pays competitively at the start, but fussing with the curve is enough more work that I want feedback on which approach is preferable before doing so.

NathanKell: Yeah I think it's fine to go that approach rather than messing further with funding or curve 👍

NathanKell:
Got this on launch:
```
[ERR 23:46:29.967] ContractConfigurator.AcceptContractRequirement: contractType 'RocketPlaneDevelopment' must either be a Contract sub-class or ContractConfigurator contract type

[ERR 23:46:29.967] ContractConfigurator.CompleteContractRequirement: contractType 'RocketPlaneDevelopment' must either be a Contract sub-class or ContractConfigurator contract type

[ERR 23:46:29.970] ContractConfigurator.ContractType: CONTRACT_TYPE 'RocketPlaneDevelopmentOptional': Error parsing description

[EXC 23:46:29.970] Exception: Unknown identifier '@/rewardFactorPercent'.
```

PR #2205, opened 2023-08-29
Update Recommended/Suggested packages in ckan

Qazerowl:
The current ckan metadata for the RP-1 package does not match the more up-to-date [recommended extra mods list on the wiki](https://github.com/KSP-RO/RP-1/wiki/Extra-Mods-to-Consider). I have fixed this along with a couple other minor improvements:
- Updated .netkan to most recent spec/format
- Corrected license (now automatically pulls from this repo's metadata)

I did not make any changes to the dependencies or conflicting packages, only recommended/suggested. The following types of packages were set to be recommended by the core RP-1 package:
- All packages that are _required_ by the express install
- All packages that are _recommended_ by the express install
- The 3 packages that are described as "highly recommended" on the wiki: Principia, kOS, and Docking Port Sound FX.

All of the remaining mods on that wiki page were set to suggested. No other mods are set to recommended or suggested.

I believe there are two things that somebody could object to. First: the current netkan recommendations offer choices between TestFlight/TestLite, RCSBuildAid/RCSBuildAidContinued, and VenStockRevamp/Restock. I have elected to simply recommend only the option that is installed by the express installer. There is nothing stopping a user from simply installing a different option if they wish to do so, I think this simplification is for the better.
The second thing to object to is that I included Principia. I know it is not on ckan. I know it will probably never be on ckan. I think because it is "highly recommended" that it should be included. Doing so will not cause any errors during installation, it will simply serve as a reminder to install it should anybody bother to read the list of recommended mods in the sidebar of ckan.

Please let me know if any changes are requested, or if you disagree with where I drew the line between recommended and suggested. Keep in mind that this is not a change to the express installer at all, just the core package for advanced user installation. (And more importantly, this fixes what a user will see after manually running a recommendations audit after an express install).
I can also update the "RP-0" version of this config if so desired.


NathanKell:
I think we probably need to figure out whether we still want to Highly Recommend kOS. Also, I'm kinda concerned about the Principia dep. @HebaruSan are there any downsides to it? From a CKAN perspective--I assume the user-facing downside is that it may be confusing.

Apart from that, looks great! Thanks!!

HebaruSan:
> Also, I'm kinda concerned about the Principia dep. @HebaruSan are there any downsides to it? From a CKAN perspective--I assume the user-facing downside is that it may be confusing.

Hi @NathanKell, no downsides that I know of. As @Qazerowl says, it will be visible in the Relationships tab if someone is browsing the main mod list, but recommendations that can't be satisfied are hidden during installation.

(BTW, I did do some [further investigation](https://github.com/KSP-CKAN/CKAN/issues/1739#issuecomment-1694453233) trying to address the technical side of adding Principia to CKAN, and it's even uglier than I imagined. Nobody in the OS space seems to care about letting us query for an installed library conveniently. On Windows we either have to memorize a list of dozens of UUIDs to pass to an MSI-related API call or use an [outdated library](https://github.com/enclave-networks/MscVersion/) that does that for us. On Linux we can run `ldconfig -p`, but it doesn't include any versioning, so there's no way to be sure the user has the right version of `libc++.so.1` and `libc++abi.so.1`. Mac is the most straightforward since Principia just requires an overall OS version, which .NET/Mono provides.)

NathanKell:
We'd like to keep the any_of for restock vs Ven's, and TF vs TL, but apart from that I think this is good to merge! Sorry for the delay on this, got sucked into some RL stuff.

RP-1.netkan
```diff
@@ -1,120 +1,110 @@
...
+  - name: ToolbarController
+  - name: Kerbalism-Config-RO
+  - name: KSPCommunityFixes
+  - name: KerbalChangelog
+recommends:
+  - name: KSPBurst-Full
+  - any_of:
+    - name: ReStock
+    - name: VenStockRevamp
+  - any_of:
```
NathanKell, line 51: I think you only need one of these--copypasta?

Qazerowl: The next couple lines then offer the choice between ReStock **Plus** and Ven's. I wasn't sure of the best way to express that decision. But some testing reveals that during installation there's functionally no difference between that formatting and just recommending restock plus separately: the user has to toggle all 3 mods to switch to Ven's during the install regardless. I have simplified it.

NathanKell:
Looks good to me, thank you! Gonna merge this when we ship the next update of RP-1 since it will no longer depend on magicore.

PR #2206, opened 2023-08-31
Outer planet programs descriptions

SierraHotel84:
Updated program descriptions for:
-Jupiter Observation
-Saturn Observation
-Outer Gas Giant Survey
-Plutonian Landings


SierraHotel84:
Be gentle if I borked this up, first PR. Yay?!

NovemberOrWhatever:
The writing looks great, thank you so much for doing it! That must have been a lot of work. My only real concerns are the multi-paragraph descriptions, have you tested that they work like that? I might need to do a little bit of language cleanup after it's merged, but the writing overall is fantastic, thanks again. I couldn't have done it better myself

SierraHotel84:
That's....actually a great question.  I would have sworn I've seen some of the historical comments in other program descriptions but I don't see them anywhere else.  Let me load it up real quick and check.

PR #2208, opened 2023-08-31
Program Icons for Suborbital Development Suborbital Research and Uncr…

CXG-2827:
…ewed Lunar Surface Exploration


NathanKell:
Thanks for these!
* It'd be great if you could add some shading rather than flat-color, so they better match the existing icons
* The Uncrewed Lunar Surface icon I found a bit hard to figure out, maybe an icon based off of a landed Luna 9 ( https://en.wikipedia.org/wiki/Luna_9#/media/File:Luna-9_model.jpg ) or the Surveyor lander?

CXG-2827:
Sure, I'll add some gradient. Couldn't tell from other icons as it looked to be hard edges (Targeted Satellites and Geo).

Uncrewed lunar was an attempt to make Luna 16, but 48x48 pixels seems better for simpler icons 
https://upload.wikimedia.org/wikipedia/commons/f/f1/The_Soviet_Union_1970_CPA_3952_stamp_%28Luna_16_Leaving_Moon_%281970.09.20%29%29.jpg

I'll remake the icon with Luna 9.

From what I could tell, the icons are all based off a 3 color pallet, could you confirm?

NathanKell:
Ah! Yeah that makes sense for the probe lifting off, I didn't immediately recognize. The ground infrastructure was much harder to figure.

The icons definitely seem to have more than 3 colors, here's a closeup of the Heavy Sats one.
image
(They're certainly all grayscale, but seem to be at least 3bit) I'd say probably 4bit color and maybe even full 8bit alpha.

CXG-2827:
@NathanKell see below images. From my experimenting, the original logos are solid colors with a 0.3-0.4 pixel gaussian blur applied to the shapes to soften it and give it more shading. I was able to closely replicate the sputnik logo with this method:
Logo Example


I also tried a different approach using layer tweaks for the suborbital research logo to give it more depth. However, from my perspective it seems to clash with the styling of the existing ones:
example1

If you agree, I will move forward with using the gaussian blur method for the logos I am creating.

NathanKell:
I do agree! Thanks for digging in! :)

CXG-2827:
Icons have been updated.

PR #2211, opened 2023-09-04
Rework program progress, duration, funding to be incremental

### LINKED ISSUE #2207: Firing Milton Rosen strategically leads to more funding on ongoing programs, especially with funding peaks
It seems the calculation when you hire/fire Milton does not work properly during a program lifetime, especially for programs with peak funding.

For example, firing Milton after the Light Sat peak puts you back to before the peak.

image

And the recalculated funds clearly shows you are getting much more funding out of the program. (it thinks you got way less money than you actually did)

Same for evenly distributed programs, although less severe.

image

Hiring or Firing Milton should probably contract/expand the curve at the current position you are in the program timeline, not always at point 0 (start of the program).

NathanKell:
This PR fixes issues with Milt Rosen (and any other future leaders that might change program duration) by changing the program backend so that instead of computing everything based off (time - acceptedTime) it tracks the fraction elapsed. This means that if you take Rosen halfway through, your 20% faster will only affect the remaining 50% of program duration, and if you then fire him only the remaining duration will get extended back out. Taking or firing him after program deadline expiration still behaves a little strangely but this fixes the majority of the issues. Will close #2207

### DIFF
--- a/Source/Harmony/StrategySystem.cs
+++ b/Source/Harmony/StrategySystem.cs
@@ -184,7 +184,7 @@ namespace RP0.Harmony
                     UnityEngine.Debug.LogWarning($"[RP-0]: Warning: Program {p.name} does not have its strategy activated. Doing so now.");
                     if (s is ProgramStrategy ps)
                     {
-                        ps.PerformActivate(false, false);
+                        ps.PerformActivate(false);
                         ps.dateActivated = p.acceptedUT; // overwrite with correct activation UT
                     }
                 }
--- a/Source/Harmony/StrategySystem.cs
+++ b/Source/Harmony/StrategySystem.cs
@@ -187,9 +187,10 @@ namespace RP0.Harmony
                         ps.PerformActivate(false, false);
                         ps.dateActivated = p.acceptedUT; // overwrite with correct activation UT
                     }
-                    
                 }
             }
+
+            ProgramHandler.Instance.OnLoadStrategiesComplete();
 
             KerbalConstructionTime.KCTGameStates.RecalculateBuildRates();
         }
--- a/Source/Leaders/StrategyRP0.cs
+++ b/Source/Leaders/StrategyRP0.cs
@@ -117,7 +117,7 @@ namespace RP0
             if (!CanBeActivated(out _))
                 return false;
 
-            PerformActivate(true, true);
+            PerformActivate(true);
 
             
 
@@ -127,7 +127,7 @@ namespace RP0
         /// <summary>
         /// Teh
         /// </summary>
-        public void PerformActivate(bool useCurrency, bool recalcRates)
+        public void PerformActivate(bool useCurrency)
         {
             isActive = true;
             Register();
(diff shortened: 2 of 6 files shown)


Karhgath:
I tested the fix since I had a save just before firing Milton after two program peaks. It seems that on load, the program loaded as if Milton was not active. You can see the date, acceptance date and current green line on the graph on the first shot does not line up. However, the funding summary looks correct vs pre-fix. I included my save as well. The save should have both Crewed and Inner after the peak before firing Milton.

Then firing Milton seems to recalculate things properly and change the payout and timeline AFTER the current date (but the current progression of the contract was not correct). The deadline date of the program also did not change and got a weird negative budget for Year 2 - my current budget balance dropped to (95,067) but was also weirdly negative before firing Milton (probably due to the load issue) - (93,677). Beside Year 2 tho, the values seems to the the same are the pre-fix post-firing values. The year 2 negative value DOES balance the remaining values to ~1,800 (which is the diff before the current earning vs total funding - so remaining funding)

In my normal pre-fix game, I included the second set of screenshots as well of where I should be in the program timeline, my budget balance also was +27,517.

With the fix PR
image
image

Original situation with release version
image
image

Save game
[_miltonFire.zip](https://github.com/KSP-RO/RP-1/files/12548560/_miltonFire.zip)

NathanKell:
So you're trying this fix on a save that already had him appointed? Have you tried on a save without him appointed to see how that works, and/or a fresh save?
I'm trying to narrow down if it's an issue with how the save is upgraded, or with the code itself. Because for me it worked fine on a fresh save.

Karhgath:
Yeah, a save where he was already appointed. I have a save just before appointing him, I can test that as well. I assume the fresh save path was already tested. I'll test some more and report.

NathanKell:
Thanks! I'm hoping it's just a bug with how I'm backfilling data on programs (it's likely I was just assuming he wasn't appointed already, which is dumb :D ) and that shouldn't be too bad to fix. So if you confirm it's working in a live save without him currently active, then with your save I should be able to figure it out. Cheers!

Karhgath:
After more tests, loading in progress programs without Milton seems ok, balance was good, however hiring Milton with in progress program seems to calculate some things wrong, the current date kind of went back in time. I am thinking however that this happened pre-fix as well. However the funding calculations seems good all around (100% vs 80%)

Also the deadline date does not change between Milton and Non-Milton (hiring or firing). Other than that, seems good when loading a previous save without Milton active.

Initial Load
Left : Before Hiring Milton (loading with a program in progress)
Right : After Hiring Milton
image
image
image


Fast Forward 
Left: Before Firing Milton
Right: After firing Milton
image
image
image

NathanKell:
@Karhgath can you also upload the pre-Milton save? Thanks! (The pre-Milton, _pre trying this branch_ save, so I get the upgrade to run on it clean each time)

NathanKell:
Debugged.
* The fact deadline wasn't changing was cosmetic, had you left and reentered admin it would have (it is updated whenever funding is updated, but the game is paused in those KSC takeover UIs). Fixed by force-recalculating it on leader change.
* Leaders load after programs do. Fixed through some hijinks and now the back-compat works.
* The "going back in time" thing is because the "Year x: << Y funds >>" thing is in nominal years of duration, you're not actually going back in time, your position in the graph (and thus as a fraction of elapsed program time) is held constant but because the deadline is recalculated that number changes. Fixed by changing text to "Nominal Year: <<funds>>"

PR #2234, opened 2023-09-17
Make RP-1 description more clear with new system

Clayell:
The express install does not automatically select this install, but instead allows you to pick between this and legacy. This can lead to confusion, as seen in https://discord.com/channels/319857228905447436/512556137380315139/1152869006689181716.

### DIFF
--- a/RP-1.netkan
+++ b/RP-1.netkan
@@ -2,7 +2,7 @@
 	"spec_version" : "v1.26",
 	"identifier" : "RP-1",
 	"name"       : "Realistic Progression One (RP-1)",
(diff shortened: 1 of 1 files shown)


RP-1.netkan
Clayell, line 1: Unsure if this is necessary, but might be important

Clayell:
The express install instructions need to be fixed as well, for "legacy" it lists version v1.12.19.0, but the last legacy version was v1.13.2.2. I have no idea where and how to update this, I'll be asking for help in #moddev-help.

NathanKell:
Per discussion in discord, I'm pulling Legacy as an option in the express install so I don't think this is necessary in that case, it'll be very clear the choice is "career" or "no career". Cheers though! :)

Issue #2226, opened 2023-09-12
Tooling inconsistency


Codelet:
I have huge science core probe tooled, which generated for me a tooling SM-1 of size 0.8 x 0.8.
If I switch this probe to SM-2 the tooling cost is 12090 (and in tooling window it confirms the SM-2 diameter will be 0.8).
But if I create manually a SM-2 tank 0.8 x 0.8 the tooling cost is 6863.

This is rather suspicious, feels like avionics manufacturer tries to fool me  :)

NathanKell:
I believe @siimav was looking into tooling tech reducers in this regard.

siimav:
I was. But no conclusion was reached on how to handle the reducers so I shelved that idea.

NathanKell:
#2236

PR #2236, opened 2023-09-18
Allow only EC to be added inside procedural avionics.

siimav:
* Avionics tooling is just tonnage and tank size. There is no additional SM or battery tooling any longer. Previously you were forced to tool some SM diameter along with avionics. This means that upgrading from one battery level to another doesn't require any tooling either.
* No spherical tank volume hacks and no 50% penalty for calculating usable volume inside the avionics part.
* Battery tank definitions are just copies of SM I to SM IV but can have no resources other than EC.
* Battery levels are upgraded with electronics blue sky nodes instead of mat sci

### Things to consider / figure out:
* Do we want more battery levels? Current progression ends with Apollo. Although by the looks of it SM-IV is already in the ballpark of modern Li-Ion batteries.
* We have EC/kg progression but nothing for EC/l. 1 liter of volume always contains 1000EC. Do we want that? How should it be implemented (utilization limits, RF internal tank density)?
* SM tanks can have EC too and I doubt we want to remove that option. So if we want more battery levels then there's a mismatch between EC stored in SM tanks and EC stored in avionics.

Depends on https://github.com/KSP-RO/RealismOverhaul/pull/2915


NathanKell:
Closes #2226

PR #2238, opened 2023-09-19
Earth Observation/Commercial Applications Programs

SierraHotel84:
Added commercial applications program with contracts, disabled EOS2/CA2

Cherenkov Detector created, placed, and configured

<s>-Created placeholder EOS2, Commercial Apps, Commercial Apps Advanced programs</s>
<s>-Moved all com/nav sat contracts out of Targeted Sats</s>
<s>-Renamed Targeted Sats to EOS1</s>
<s>Included Corona and Orbital Perturbation for now, can update these later</s>
<s>-Working on adding Commercial Apps 1</s>
<s>Need to test a couple of minor changes in the optional contracts</s>
<s>Have given rep values only a cursory look so far, need to go over those a little closer</s>
<s>Sort keys are all over the place</s> Removed


NathanKell:
Awesome!! Timing couldn't be better, I'm wrapping up work on #2231 and would love to ship this with it, it makes sense!

Quick suggestion: please all old programs like the old Crewed Orbit program, where you set isDisabled=true and then in the contract use an ANY for the program requirements to target either the old program or the new one. This is because otherwise it'll 100% break saves and we'd really prefer not to do that. :)

Alternative to using the 'any' approach is to just duplicate the contracts. That approach might well make sense because you'll probably also want to be making _changes_ to the contracts, too.

SierraHotel84:
Easy fix. Wanted to get everything I have uploaded as I'm working on it in 2 places.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

SierraHotel84:
> Alternative to using the 'any' approach is to just duplicate the contracts. That approach might well make sense because you'll probably also want to be making _changes_ to the contracts, too.

Are you talking about the way the Earth Satellite - Heavy program was made by using the original ES contracts, but adding Heavy to adjust them? Leaving it with the ANY for now until I can take a look at the individual contracts in CA, may switch it to this if they get updated.

NathanKell:
Yeah it gives you much more freedom to do what you like with the contracts if you don't have to worry about changes to them applying to (deprecated, but still in use in release) programs or, worse, live existing contracts that folks have accepted (where IIRC parameter changes don't take effect, _but requirements changes do_ so contracts can silently fail on folks).

So long as the contracts stay unaltered though the Any route should work fine! As it did for the crewed orbit rework.

SierraHotel84:
> So long as the contracts stay unaltered though the Any route should work fine! As it did for the crewed orbit rework.

Missed this key point.  Does the reused contract have to stay 100% unaltered, or will adding in an Any to the requirements with the new program allow it keep functioning?

SierraHotel84:
Added in the Cherenkov detector. In the Configure Experiments window in the PAW, everything shows up correctly, however in the Tech Tree, it's missing the entire block that should have the data rate, EC requirements, situations, etc. 

paw

tech tree

NathanKell:
Quick note, please include both the changed json file(s) as well as the generated cfg file(s), as otherwise the next time someone generates cfg your changes will be lost.

SierraHotel84:
Helps to read the entire instructions.  Should be good now.

GameData/RP-1/Science/Experiments/Cherenkov.cfg
```diff
@@ -0,0 +1,144 @@
...
+	@MODULE[DMModuleScienceAnimateGeneric]
+	{
+		@totalScienceLevel = 0.375
+	}
+}
+
+// ============================================================================
+// Add experiment to parts
+// ============================================================================
+@PART[RP0Cherenkov]:BEFORE[RP-0-Kerbalism]
```
NathanKell, line 125: This block shouldn't be here--in the original cosmic ray file, it's removing cosmic ray 1 from the cosmic ray 2 part, and adding a different module. But in your case you only have 1 tier, so I think you just nuke this whole @PART node. Honestly this might be broken for cosmic ray 2 as well, I gotta check. Honestly this whole thing is kind of a mess, because original ROKerbalism handled the science stuff, but it got merged into RP-1, but when it was merged instead of unifying the patches _both sets of patches_ were left around, which means half this file is adding stuff and the other half is removing it and replacing it with _other_ stuff, when we should just be adding the final result.

NathanKell:
It also helps if _I_ read all the messages above! Sorry!

GameData/RP-1/Science/Experiments/Cherenkov.cfg
NathanKell: Checked and cosmic ray 1 and cosmic ray 2 are added differently (!) in RP-1. And the issue you hit is the above, you kept the adding that was done with cosmic ray 1 (depending on first adding the DP module) but _also_ kept something associated with cosmic ray 2 (nuking that module when found). I'm sorry that file is such a mess!

SierraHotel84: Just glad I wasn't the only one confused by that!

GameData/RP-1/Contracts/Earth Observation 1/EOSGrav1.cfg
```diff
@@ -0,0 +1,121 @@
...
+				completionText = Stable orbit: Confirmed
+			}
+			
+			PARAMETER
+			{
+				name = RP1CollectScience
+				type = RP1CollectScience
+				targetBody = HomeWorld()
+				situation = InSpaceLow
+				fractionComplete = 0.0083
```
NathanKell, line 115: Like with the others, probably want to up this to something like 2 months?

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1.cfg
```diff
@@ -0,0 +1,128 @@
...
+				completionText = Stable orbit: Confirmed
+			}
+		
+			PARAMETER
+			{
+				name = RP1CollectScience
+				type = RP1CollectScience
+				targetBody = HomeWorld()
+				situation = InSpaceLow
+				experiment = RP0infraredRad1
```
NathanKell, line 113: This needs some kind of frac?

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1.cfg
```diff
@@ -0,0 +1,128 @@
...
+				title = Transmit infrared weather data from low orbit
+			}
+			PARAMETER
+			{
+				name = RP1CollectScience
+				type = RP1CollectScience
+				targetBody = HomeWorld()
+				situation = InSpaceLow
+				experiment = RP0visibleImaging1
+				fractionComplete = 1
```
NathanKell, line 123: I don't think you want 1 here, especially for a non-polar contract. This is biome-dependent so 1 would mean "complete data from all biomes", including (a) poles, and (b) shores, which are super tough to get at warp. You probably want something pretty low here and to more rely on the IR one.

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1Rep.cfg
```diff
@@ -0,0 +1,160 @@
...
+				minQuantity = 1.0
+				title = Craft must have ElectricCharge after one day
+				hideChildren = true
+			}
+		}
+		PARAMETER
+		{
+			name = Orbit
+			type = Orbit
+			minInclination = Round(Random(30, 80))
```
NathanKell, line 114: This probably should just be based on completing IR Rad in space low? Otherwise it's just "launch a thing with no payload to a random orbit" which doesn't feel like it fits the new program. If it's just "get max 80%" of IR rad though then there's only one go at this contract, so not sure what else to add.

GameData/RP-1/Contracts/Earth Observation 1/EOSMagScan1.cfg
```diff
@@ -0,0 +1,127 @@
...
+				minQuantity = 1.0
+				title = Craft must have ElectricCharge after one day
+				hideChildren = true
+			}
+
+			PARAMETER
+			{
+				name = Orbit
+				type = Orbit
+				minApA = 36000000
```
NathanKell, line 90: Same deal as Cherenkov, needs a max Pe (and probably a max Ap as well, something like 250Mm ?)

GameData/RP-1/Contracts/Earth Observation 1/EOSMagScan1.cfg
```diff
@@ -0,0 +1,127 @@
...
+				completionText = Stable orbit: Confirmed
+			}
+			
+			PARAMETER
+			{
+				name = RP1CollectScience
+				type = RP1CollectScience
+				targetBody = HomeWorld()
+				situation = InSpaceLow
+				fractionComplete = 0.03
```
NathanKell, line 111: Same deal as the others, need the min to be higher.

GameData/RP-1/Contracts/Earth Satellites/FirstMolniyaSat.cfg
```diff
@@ -7,7 +7,7 @@ CONTRACT_TYPE
...
+	description = <b>Program: Early Commercial Applications<b> or <b>Targeted Satellites<br>Type: <color=green>Required</color></b><br><br>A Molniya orbit is a type of highly elliptical orbit with an inclination of 63.4 degrees, an argument of perigee of -90 degrees, and an orbital period of one half of a sidereal day. Place a satellite into a Molniya orbit.&br;&br;There will be an orbit showing in the map view and tracking station. You do not have to place the satellite into this orbit, but it is instead meant to show you where your perigee and apogee should approximately be located, but more importantly, it shows the proper shape of the orbit.
```
NathanKell, line 10: But CA now includes its own FirstMolniyaSat ...?

GameData/RP-1/Contracts/Earth Satellites/FirstTundraSat.cfg
```diff
@@ -7,7 +7,7 @@ CONTRACT_TYPE
```
NathanKell, line 7: I think doing anything with Tundra should wait for Comm App 2, it doesn't fit in Comm App 1.

GameData/RP-1/Parts/Science/CR01_5-CherenkovCounter.cfg
```diff
@@ -0,0 +1,52 @@
...
+	}
+	scale = 1.0
+	rescaleFactor = 1.2
+	node_attach = 0.0, 0.0, 0.0, 0.0, -1.0 , 0.0
+	TechRequired = basicScience
+	entryCost = 0
+	cost = 250
+	category = Science
+	subcategory = 0
+	title = CR01.5 - Cherenkov Cosmic Ray Detector
```
NathanKell, line 31: This should just be CD01 I think, Cherenkov Detector, because it's not actually in the line of cosmic ray detectors, it's a separate instrument for us

GameData/RP-1/Parts/Science/CR01_5-CherenkovCounter.cfg
```diff
@@ -0,0 +1,52 @@
...
+	PhysicsSignificance = 1
+	bulkheadProfiles = srf
+	
+	tags = cosmic radiation scanner rad science detector muller
+
+}
+
+@PART[RP0Cherenkov]:NEEDS[!Kerbalism]:AFTER[RP-0]
+{
+	@description ^=:$: \n\nLevel 1.5 Cosmic Ray Experiment Device. Can gather 100% of Cosmic Ray Science.
```
NathanKell, line 51: I wouldn't actually let it gather Cosmic Ray science tho?

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -287,11 +288,109 @@ RP0_PROGRAM
...
+	REQUIREMENTS
+	{
+		ANY
+		{
+			complete_program = EarlySatellites
+			complete_program = EarlySatellites-Heavy
+		}
+  		ANY
+    		{
+      			active_program = CrewedOrbit
```
NathanKell, line 313: Uh, why does CrewedOrbit need to be active or completed?

GameData/RP-1/Science/Experiments/Cherenkov.cfg
```diff
@@ -0,0 +1,134 @@
...
+		experimentsLimit = 1
+		totalScienceLevel = 1.0
+		asteroidReports = false
+	}
+}
+
+
+//	COSMIC RAY LEVEL 1.5
+//	Cherenkov Detector
+//	Based on the Geiger-Muller Counters used on many spacecraft.
```
NathanKell, line 110: Yeah I'm not seeing where it actually includes the previous level here? So the comment is wrong? (Per above it's good it doesn't include any cosmic ray levels)

NathanKell (changes requested):
Finally had a chance to actually review this! Thanks again for putting this together.

Note you can press `.` (period) to open a much better editor than the old web editor, that should make your life easier if you don't want to have a local client.

Lastly, it'd be awesome if (like a bunch of the historic "first" contracts we have in RP-1) you could mention in the contract description what the contract was based off of? Like Explorer 6, the Transit network, Telsar, etc. But no worries if that's a hassle, that's just the cherry on top. :)

NathanKell:
Also tagging @ppboyle to check the numbers :)

SierraHotel84:
Easy fixes, will take care of them tonight.

Didn't know about the . thing, will definitely check that out.  Doesn't really matter on this laptop but will definitely make it easier on my work one.

Nope, thanks for the reminder on the historicals!  Meant to go in and make sure those are formatted the same across all the contracts.
________________________________

NathanKell:
Sweet, thanks!!
Also I took the liberty of editing your comment since GH didn't trim the quoted email :D

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -287,11 +288,109 @@ RP0_PROGRAM
...
+}
+
+RP0_PROGRAM
+{
+	name = EarlyCommApp
+	title = Early Commercial Satellites
+	description = Although it was through government programs that the earliest satellites were launched, commercial use of them was always envisioned as well.  Through this program, the earliest iterations of commercial-use satelites will be launched.  With wildly varying needs, each satellite is dedicated to a specific requirement, typically far more focused than the earlier broad scientific satellites. 
+	requirementsPrettyText = Complete Early Satellites program
+	objectivesPrettyText = Launch the first commercial-use satellites and build a satellite communication network. 
+	nominalDurationYears = 2
```
NathanKell, line 350: With the two networks now this is at least 3 years I think. It's possible both should be 4 but I'm unsure, I'll let Max speak more authoritatively :D

ppboyle: Agreed on 3 years.

ppboyle:
Hi, a few notes:

1. The easier contracts in EOS should be 40 rep - that'd be IRRad, Cherenkov, and CosmicRay, as well as their repeatables (IRRadRepeatable is weirdly low at 20 currently)
2. Corona I'd leave at 80
3. Grav scan I'd set to 60 - since its a lighter craft requirement than the previous, comparable GEO comm sat.
4. There is NO anti-grind implemented on the EOS repeatbles - there 100% must be.

I'm concerned the player can dovetail many of these with relatively easy on-orbit maneuvers. We might need to add specific payload requirements to try to prevent that - eg: A specific amount of a new science sat payload type (eg - require 40-45 for this contract, 50-55 for that contract, etc)  this both makes the contracts a little more challenging, and helps isolate them from each other.  Between that and removing the requirement to put something in GEO, I think this program is now too easy for 3 years and 430k, as potentially it can be done in 2-3 launches.

Also, I think if you split your EOS contracts into two vesselparametergroups - the first group containing a "define = Foo", the new vessel requirement and an initial orbit requirement, and the 2nd vesselparametergroup that has "vessel = Foo", the requirements for the time on orbit and the experiment transmit, you won't have the issue where the contract pieces can be completed by any vessel.   



No big concerns with CA as long as those contracts have mirrored the previous versions, other than maybe giving it 3 years.  You can use the same program tuning from the EOS when you bump it t 3 years.

NathanKell:
Per discussion it seems like longer orbit duration requirements might go a long ways towards preventing mission-combining abuse. Something on the order of a month.

Also probably want to remove the solar/EC requirements from com/nav contracts given #2249 ?

Re: repeatables, for EOS since the "repeatables" are (or will be) less "repeatable" than "optional Extended Mission", they probably need to be priced accordingly? That is the idea being the base versions of the contracts set a given experiment completion threshold, and then the optional follow-on makes that requirement 100% (or, in the case of biome science, maybe 80-90%?)
Grav is weird because of the duration tho. Also because of the spammability.

GameData/RP-1/Contracts/Commercial Applications 1/EarlyNavSat-CA.cfg
```diff
@@ -0,0 +1,158 @@
...
+			hideChildren = true
+		}
+		PARAMETER
+		{
+			name = Orbit
+			type = Orbit
+			minInclination = Round(Random(30, 90))
+			minPeA = Round(Random(300000, 400000), 1000)
+			minApA = Round(Random(500000, 1000000), 1000)
+			maxApA = Round(Random(1000000, 3000000), 1000)
```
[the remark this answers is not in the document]
SierraHotel84: Will it accept functions?  Can I do minApA + 50000 or just simplify it at 1050000?

GameData/RP-1/Contracts/Earth Satellites/FirstMolniyaSat.cfg
SierraHotel84: Oh, I see, added that before I created the new one.

GameData/RP-1/Contracts/Earth Satellites/FirstTundraSat.cfg
SierraHotel84: That's the original program, didn't make any changes to it. 

Should we maybe add an exclusion to the new programs so that if someone as already accepted the original Earth Satellites they won't get the new EOS/CA?

GameData/RP-1/Parts/Science/CR01_5-CherenkovCounter.cfg
SierraHotel84: Seems so obvious now...

GameData/RP-1/Programs/Programs.cfg
SierraHotel84: Because of the Corona contracts needing the first_OrbitReturn (or whatever the actual name is) completed.  I figured this way it would show up in the program screen as being needed, instead of being a surprise when someone accepts the contract and then find out they needed that contract done.

GameData/RP-1/Contracts/Commercial Applications 1/EarlyNavSat-CA.cfg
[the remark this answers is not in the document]
Teykn: > Will it accept functions? Can I do minApA + 50000 or just simplify it at 1050000?

Not 100% in the case above, though seems to be possible according to [the CC wiki for Expressions](https://github.com/jrossignol/ContractConfigurator/wiki/Expressions) Might be easier to put those values into `DATA` nodes and use an @. 

Example from OrbitFlight2.cfg:
`endApA = @startApA + Round(Random(150000, 800000), 25000)`

GameData/RP-1/Programs/Programs.cfg
NathanKell: I think it's a much better approach to make the first orbit return contract depend on either EOS1 or Crewed Early (in the same way you changed the commsat contracts), and make completing it required for both programs. You'll only have to fly it once of course, but this solves either doing Corona before that contract, or having EOS1 gated on Crewed Orbit.

GameData/RP-1/Contracts/Earth Satellites/FirstTundraSat.cfg
NathanKell: Except the changes shown here involve changing the description to mention it's from either program (Targeted or CA1), the requirements now have an ANY for either of those programs (and an ANY for its prereq contract) and some textual changes. I would have expected there to be no changes in the file since the old program has all its own contracts untouched, so no changes are needed here to check for completions of the other contracts.

You're very right however that EOS and CA should not be allowed if you've already taken Targeted, and Targeted or Comm Net respectively. Don't worry about that part however, since I have to do that in code (and in a somewhat ugly way, at that!)

NathanKell:
Note that https://github.com/KSP-RO/ContractConfigurator/pull/26 will let us check for experiments by ID rather than only being able to check for specific parts. :)

GameData/RP-1/Contracts/Earth Satellites/FirstTundraSat.cfg
SierraHotel84: > Except the changes shown here involve changing the description to mention it's from either program (Targeted or CA1), the requirements now have an ANY for either of those programs (and an ANY for its prereq contract) and some textual changes. I would have expected there to be no changes in the file since the old program has all its own contracts untouched, so no changes are needed here to check for completions of the other contracts.


Oh yeah that.

SierraHotel84:
>     4. There is NO anti-grind implemented on the Rad repeatable and there should be given that it has a 7 completion limit.  Corona is probably ok due to its inherent long duration.  tbh not sure how it makes any sense if you finish the experiment.
> 

Going to need some help with this part, tried adding it into the EarlyNavSat contracts.  I did a copy/paste from EarlyComSat, which is verified to work in-game.  Tweaked all the comsats to be navsats, and cannot get it to load, it fails the contract configurator checks.

GameData/RP-1/Contracts/Commercial Applications 1/EarlyNavSat-CA.cfg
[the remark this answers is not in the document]
NathanKell: I believe you can also self-reference there if you like, because certainly we do it all the time in `title` blocks where we reference a value in the node itself. But @Teykn 's approach is I think cleaner since it puts all the 'mission-critical' variables at the top in one place, rather than strewn throughout the parameters.

NathanKell:
> Going to need some help with this part, tried adding it into the EarlyNavSat contracts. I did a copy/paste from EarlyComSat, which is verified to work in-game. Replaced all the comsats with navsats, and cannot get it to load, it fails the contract configurator checks.

I can take a look tomorrow if you pastebin the changed contract somewhere, but @ppboyle or @siimav I think are better at it than I am (they've either done it, or fixed my attempts :D )

GameData/RP-1/Contracts/Earth Observation 1/EOSCorona.cfg
```diff
@@ -0,0 +1,141 @@
...
+			{
+				name = PartValidation
+				type = PartValidation
+				
+				VALIDATE
+				{
+					MODULE
+					{
+						name = Experiment
+						experimentID = RP0photos2
```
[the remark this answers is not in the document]
SierraHotel84: Even though we have photos1 in the sounding era for the V2s?  I thought this was suppsoed to mimic early Corona sats, hence the name?

GameData/RP-1/Contracts/Earth Observation 1/EOSCosmicRay1.cfg
```diff
@@ -0,0 +1,128 @@
...
+			maxCrew = 0
+			title = Uncrewed
+			hideChildren = true
+		}
+
+		PARAMETER
+		{
+			name = OrbitSequence
+			type = Sequence
+			title = Survive in orbit for a day and transmit science
```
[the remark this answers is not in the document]
SierraHotel84: Yeah, didn't realize when I first started these that was a date range, thought it was somehow tied to the duration param

GameData/RP-1/Contracts/Earth Observation 1/EOSCorona.cfg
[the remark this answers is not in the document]
NathanKell: Oh huh. For some reason I thought this was the one in AdvCapsules science along with cosmic ray 2, but I think that change never got made (it was mooted years ago, and then reintroduced for Zephyr's PR?)

I think until that PR actually gets done I gotta do some balance on the part and the experiment to make it viable to fly these on Thor-class LVs with ~1960-1962 tech.

GameData/RP-1/Contracts/Earth Observation 1/EOSCosmicRay1.cfg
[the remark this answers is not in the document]
NathanKell: It maybe could be? But I don't know how you'd do that, alas.

NathanKell:
Any luck with (or any deets on) the soft cooldown issue?

Also the photo experiment to use is now photo 1a not photo2. :) (The optional should rely on photo2 still I think). Note both of them are now short-duration experiments so you should expect to get ~50% of the science for the required contract and then 30, 60, 90 of photo2 for some optionals?

SierraHotel84:
>     4. There is NO anti-grind implemented on the Rad repeatable and there should be given that it has a 7 completion limit.  Corona is probably ok due to its inherent long duration.  tbh not sure how it makes any sense if you finish the experiment.

Going to need some help with this one.  I tried adding it in with a copy/past 

> Any luck with (or any deets on) the soft cooldown issue?
> 
> Also the photo experiment to use is now photo 1a not photo2. :) (The optional should rely on photo2 still I think). Note both of them are now short-duration experiments so you should expect to get ~50% of the science for the required contract and then 30, 60, 90 of photo2 for some optionals?

Sorry, didn't have cell service most of the day.  Trying one more thing on the anti-grind real quick.

Sounds good on the photo1a and photo2.

SierraHotel84:
Did a comparison in VSCode and found no difference between the two anti-grind portions besides changing out Com for Nav.  At a loss.

https://pastebin.com/GNS8cPNN

Copy/pasted the requisite portions of both contracts into one file, though they're both updated in my most recent commit

SierraHotel84:
Just missing the anti-grind in EarlyNav and EarlyMolniya, and finish up the tweaks to the Corona contracts and this will be all wrapped up.

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1.cfg
```diff
@@ -117,3 +118,258 @@ CONTRACT_TYPE
 		}
 	}
 }
+
+CONTRACT_TYPE
```
SierraHotel84, line 122: Thoughts on doing 2 follow-on contracts like this and dumping the repeatable one?  Gets rid of the "launch more satellites for reasons" per one of your other comments, adds inclination to get more science for each one, and lowers rep so that the science is the goal.   Also mirrors the real launches since TIROS-9 was polar.

SierraHotel84:
> I'm concerned the player can dovetail many of these with relatively easy on-orbit maneuvers. We might need to add specific payload requirements to try to prevent that - eg: A specific amount of a new science sat payload type (eg - require 40-45 for this contract, 50-55 for that contract, etc) this both makes the contracts a little more challenging, and helps isolate them from each other. Between that and removing the requirement to put something in GEO, I think this program is now too easy for 3 years and 430k, as potentially it can be done in 2-3 launches.

Couple of things coming to mind on this.  I just checked over the orbital requirements for all of the EOS contracts and I come up with it requiring 6 launches after removing the IR repeatable and making it into 3 distinct launches per my comment below, with the 2nd Corona still an optional one.  The only ones I really see that could be combined are cosmic ray, cherenkov, and mag scan, which kind of makes sense.  With a minor tweak the gravity one would be a standalone and couldn't be combined with the IR ones. 

To your comment about requiring payload, plenty of examples of contracts requiring a certain amount of X payload, is it possible to set an upper limit on that?  Something like a minimum of 50 units of X on A contract, but a maximum of 40 units on B contract?

SierraHotel84:
@ppboyle @NathanKell 

Regarding antiGrind, I figured out the issue.  RP0expectedDays_ is the culprit.  Running down some more details to make an issue because it's a little more widespread than just this program, but that's the root cause.

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1.cfg
NathanKell: Yep I think it makes sense to swap to fixed extras rather than infinite repeatables. Regarding rewards, generally optionals should have higher rep rewards than requireds, and my point about science as the goal was more that science was the goal _of the program_, rather than the goal of the player (who doesn't need a contract to gather science). Having science be the focus of the program contrasts with, say, CA1 (or even Crewed Orbit), where the goal is to Do Stuff In Space and science is at best a byproduct.

NathanKell:
> > I'm concerned the player can dovetail many of these with relatively easy on-orbit maneuvers. We might need to add specific payload requirements to try to prevent that - eg: A specific amount of a new science sat payload type (eg - require 40-45 for this contract, 50-55 for that contract, etc) this both makes the contracts a little more challenging, and helps isolate them from each other. Between that and removing the requirement to put something in GEO, I think this program is now too easy for 3 years and 430k, as potentially it can be done in 2-3 launches.
> 
> Couple of things coming to mind on this. I just checked over the orbital requirements for all of the EOS contracts and I come up with it requiring 6 launches after removing the IR repeatable and making it into 3 distinct launches per my comment above, with the 2nd Corona still an optional one. The only ones I really see that could be combined are cosmic ray, cherenkov, and mag scan, which kind of makes sense. With a minor tweak the gravity one would be a standalone and couldn't be combined with the IR ones.

6 distinct launches works for requireds. Less than that for required (or 6 total including optionals) feels very light tbh.

> To your comment about requiring payload, plenty of examples of contracts requiring a certain amount of X payload, is it possible to set an upper limit on that? Something like a minimum of 50 units of X on A contract, but a maximum of 40 units on B contract?

Yes that's certainly doable, though I think in practice (for the player) it would feel very strange. My hunch is that duration-in-given-orbit requirements will be sufficient.

SierraHotel84:
> 6 distinct launches works for requireds. Less than that for required (or 6 total including optionals) feels very light tbh.

6 required, with optional 7th as it currently stands.  I'll adjust the gravity one to make it 8 and leave that one optional as it is now. 

> Yes that's certainly doable, though I think in practice (for the player) it would feel very strange. My hunch is that duration-in-given-orbit requirements will be sufficient.

Makes sense, was more one of those "I wonder" kind of things as I was thinking through it.

GameData/RP-1/Contracts/Commercial Applications 1/EarlyComSat-CA.cfg
```diff
@@ -133,9 +133,9 @@ CONTRACT_TYPE
 			name = Orbit
 			type = Orbit
 			minInclination = Round(Random(30, 31))
-			minPeA = Round(Random(700000, 1000000), 1000)
-			minApA = Round(Random(3000000, 10000000), 1000)
-			maxApA = Round(Random(10000000, 30000000), 1000)
+			minPeA = Round(Random(700000, 1000000), 10000)
+			minApA = Round(Random(3000000, 10000000), 10000)
```
NathanKell, line 137: This still feels like a ton of variation? My hunch is we might not actually want variation here, per other limit-variation stuff @ppboyle has done. If there is variation, and since there's no decline penalty, the player can just keep declining until they get what they want. If we want variance we could vary things like LAN, or payload (using the payload scaling logic in the old GEO repeatable?)...not sure what else?

GameData/RP-1/Contracts/Commercial Applications 1/EarlyComSat-CA.cfg
```diff
@@ -144,7 +144,7 @@ CONTRACT_TYPE
 				name = Duration
 				type = Duration
 
-				duration = 2m
+				duration = 30d
```
NathanKell, line 147: Since commercial sats are destroyed on contract complete there's no need for more duration than 2m. Sorry I wasn't clear, I meant to pump up the duration on the EOS side, on the CA side 2m stability checks are fine.

GameData/RP-1/Contracts/Commercial Applications 1/EarlyNavNetwork3.cfg
```diff
@@ -95,7 +91,7 @@ CONTRACT_TYPE
 			name = HasPower
 			type = HasResource
 			resource = ElectricCharge
-			minQuantity = 1.0
+			minQuantity = 2d
```
NathanKell, line 94: That's a duration being specified for a quantity param?

GameData/RP-1/Contracts/Commercial Applications 1/EarlyNavNetwork3.cfg
```diff
@@ -126,23 +122,21 @@ CONTRACT_TYPE
...
-
 			vessel = NavSat III
 		}
 		PARAMETER
 		{
 			name = HasPower
 			type = HasResource
 			resource = ElectricCharge
-			minQuantity = 1.0
+			minQuantity = 2d
```
NathanKell, line 139: ditto. I'm surprised the contract loads?

GameData/RP-1/Contracts/Commercial Applications 1/EarlyNavNetwork3.cfg
```diff
@@ -223,8 +215,35 @@ CONTRACT_TYPE
...
+		type = DialogBox
+		DIALOG_BOX
+		{
+			title = Vessel Ownership Transferred
+			condition = CONTRACT_SUCCESS
+			position = CENTER
+			width = 0.5
+			TEXT
+			{
+				text = The contract has been completed successfully and the satellite has been transferred to the customer.
```
NathanKell, line 245: Minor, but 'satellites' not 'satellite'

GameData/RP-1/Contracts/Commercial Applications 1/EarlyNavSat-CA.cfg
```diff
@@ -122,7 +152,7 @@ CONTRACT_TYPE
 				name = Duration
 				type = Duration
 
-				duration = 2m
+				duration = 30d
```
NathanKell, line 149: as above re: durations

NathanKell (changes requested):

GameData/RP-1/Contracts/Earth Observation 1/EOSCherenkov.cfg
```diff
@@ -73,23 +73,14 @@ CONTRACT_TYPE
...
-				minQuantity = 1.0
-				title = Craft must have ElectricCharge after one day
-				hideChildren = true
-			}
-
 			PARAMETER
 			{
 				name = Orbit
 				type = Orbit
 				minEccentricity = 0.085
```
NathanKell, line 80: If this is supposed to be elliptical (and go into the Van Allen belts) then the min ecc probably needs to be about 0.15 and the max more like 0.3 or so?

NathanKell: Wait a minute, looking down later the science requirement is for space high, which means the min ecc has to be _way_ higher if the max Pe is 500km. You have to be able to _reach_ into spacehigh from a 200-500km Pe...

GameData/RP-1/Contracts/Earth Observation 1/EOSCherenkov.cfg
```diff
@@ -73,23 +73,14 @@ CONTRACT_TYPE
 			type = Sequence
 			title = Survive in orbit for a day and transmit science
 
-			PARAMETER
-			{
```
NathanKell: Sorry, it's been a bit. Why are the EC requirements being removed from these?

GameData/RP-1/Contracts/Commercial Applications 1/EarlyComSat-CA.cfg
SierraHotel84: Easy change.

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1.cfg
```diff
@@ -120,7 +111,7 @@ CONTRACT_TYPE
 				targetBody = HomeWorld()
 				situation = InSpaceLow
 				experiment = RP0visibleImaging1
-				fractionComplete = 1
+				fractionComplete = .1
```
NathanKell, line 115: I think it's worth having this at 0.5 as well

GameData/RP-1/Contracts/Commercial Applications 1/EarlyNavNetwork3.cfg
SierraHotel84: I didn't ask the right question on a previous iteration of the EC requirement and should have clarified it then.  Realized the better way was the power generation check anyway.  Even with the short time requirement, it made sense to me to keep it on those since it would be required in actuality.

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -308,13 +308,7 @@ RP0_PROGRAM
...
 			complete_program = EarlySatellites-Heavy
 		}
-  		ANY
-    		{
-      			active_program = CrewedOrbit
-	 		active_program = CrewedOrbitEarly
-    			complete_program = CrewedOrbit
-       			complete_program = CrewedOrbitEarly
-	  	}
+		complete_contract = first_OrbitRecover
```
NathanKell, line 311: Uh, this is putting it as a requirement not an objective.

NathanKell (changes requested):

GameData/RP-1/Contracts/Earth Observation 1/EOSCherenkov.cfg
SierraHotel84: My logic was that since we have the science requirement, they'll drive needing EC

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1.cfg
```diff
@@ -80,8 +80,9 @@ CONTRACT_TYPE
 				maxApA = 1000000
 				minPeA = 500000
 				maxEccentricity = 0.005
+				maxInclination = 45
 				disableOnStateChange = true
-				title = Achieve orbit between 500km and 1,000km in altitude with no more than .005 eccentricity.
+				title = Achieve orbit between 500km and 1,000km in altitude with no more than .005 eccentricity and inclination lower than 50 degrees.
```
NathanKell, line 85: I don't think I'd suggest having a max inclination here? It makes it basically impossible to complete when launching from high latitudes. I think it's fine for later contracts to have _minimum_ inclination requirements, but a max requirement seems off to me.

Also the number and the text don't match. ;)

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1.cfg
```diff
@@ -117,3 +118,258 @@ CONTRACT_TYPE
...
+			title = Survive in orbit for 2 days and transmit science
+
+			PARAMETER
+			{
+				name = Orbit
+				type = Orbit
+				maxApA = 1000000
+				minPeA = 500000
+				maxEccentricity = 0.005
+				maxInclination = 75
```
NathanKell, line 210: same with max vs min

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1.cfg
```diff
@@ -117,3 +118,258 @@ CONTRACT_TYPE
...
+				disableOnStateChange = true
+				title = Achieve orbit between 500km and 1,000km in altitude with no more than .005 eccentricity and inclination lower than 75 degrees.
+			}
+
+			PARAMETER
+			{
+				name = Duration
+				type = Duration
+				duration = 2d
```
NathanKell, line 219: See previous review re: durations better long on EOS contracts and short on CA contracts.

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1.cfg
```diff
@@ -117,3 +118,258 @@ CONTRACT_TYPE
...
+				fractionComplete = .75
+				title = Transmit 75% of infrared weather data from low orbit
+			}
+			PARAMETER
+			{
+				name = RP1CollectScience
+				type = RP1CollectScience
+				targetBody = HomeWorld()
+				situation = InSpaceLow
+				experiment = RP0visibleImaging1
```
NathanKell, line 241: Since I suggested bumping previous contract to 0.5, I'd move this to 0.7

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1.cfg
```diff
@@ -117,3 +118,258 @@ CONTRACT_TYPE
...
+				title = Transmit 100% of infrared weather data from low orbit
+			}
+			PARAMETER
+			{
+				name = RP1CollectScience
+				type = RP1CollectScience
+				targetBody = HomeWorld()
+				situation = InSpaceLow
+				experiment = RP0visibleImaging1
+				fractionComplete = 1
```
NathanKell, line 370: Since this is biome based, probably best to stop at 0.9 instead of 1.0

NathanKell (changes requested):

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1Rep.cfg
```diff
@@ -0,0 +1,160 @@
+CONTRACT_TYPE
```
NathanKell, line 1: I thought the Rad2 and 3 contracts replaced this? But this is still here?

GameData/RP-1/Contracts/Commercial Applications 1/EarlyComSat-CA.cfg
SierraHotel84: Funny enough that was lifted straight out of the original EarlyComSat.  This is in the geo repeatable: payload = RP1CommsPayload().  I added back in the random payload from the original, but I'm fine with whichever on it. Tightening up the orbit variation but adding in a random LAN does sound like a good change to it, though.

GameData/RP-1/Contracts/Earth Observation 1/EOSIRRad1.cfg
SierraHotel84: Ahh, right, the whole not everyone launches from the Cape bit.

GameData/RP-1/Contracts/Commercial Applications 1/EarlyNavNetwork3.cfg
```diff
@@ -0,0 +1,230 @@
...
+
+			vessel = NavSat III
+		}
+		PARAMETER
+		{
+			name = HasPower
+			type = HasResource
+			resource = ElectricCharge
+			minQuantity = 1.0
+			title = Craft must have ElectricCharge after one day
```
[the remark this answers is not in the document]
SierraHotel84: This is where the confusion came from, you said titles and I interpretted it as the quantity.

GameData/RP-1/Contracts/Earth Observation 1/EOSCherenkov.cfg
SierraHotel84: Yeah, no clue where that came from.

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -287,11 +288,103 @@ RP0_PROGRAM
...
+{
+	name = EarlyCommApp
+	title = Early Commercial Satellites
+	description = Although it was through government programs that the earliest satellites were launched, commercial use of them was always envisioned as well.  Through this program, the earliest iterations of commercial-use satelites will be launched.  With wildly varying needs, each satellite is dedicated to a specific requirement, typically far more focused than the earlier broad scientific satellites. 
+	requirementsPrettyText = Complete Early Satellites program
+	objectivesPrettyText = Launch the first commercial-use satellites and build a satellite communication network. 
+	nominalDurationYears = 3
+	baseFunding = 280000
+	repDeltaOnCompletePerYearEarly = 110
+	repPenaltyPerYearLate = 110
```
SierraHotel84, line 347: Should this get bumped up a little higher since we're doing considerably more than the original Comm Network program required?  We added a year to the duration, but with the number of launches it seems like it should be higher

GameData/RP-1/Contracts/Earth Observation 1/EOSCoronaRep.cfg
```diff
@@ -0,0 +1,149 @@
...
+			{
+				name = PartValidation
+				type = PartValidation
+				
+				VALIDATE
+				{
+					MODULE
+					{
+						name = Experiment
+						experimentID = RP0photos2
```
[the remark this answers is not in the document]
SierraHotel84: Ok, this one has been confounding me the last few days and I'm throwing in the towel.  How do I require a certain number of biomes to be collected?  I get the + 1 + Repeat_Count part, but I cannot figure out how to actually call the first part.  The Contract Configurator wiki only shows specific biomes by name.

NathanKell:
> > 6 distinct launches works for requireds. Less than that for required (or 6 total including optionals) feels very light tbh.
> 
> 6 required, with optional 7th as it currently stands. I'll adjust the gravity one to make it 8 and leave that one optional as it is now.
Cool! Uh, but I thought you added two optional IR contracts? I'm confused now. Are you not counting the optional follow-ons to required contracts?

GameData/RP-1/Contracts/Commercial Applications 1/EarlyComSat-CA.cfg
NathanKell: Yeah I think Max only hit some of the old commsat contracts, not all? Might have been we punted until the revamp anyway, which is...now :D

SierraHotel84:
> > > 6 distinct launches works for requireds. Less than that for required (or 6 total including optionals) feels very light tbh.
> > 
> > 
> > 6 required, with optional 7th as it currently stands. I'll adjust the gravity one to make it 8 and leave that one optional as it is now.
> 
> Cool! Uh, but I thought you added two optional IR contracts? I'm confused now. Are you not counting the optional follow-ons to required contracts?

I did, all 3 are now in the IRRad1 file and dumped the separate Repeatable one

GameData/RP-1/Programs/Programs.cfg
NathanKell: Yep, good point. Probably 420k instead of 280k? Keeping the budget at 140k/yr which seems about right.

GameData/RP-1/Contracts/Commercial Applications 1/FirstComSat-CA.cfg
```diff
@@ -0,0 +1,131 @@
...
+			minApA = 5000000
+			disableOnStateChange = true
+			title = Achieve orbit with a minimum inclination of 40 degrees, a minimum perigee of 850 km, and an apogee of more than 5,000 km
+			
+			PARAMETER
+			{
+				name = Duration
+				type = Duration
+
+				duration = 30d
```
NathanKell, line 99: This still has the 30d duration requirement. All the contracts in CA should have duration requirements of 2 minutes, like they had in Targeted Sats, except the network ones that should keep their 2 day requirements from the template network contracts.

GameData/RP-1/Contracts/Commercial Applications 1/FirstNavSat-CA.cfg
```diff
@@ -0,0 +1,125 @@
...
+		}
+		PARAMETER
+		{
+			name = Orbit
+			type = Orbit
+			minInclination = 50
+			minPeA = 350000
+			maxEccentricity = 0.26
+			disableOnStateChange = true
+			title = Achieve orbit with a minimum inclination of 50 degrees and a minimum perigee of 350 km
```
NathanKell, line 86: I think this doesn't need a title anymore? Because it's got a bunch of parameters.

GameData/RP-1/Contracts/Earth Observation 1/EOSCorona.cfg
```diff
@@ -0,0 +1,141 @@
...
+
+			PARAMETER
+			{
+				name = ReturnHome
+				type = RP1ReturnHome
+				title = Return the film to the surface
+				hideChildren = true
+				completeInSequence = true
+			}		
+		}
```
NathanKell, line 139: This doesn't have a science parameter--I assume the issue is the same as the EOSCoronaRep that you mentioned?

GameData/RP-1/Contracts/Earth Observation 1/EOSCoronaRep.cfg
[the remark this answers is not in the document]
NathanKell: This isn't a base CC parameter, it's one we add in RP-1. Here are the fields you can use: https://github.com/KSP-RO/RP-1/blob/master/Source/CC_RP0/RP1CollectScienceFactory.cs#L22-L28

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -287,11 +288,104 @@ RP0_PROGRAM
...
+		complete_contract = EOSIRRad1
+		complete_contract = EOSIRRad2
+		complete_contract = EOSIRRad3
+  		complete_contract = EOSCorona
+		complete_contract = first_OrbitRecover
+	}
+	
+	OPTIONALS
+	{
+   		EOSGrav1 = true
```
NathanKell, line 327: IR Rad 2 and 3 should be here rather than required

NathanKell (changes requested):

GameData/RP-1/Contracts/Commercial Applications 1/FirstComSat-CA.cfg
SierraHotel84: Caught that as I was scrolling through the changed files.

GameData/RP-1/Contracts/Earth Observation 1/EOSCorona.cfg
SierraHotel84: Correct, have been holding off on those 2 until I knew how to do the parameter

GameData/RP-1/Contracts/Earth Observation 1/EOSCoronaRep.cfg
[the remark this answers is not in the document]
SierraHotel84: Oh, so I just need to include "minSubjectsToComplete" and the number of biomes.  Easy enough.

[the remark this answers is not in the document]
SierraHotel84: Ok, not entirely sure if this is implemented correctly but I used the math from the updated photos2 and went slightly under.  I think.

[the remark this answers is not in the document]
Teykn: For the `minSubjectsToComplete`, think you want to set it to `@requiredBiomes`? Set to a static 3 right now.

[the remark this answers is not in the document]
SierraHotel84: Oh, right, because that only increments the text, not the actual requirement.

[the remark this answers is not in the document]
SierraHotel84: Also realized that math doesn't work out anyway.  Not sure what the best way to increment this is, either the percentage completion or the number of biomes.  If it's 8 launches required for 100% of 10 biomes, the simplest math is 12.5% per launch over all biomes?

GameData/RP-1/Contracts/Earth Observation 1/EOSCoronaRep.cfg
```diff
@@ -144,8 +138,8 @@ CONTRACT_TYPE
 				situation = InSpaceLow
 				experiment = RP0photos2
 				fractionComplete = 0.9
-				minSubjectsToComplete = 3 
-				title = Collect 90% of the available imagery across @requiredBiomes biomes
+				minSubjectsToComplete = 3 + @index
+				title = Collect 90% of the available imagery across @minSubjectsToComplete biomes
```
SierraHotel84, line 142: Realized this math doesn't work.  Suggestions for starting percentage and number of biomes?

Source/Tech Tree/Parts Browser/data/RP_1__Bluedog_Model.json
```diff
@@ -523,7 +523,7 @@
         "category": "SCIENCE",
         "info": "Cosmic Ray",
         "year": "1963",
-        "technology": "scienceHuman",
+        "technology": "basicScience",
```
NathanKell, line 526: This doesn't exist in our tech tree

NathanKell (changes requested):

Source/Tech Tree/Parts Browser/data/RP_1__Bluedog_Model.json
SierraHotel84: Whoops, other one was wrong, then.

NathanKell:
Ok! I'd say the next step is to test and make sure you can play through the two programs and every contract is completable! (I mean, obviously you don't need to do all 8 repeatable launches for the Corona repeatable for example, but at least a couple to be sure they increment properly for everything that increments, and at least once for every individual CONTRACT to be sure it loads and plays right).

Note you will temporarily need to change RP0photos1a to RP0photos1 since phots 1a is in the bigbalance branch.

ppboyle:
Took a quick look at this branch.  

Currently Corona and Molyniya contracts are not working due to CC issues - I imagine that's know?  Do you need assistance?

I'm still concerned about how easy it is to complete EOS with 3 launches - 1 launch for corona, one for IR and then a small kick to do gravimetric - so 7 days + 60 days - and 1 launch that can simultaneously do Magnetometer + Cosmic Ray + Cherenkov, without even doing a kick - in fact the requirements on those 3 are so similar players would almost be stupid not to do them simultaneously.

The gravimetric contract's inclination requirement is odd - it's requiring 48 to 52 degrees inclination, while the experiment itself requires 50 to 130.  So actually the player can only complete the contract at 50 to 52?  I think the contract should match the experiment.

The Cosmic Ray Beyond LEO contract says it requires a TV camera in text, but doesn't actually.

Having Early Comms or Early Nav contracts active at the same time as the constellation contracts may cause problems - eg: A satellite for one can be counted by the other.  You may wish to resolve this by either locking out the corresponding contracts, or making the requirements mutually exclusive.

Teykn:
> Currently Corona and Molyniya contracts are not working due to CC issues - I imagine that's know? Do you need assistance?

@SierraHotel84 
Suspect that in Corona, you need to use `@/maxCompletions` instead of `@maxCompletions`, and in Molniya, `@payload` instead of `@/payload`

SierraHotel84:
@ppboyle thanks, all good points.  The one thing I did just think about is that the Cherenkov unlocks in the next science node, so it'll take quite a bit more science to complete it, which I think basically forces it into its own launch.  Will take a look at that, though.  I also wasn't thinking about the fact that all of the contracts could be accepted at once, the New Vessel node had me thinking it would block that but then realized it's how you can combine contracts for first satellite, so I'll tweak that as well.  Good catch on the grav one, I was basing the orbital requirement off the historical satellite orbit, but hadn't looked at the experiment req's yet.  

@Teykn thanks, looking at the CC errors next

NathanKell:
> @ppboyle thanks, all good points. The one thing I did just think about is that the Cherenkov unlocks in the next science node after CR1 and Mag1, so it'll take quite a bit more science to complete it, which I think basically forces it into its own launch.

Why would it, rather than just pushing the player to wait to launch until it's available, like players often wait to combine FO and FSO? Note that the collect-science parameter doesn't require _that vessel_ to collect the science, so the player could be collecting science with the named experiments prior to that launch and merely "finish off" the science with that launch.

SierraHotel84:
Hmm, fair point.

NathanKell:
@SierraHotel84  LMK when everything's tested and working and I'll merge! :)

SierraHotel84:
Will do, hoping to get it all wrapped up tomorrow

SierraHotel84:
tech tree costs

Where are these values stored?  I thought I'd updated all of the configs for the new part but can't seem to find this one.

So far this is proving to be a challenging set of contracts due to the limited science early on.  I think for the average player with the new balance changes, this is going to be a much more engaging couple of programs.  Will upload all of the tweaks/changes at once later today.

NathanKell:
> Where are these values stored? I thought I'd updated all of the configs for the new part but can't seem to find this one.
> 
> So far this is proving to be a challenging set of contracts due to the limited science early on. I think for the average player with the new balance changes, this is going to be a much more engaging couple of programs. Will upload all of the tweaks/changes at once later today.

When you add the part to the parts browser (or once added, by searching for the part and clicking on it), you can change the cost there. Then commit your changes to json, and then generate configs.

Source/Tech Tree/Parts Browser/data/RP_1__Bluedog_Model.json
```diff
@@ -512,5 +512,28 @@
...
+        "info": "Cosmic Ray",
+        "year": "1963",
+        "technology": "scienceHuman",
+        "ro": true,
+        "orphan": false,
+        "rp0_conf": true,
+        "spacecraft": "CD-1",
+        "engine_config": "",
+        "upgrade": false,
+        "entry_cost_mods": "Cherenkov",
```
NathanKell, line 533: this should not have an ECM set. Clear the Entry Cost Modifiers field in the parts browser and re-export, and instead just set a fixed Entry Cost of a few thousand.

SierraHotel84: Easy fix. Thanks.

SierraHotel84:
@NathanKell Ok, I think I'm calling it ready to go!  Per @ppboyle's comments, I gated Cherenkov and MagScan behind CosmicRay1, which also fits historically since they were a few years later.  Also adjusted orbit requirements to make them far less likely to be combined.  Added blocks in Com and Nav networks for the repeatables.

NathanKell:
@SierraHotel84 great!! And you were able to successfully complete all contracts in both programs?

SierraHotel84:
I didn't do the com network since it is already known to work, or nav network since it's basically a direct port of the com ones. Corona and Corona Rep I had to cheat a little since I didn't have the new photo experiments. I didnt try the new orbital requirements on the ones mentioned above but previously tested them.

NathanKell:
I'm confused with Corona--that's why I mentioned using RP0photos1 since you would have that experiment. Or did you mean you did do them with the photos1 experiment and they worked?

For CA, while it's true the contracts are broadly a port, it's still worth testing--if nothing else that will verify the contract flow is correct, and that nothing re: rewards or other things got busted in the porting over.

Yeah the orbital-requirement change stuff makes sense.

Great to hear with the others!

SierraHotel84:
Fair point on the nav network, will run through it tomorrow to verify. CoronaRep would have completed with the updated shortened experiment length, just didn't have enough battery to run full original length. I'll grab photos1a out of BB and do Corona again to be sure.

NathanKell:
Eh doing it with photo1 is fine! It's not like they're really different. Feel free to use infinite EC or something :)

SierraHotel84:
Had to ditch the CollectScience from CoronaRep, tried using the original Photo2, the one from BB, and a custom one, and it would not complete the collect step no matter how I rearranged the config.  Left it as a part validation for the camera and a time requirement that matches the new photos2 experiment.  All other contracts checked out and validated as intended.

NathanKell:
Thanks for doing that final pass! Looks like you caught a bunch of stuff we both missed. :)

And THANK YOU for making these programs! Merging with pleasure.

PR #2239, opened 2023-09-19
X-Planes tweaks

Blothorn:
Fix required/optional tags for new contracts, and reduce hypersonic requirements somewhat--it seems significantly harder than I remember it being to keep an X-15 cockpit intact at 3km/s, possibly a result of it getting the non-reentry-rated tag.


NathanKell:
The tag should only affect whether the part is able to be occluded, but perhaps weird things are happening in the dragonland that is thermo.

Blothorn:
Hm. Possible I also just messed up the earlier tests, or that particular plane had some weirdness--in any event, I just tested at 2.6km/s and am pretty confident it's reasonable.

PR #2240, opened 2023-09-20
Minor fixes on X-Plane Program

QiZhengyuan:
...such as maxCompletions in Mach 2+ supersonic contracts, and contract type of the Karman line contract.

Somone might had forgotten that the total amount of speed records in Mach 2+ supersonic contracts had reduced to 6, while it might be helpful to mark the most challenging contract(s) of a program (Karman line in this case) as "Capstone", like the other programs did.


Blothorn:
Thanks!

PR #2241, opened 2023-09-20
Fix contract confidence display with leader reputation affects

Blothorn:
Noticed when taking Glushko as admin--standard-difficulty SR at 100% rewards showed as 48 confidence in admin, but provided the correct 50 when actually accepted and completed. It seems that the display was already using base rep. This was tested using Glushko--altitude SR gave increased reputation (11 after rounding) but 50 confidence in the new total, total confidence earned summary, MC contract summary, contract completion message,  and career log.

### DIFF
--- a/Source/Harmony/Contract.cs
+++ b/Source/Harmony/Contract.cs
@@ -58,8 +58,7 @@ namespace RP0.Harmony
 
             if (Programs.ProgramHandler.Instance != null && Programs.ProgramHandler.Instance.RepToConfidenceForContract(_contract, _isReward) is float repToConf && repToConf > 0f)
             {
-                double correctedRep = _contract.ReputationCompletion / CurrencyUtils.Rep(TransactionReasonsRP0.ContractReward, 1d);
-                var cmq = CurrencyModifierQueryRP0.RunQuery(TransactionReasonsRP0.ContractReward, 0d, 0d, 0d, correctedRep * repToConf, 0d);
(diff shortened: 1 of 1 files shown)


NathanKell:
As you say it's correct on contract summary (which I assume you mean the completed contract tab?) so changing this back will then break completed contracts, no? The issue is that we don't actually record confidence gained from a contract (because we can't add data to KSP, just change code, with Harmony) so it's a bit of a pickle.

NathanKell:
lolnope I'm wrong, KSP is super dumb about this and doesn't store the actual awards granted on contract success, it recalculates based on current CMQs.

Blothorn:
I meant that I've tested that with the change confidence is displayed consistently and correctly (i.e. 50 despite the reputation bonus from Glushko) in all those places--I haven't checked whether the problem affects the completed contract list, but they are displaying correctly with it.

Worth note is that if I understand how the code works correctly, the change to use base reputation didn't actually change the confidence added (https://github.com/KSP-RO/RP-1/blob/master/Source/Programs/ProgramHandler.cs#L346), just various places it was displayed--I think the contract's `ReputationCompletion` has been base reputation all along.

NathanKell:
Yup you're 100% correct, per above KSP also Does The Dumb and just recalculates the base rep based on the current CMQ regardless of what the actual advances/awards were when the contract was accepted. So we'll just keep following in those footsteps.

PR #2243, opened 2023-09-21
Ensure that reconstruct does not downsize the hangar

Blothorn:
Unlike in LCs, the hangar does not pay maintenance for increases to maximum dimensions, and since they dictate maximum engineer count it's important not to downsize it unnecessarily. This is particularly so because many planes can use several times more engineers than a hangar built to their dimensions supports (https://discord.com/channels/319857228905447436/845457276641738782/1154214299078774834)--presently pressing "reconstruct" can cripple your ability to build the plane selected during reconstruction.

I've tested that this change provides what I believe to be the correct behavior--hangar reconstruction and upgrades upsize but do not downsize, while LC reconstruction downsizes but upgrades do not.

### DIFF
--- a/Source/KerbalConstructionTime/GUI/GUI_NewLC.cs
+++ b/Source/KerbalConstructionTime/GUI/GUI_NewLC.cs
@@ -64,9 +64,11 @@ namespace KerbalConstructionTime
 
         private static void SetFieldsFromVessel(BuildListVessel blv, LCItem lc = null)
         {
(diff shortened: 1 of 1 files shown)


NathanKell:
I've actually been thinking about removing "Reconstruct' as a preset option generally, because it leads to this confusion for LCs as well as hangars (it's just way more damaging with hangars).

NathanKell:
Did the change. Thanks for poking me on this, and thanks for the PR! (And getting a working build setup :D )

PR #2249, opened 2023-09-24
RP-1 Contract Satellites

ryanc55:
Remove Nav/Com/Weather payloads from SM tanks. Introduce custom RP-1 payload satellites for NavSatPayload and ComSatPayload.
Supported by https://github.com/KSP-RO/ROLibrary/pull/24


GameData/RP-1/Contracts/Earth Satellites/FirstWeatherSat.cfg
Clayell, line 1: Sorry, just popping in because I'm curious about this. Why would you replace the payload for the weather sat contract with navsat payload? Have I misinterpreted this?

NathanKell: Because there are two lines of models, so there have to be 2 resources. :)

Clayell: Uh... I'm not really sure what that means. Do you mind explaining that to an idiot?

NathanKell: 1. This PR adds a part that contains ComSatPayload and a set of models for that part
2. This PR adds a part that contains NavSatPayload and a set of models for that part
3. We had 3 sat payload resources.
4. There are now only two supported ways put resources on your craft: one that puts ComSatPayload on (1) and one that puts NavSatPayload on (2)
5. So we need to have only 2 payload resources in contracts.

Clayell: Ah ok that makes sense, thank you. How come we don't have a part for weather sat though? Is it really that similar to navsat payload?

ryanc55: It's just a practical matter.   There was only one non-repeatable weather sat contract and making new models is time consuming.   We specifically wanted to have dedicated parts and models for each payload type, with several tech looks to choose from.   Weather sats may reappear in the future.

PR #2250, opened 2023-09-24
Contract Sats: Bugfixes and Balance Changes

ryanc55:
Add missing texture
Decrease temperature tolerance
Fix tank weights
Decided to remove zombie resources from all locations.   This will cause ongoing nav/com/weather missions to fail, and affected KCT craft will have to be rebuilt or edited.  But I didn't like having those half-broken configs out there causing potential issues.


NathanKell:
I'd suggest not doing the last bit, since it will actively break craft in flight and therefore be somewhat savebreaking. The existing upgrade script _already_ nukes it in KCT vessels, because KCT vessels are upgraded as craft files whenever their craft nodes are loaded (which they are on edit or launch). I just added https://github.com/KSP-RO/RP-1/commit/ef32612d29c61e1eb849f31359f3fa5e19d51ac0 which will let you distinguish the context, but as it stands (before this PR) they're being nuked from everything but already-launched / in-flight vessels, which I think is the proper compromise.

NathanKell:
Regarding temperature, you probably want to give the parts material tags if you haven't yet? https://github.com/KSP-RO/RealismOverhaul/pull/2876

NathanKell:
Wait I'm super confused. I looked at the diff and the upgradescript's context is still _only_ craft files, so it won't actually hit inflight vessels? What are the actual code changes going on??

NathanKell:
(handled via discord)

PR #2254, opened 2023-10-04
EOS/CA Corrections

SierraHotel84:
-Missed the confidence cost to the Commercial Apps program
-Corrected camera requirement for Corona 1


SierraHotel84:
Suggest incrementing the photography experiment ID numbers to match the cameras.  Corona completed correctly with photos1 in testing but it has the same problem photos2 did, so I just fixed the completion requirements to be identical and set the duration long enough to collect a full sample.

NathanKell:
Increasing the ID numbers breaks saves. That's why the new camera was called 1a. However we can call the display stuff whatever we want.

NathanKell:
Gonna go ahead and merge this for now 👍

PR #2256, opened 2023-10-06
Grammar tweaks - MagScan and Visible Imaging

SierraHotel84:
- Corrected MagScan duration title, stated 14 days, requires 45
- Changed Visible Imaging experiments titles to Video Imaging


NathanKell:
Turns out this was save-breaking because we forgot that Kerbalism's Configure module uses the experiment _title_, not its ID, when storing the active configuration. See https://github.com/KSP-RO/RP-1/issues/2262

SierraHotel84:
Ahh...well, that's less than ideal.

PR #2260, opened 2023-10-07
Lunar Orbiter & Mapper PE Max Updates

westlakem:
First Lunar Orbiter & Mapper (Uncrewed) & the optional mission afterwards, requires you to collect visual imaging 2 (VI2) science from around the moon.  The current contract doesn't require any PE Max, however the biome science has to be collected from space low.  To see why I wasn't collecting the science, I had to reference 3 different materials:

- the contract requirements (to see it was biome specific), 
- the Science window (too see that biome science only comes from low orbit), 
- and the wiki to see where "space low" started

In talks with the dev-feedback channel in discord, someone pointed out that someone else on the discord had to "cheat" their way to completing the contract because they weren't collecting the science, and assumed it's because they were in "space high" since you can complete the orbit requirements without ever dipping into space low.  This PR is an attempt to bring the orbit requirements of the contract in line with the altitude requirements for the science.

Alternative: We could also lower the AP requirements to 150KM so they were completely in space low, but that may be more constrictive than the contract intends to be.

DISCLAIMER: I have not tested these changes and only assume this will edit the contract properly.  Please advise if I need to change this (or feel free to do so yourself)

### DIFF
--- a/GameData/RP-1/Contracts/Lunar Surface Probes/MoonOrbiter.cfg
+++ b/GameData/RP-1/Contracts/Lunar Surface Probes/MoonOrbiter.cfg
@@ -45,7 +45,7 @@ CONTRACT_TYPE
 	{
 		type = float
 		minPe = 20000
-		maxPe = 225000
+		maxPe = 250000
 		maxAp = 500000
 	}
 
--- a/GameData/RP-1/Contracts/Lunar Surface Probes/MoonOrbiter.cfg
+++ b/GameData/RP-1/Contracts/Lunar Surface Probes/MoonOrbiter.cfg
@@ -45,6 +45,7 @@ CONTRACT_TYPE
 	{
 		type = float
 		minPe = 20000
+		maxPe = 150000
 		maxAp = 500000
 	}
 
@@ -116,6 +117,7 @@ CONTRACT_TYPE
 			targetBody = Moon
 			situation = ORBITING
 			minPeA = @/minPe
+			maxPeA = @/maxPe
 			maxApA = @/maxAp
 			disableOnStateChange = true
 			
--- a/GameData/RP-1/Contracts/Lunar Surface Probes/MoonOrbiter.cfg
+++ b/GameData/RP-1/Contracts/Lunar Surface Probes/MoonOrbiter.cfg
@@ -45,7 +45,7 @@ CONTRACT_TYPE
 	{
 		type = float
 		minPe = 20000
-		maxPe = 150000
+		maxPe = 225000
 		maxAp = 500000
 	}
 
--- a/GameData/RP-1/Contracts/Lunar Surface Probes/MoonOrbiterOptional.cfg
+++ b/GameData/RP-1/Contracts/Lunar Surface Probes/MoonOrbiterOptional.cfg
@@ -53,7 +53,7 @@ CONTRACT_TYPE
 	{
 		type = float
 		minPe = 20000
-		maxPe = 225000
+		maxPe = 250000
 		maxAp = 500000
 	}
 
--- a/GameData/RP-1/Contracts/Lunar Surface Probes/MoonOrbiterOptional.cfg
+++ b/GameData/RP-1/Contracts/Lunar Surface Probes/MoonOrbiterOptional.cfg
@@ -53,6 +53,7 @@ CONTRACT_TYPE
 	{
 		type = float
 		minPe = 20000
+		maxPe = 150000
 		maxAp = 500000
 	}
 
@@ -125,6 +126,7 @@ CONTRACT_TYPE
 			targetBody = Moon
 			situation = ORBITING
 			minPeA = @/minPe
+			maxPeA = @/maxPe
 			maxApA = @/maxAp
 			disableOnStateChange = true
 			
--- a/GameData/RP-1/Contracts/Lunar Surface Probes/MoonOrbiterOptional.cfg
+++ b/GameData/RP-1/Contracts/Lunar Surface Probes/MoonOrbiterOptional.cfg
@@ -53,7 +53,7 @@ CONTRACT_TYPE
 	{
 		type = float
 		minPe = 20000
-		maxPe = 150000
+		maxPe = 225000
 		maxAp = 500000
 	}
 


hallucinogender:
Imposing a maximum periapsis limit is probably preferable from a gameplay perspective because it will still allow eccentric orbits, but with the condition that they approach closely enough to the Moon at least _some_ of the time.

lpgagnon:
Given the goal of helping the player go to an orbit where the needed experiment will actually function, a max Pe of 150 may not do the best job. Yes, it will force the player to spend _some_ time in the right situation; but a 499x149 orbit that spends 10s per orbit in the right situation is going to take forever to gather the required amount of science, and by the time the player realizes it, there's a good chance it's too late to save the mission.

So this is no substitute for actually spelling out what the experiment needs; it also highlights the broader problem that there's nothing at all in-game that explains where the low/high boundaries are.

westlakem:
> Given the goal of helping the player go to an orbit where the needed experiment will actually function, a max Pe of 150 may not do the best job. Yes, it will force the player to spend _some_ time in the right situation; but a 499x149 orbit that spends 10s per orbit in the right situation is going to take forever to gather the required amount of science, and by the time the player realizes it, there's a good chance it's too late to save the mission.
> 
> So this is no substitute for actually spelling out what the experiment needs; it also highlights the broader problem that there's nothing at all in-game that explains where the low/high boundaries are.

Yea, so the options are set the peMax to 150 so they spend _some_ of their time there, or change the apMax to 150 to they spend _all_ of their time in space low.

lpgagnon:
the other other option is to actually _tell_ the player what he's being asked to do instead of just trying to force him into the right orbit without telling him why

westlakem:
> the other other option is to actually _tell_ the player what he's being asked to do instead of just trying to force him into the right orbit without telling him why

I'll leave it to the mods if they want to update the description to tell them why they need to be under 150, as it would be the only contract with that type of description.

hallucinogender:
If no other contracts which require specific experiments include written details of the experiment parameters, that raises a bigger question: is the problem _actually_ the fact that this particular contract doesn't have good enough orbital parameters, or is the problem that players don't know how to figure out what conditions an experiment needs in order to run? If the latter, then that needs to be addressed.

As for this contract, there should ideally be some possible leeway with orbit eccentricity. A satellite which reaches orbit of the Moon with an adequate periapsis but fails to circularize would, in reality, still be able to collect some useful data and fulfil its purpose, so ideally the same should be true for this contract in the game.

Perhaps a sufficient solution would be to introduce a condition for maximum periapsis that is _lower_ than the upper bound for the space low environment (perhaps 10 to 20 kilometers lower than the upper bound). This would guarantee that a satellite which fulfils the contract's orbital parameters is also capable of collecting the data, _and_ would mean that it spends more than the absolute minimum time in the low space environment.

westlakem:
I agree from a "completing the mission" aspect, 130 or something like that
makes more sense, but then I would argue that that number is arbitrary and
isn't exactly required either (you could technically circularize at 140 and
get the science but never complete the contract.

SierraHotel84:
> the other other option is to actually _tell_ the player what he's being asked to do instead of just trying to force him into the right orbit without telling him why

Sure, but that should be in addition to setting the orbit requirements correctly.  What's the point of having mission parameters allowed that don't actually accomplish the purpose of the mission?  If the goal of the mission is to map the lunar surface, then you shouldn't be able to complete the mission without actually gaining the science that was the intent of the mission anyway. If someone wants to go send an orbiter to the moon with the stated purpose of mapping the moon, and then not map the moon, they shouldn't get credit for completing the mission.

Tsits11:
This is ridiculous. Visible imaging experiment specifies that I only gathers biome specific data in space LOW condition . A player can easily figure that out, no need for any changes as things stand really….

NathanKell:
I think it's worth considering whether we need to keep the spacelow threshold at 150km anyway. I'd certainly not like to enforce a Pe or Ap so low for this contract. It is equally true that other experiments also have requirements that can bite you--for example the experiments with minimum eccentricity, which also do not work in frozen orbits.

hallucinogender:
> I'd certainly not like to enforce a Pe or Ap so low for this contract.

That probably necessitates raising the space low threshold then. Are there any contracts designed around experiments which require space high around the Moon? If so, the lower bounds for orbits required by these contracts should be considered the upper bound for how much the space low threshold _can_ be raised without interfering with them.

westlakem:
> It is equally true that other experiments also have requirements that can bite you--for example the experiments with minimum eccentricity, which also do not work in frozen orbits.

Yes, however they do not have specific missions with collecting that science in them, that also has conflicting orbit requirements.  EX: there's no mission that requires you to do a meteorite scan, as well as be in GSO.  You're just unable to collect the science while in GSO.

> I think it's worth considering whether we need to keep the spacelow threshold at 150km anyway. I'd certainly not like to enforce a Pe or Ap so low for this contract. 

Raising the spacelow threshold to 500km would fix it for this contract (since it already has the apMax requirement at 500km), but I don't know what side effects that would have.  I'm kind-of torn on this, as it would make getting space low a lot easier, but If I wasn't forced to be so close to the moon on my first principia playthrough, I would have never gone "Why are all my space low sats getting destroyed" and learn about frozen orbits.

hallucinogender:
> Raising the spacelow threshold to 500km would fix it for this contract (since it already has the apMax requirement at 500km), but I don't know what side effects that would have. I'm kind-of torn on this, as it would make getting space low a lot easier, but If I wasn't forced to be so close to the moon on my first principia playthrough, I would have never gone "Why are all my space low sats getting destroyed" and learn about frozen orbits.

It's probably best not to raise the space low threshold as high as 500km in that case. Perhaps this combined solution would work: raise the spacelow threshold to either 200km, 250km, or 300km, and impose a maximum periapsis limit 50km below that. As long as this doesn't conflict with any _other_ contracts, I don't see an obvious reason why it wouldn't resolve the issue.

westlakem:
So who takes it from here?  I'm not a maintainer and need some direction here from someone who is.

siimav:
I feel like both raising the space low threshold and clarifying the requirement in contract description is the way forward.
Btw, I wonder why was 150km chosen in the first place? Sounds awfully low, especially when using Principia. I would suggest 250km instead.

Capkirk123:
The moon space low threshold has been increased to 250 km [here](https://github.com/KSP-RO/RealSolarSystem/pull/301). I'm not sure if we still want to specify a max perilune to force an orbit that passes through moon space low, but the contract should be more reasonable now.

Capkirk123:
Bumped up the max Pe value to 250 km, to match the new moon space low value/

PR #2275, opened 2023-10-17
GEOCommNetwork Program rework

Ballatik:
Add Commercial Applications 2 program, deprecate GEOCommNetwork, move applicable contracts over
Add Optional FirstGeoSync contract.  Requires positive power, player keeps craft
Update Required FirstGEOSat contract.  Requires positive power, player keeps craft
Add Required FirstTargetedGeo contract.  Requires comsat payload, craft transfers, bonus for precise targeting 
Update Optional GEORepeatComSats contract to require target.  Bonus for precise targeting
Replace GEOComSatNetwork with 3 and 4 sat versions.
Add optional targeted Molniya contract
Add MolniyaNetwork3 contract
Add optional FirstTundraSat contract. Requires positive power, player keeps craft
Add required FirstTargetedTundra contract. Requires comsat payload, craft transfers. 
Add optional repeatable targeted molniya and tundra contracts
All repeatable contracts scale from RP1ComPayload setting, and payloads get larger after each completion.

Notes:

- The first sync, stat, and tundra contracts require power instead of payload to encourage the player to make useful satellites for those, and they get to keep them.  They are optional, so if you don't want one there you can skip it.  
- The rest require payload and are transferred on completion.  Payloads are increased from similar ComApp1 contracts.
- ~~Lowered funding to bring it in line with ComApp, but kept duration at 2 years since it can be done with 5 launches, 3 if you've got a big enough LV.  I was able to do it in 5 with a small upgrade to the LC that I already had for early lunar, and the total build time for those launches was a little under 700 days without maxed engineers.~~ Up to 5/7 launches, more of which are higher dV.  Adjusted to 3 years, 800k funding.
- Windows for the targeted geosats are very loose.  4 km works out to about 36 degrees, with a bonus at 2km or 18 degrees.  In testing this was tight enough to make me think about how to get to the right spot at the right time, but was pretty easy to execute once I thought about it.  
- Added the Molniya network option for those in the higher latitudes.  I can't think of a way to check spacing like we can for geostationary networks, ~~so I removed the 2 minute stability check.  That way if the player launches them on one vessel, they will at least need to separate them before getting into the right orbit or they will all get assigned as Sat1.~~ Added a maxpayload during an intermediate orbit parameter instead.  It ensures that the player decouples when the period is less than 6 hours and then sends each satellite to the intended orbit.
- I did all of my testing with Principia installed, though that generally that makes contracts more finicky, not less.


NathanKell:
This is neat! I think this would work even better as just Commercial Applications II. You're already part of the way there by including Molniya contracts, but I think a program centered around first GEO and Tundra launches, repeatable targeted Molniya/GEO/Tundra launches (where Molniyas have way more comsat payload than in CommAppI), and some networks.

The idea being that you should be able to complete CAII missions at the low end with something on the order of 70t on-the-pad (i.e. Thorad or Delta L class), maybe ~130t for the heavier end of the GEO payload range.

SierraHotel84:
Since it came up in the CA1 program and I see it mentioned in the OP, is there any way to check orbital spacing without requiring something like Skopos (assuming it does, I've only read through the description of it)?  We can check the base orbital parameters and require specifics like orbital period, but is there a way to compare timing/distance between sats, so that a 3 or 4 sat commnet, for example, actually needs to be spaced equally in orbit?

Ballatik:
> Since it came up in the CA1 program and I see it mentioned in the OP, is there any way to check orbital spacing without requiring something like Skopos (assuming it does, I've only read through the description of it)? We can check the base orbital parameters and require specifics like orbital period, but is there a way to compare timing/distance between sats, so that a 3 or 4 sat commnet, for example, actually needs to be spaced equally in orbit?

For Geostationary, I have it requiring waypoints since if you are over a waypoint and geostationary, you will stay over the waypoint.  We can require spacing by spacing the waypoints.  

For other orbits, I have not found a way (but would love ideas!)  As a partial measure, an intermediate orbit with a max payload at least requires that single launch completions separate the 3 (or 4) parts before reaching the target orbit.  In practice, my 3 sat Molniya done this way ended up spaced about 5 minutes apart.  Increasing the orbit check duration to about 30 minutes would force you to finalize on 3 separate revolutions which would end up being spaced more, but doesn't really change what the player would be doing and doesn't force "good" spacing.

SierraHotel84:
Good call on the waypoints for geo, you're right that would work well for a geo network.  I saw that in one of the original contracts and figured it would be good for CA2.  Just don't know if there's anything similar that can be done for lower orbits.  It's too easy to just dump off 3 or 4 sats on a single launch and have them all count.

NathanKell:
Sorry I missed this was ready for review. I'll take a look tomorrow or otherwise ASAP. 👍

Ballatik:
Changed tundra and molniya contracts to use waypoints instead of orbital parameters.  It allows a bit more flexibility in specifics while still requiring you to do what those orbits are meant for.
Clarified some notes and adjusted some difficulty.
I've run through it successfully on 3 careers now (test, russian, american) with no issues.

siimav:
See #2342. Can it be fixed in the context of this PR?

Ballatik:
> See #2342. Can it be fixed in the context of this PR?

It now treats any of the prerequisites for unlocking the repeatable contract as repeat number 0, and resets the timer when completing any of them.

GameData/RP-1/Contracts/Commercial Applications 2/MolniyaNetwork3.cfg
```diff
@@ -0,0 +1,367 @@
...
+			title = Uncrewed
+			hideChildren = true
+		}
+		PARAMETER
+		{
+			name = HasComSatPayload
+			type = HasResource
+			resource = ComSatPayload
+			minQuantity = 124.9
+			maxQuantity = 249.0
```
siimav, line 123: Is the intention of max to prevent double-dipping with another contract? Since you overrode the title then the existence of this upper limit isn't communicated to the player in any way.

siimav:
Looks like the capstone of this program is either a GEO commnet or a Molniya commnet. Both of them require the same 125 units of payload which makes one way easier to do than the other. Maybe it would make sense to give Molniya a higher payload requirement to make this a bit more interesting of a choice.

GameData/RP-1/Contracts/Commercial Applications 2/FirstTargetedMolniya.cfg
```diff
@@ -0,0 +1,169 @@
...
+		name = WaypointGenerator
+		type = WaypointGenerator
+
+		WAYPOINT
+		{
+			name = Molniya Orbit Target
+			icon = thermometer
+			altitude = 0
+			latitude = @/targetLat
+        	longitude = @/targetLon
```
siimav, line 59: Mixed tabs and spaces here and in a few other spots. Should be easy enough to spot in the web GUI.

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -482,6 +487,59 @@ RP0_PROGRAM
...
+	name = CommercialApplications2
+	title = Advanced Commercial Applications
+	description = Geosynchronous, Geostationary, Molniya, and Tundra orbits have the potential to improve satellite coverage and simplify ground antenna tracking. This program tasks you with testing these orbits, launching satellites into them for commercial customers, and creating a 3 or 4 satellite Geostationary or Molniya Network for communication.
+	requirementsPrettyText = Complete the Early Commercial Applications Program
+	objectivesPrettyText = Complete a Geostationary or Molniya Communications Network.
+	nominalDurationYears = 3
+	baseFunding = 800000
+	fundingCurve = MildBackloadedFundingCurve
+	repDeltaOnCompletePerYearEarly = 420
+	repPenaltyPerYearLate = 420
```
siimav, line 501: Not sure why these values are set so high. I feel like 270 would be more appropriate.

GameData/RP-1/Contracts/Commercial Applications 2/MolniyaNetwork3.cfg
Ballatik: Ideally a Molniya network would be spaced out for constant coverage, but there's not a way to check for that.  Instead, I opted to at least make sure the player doesn't launch all 3 on one vessel and then decouple them into identical orbits.  This allows you to launch them all at once if you want to, but requires you to decouple before finalizing your orbit which will give at least some spacing.  
I'm not sure how strongly people feel about preventing the first case, but if we want to keep this check I should definitely clarify the contract.

GameData/RP-1/Contracts/Commercial Applications 2/FirstGeosync.cfg
```diff
@@ -0,0 +1,116 @@
...
+	targetBody = HomeWorld()
+
+
+	// ************ REWARDS ************
+	prestige = Trivial       // 1.0x
+	advanceFunds = 0
+	rewardScience = 0
+	rewardFunds = 0
+	failureFunds = 0
+	rewardReputation = 40
```
siimav, line 34: Not sure it's fair to give this contract a lower reward than the optional Molniya and Tundra contracts.

siimav:
Correct me if I'm wrong but 'Tundra Orbit Test Satellite' and 'First Targeted Molniya Orbit Satellite' look double-dippable...? Not sure if there are others.

siimav:
image
You should use hide children attribute in these cases.

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -482,6 +487,59 @@ RP0_PROGRAM
...
+		FirstTundraSat = true
+		GEORepeatComSats = true
+		MolniyaRepeatComSats = true
+		TundraRepeatComSats = true
+	}
+
+	CONFIDENCECOSTS
+	{
+		Normal = 450
+		Fast = 900
```
siimav, line 539: According to The Speadsheet, the correct values are 550 and 1100.

Ballatik:
> Correct me if I'm wrong but 'Tundra Orbit Test Satellite' and 'First Targeted Molniya Orbit Satellite' look double-dippable...? Not sure if there are others.

You could do them both with a single vessel, but you would need to go to a tundra orbit first and then lower to a molniya (with an additional 350ish m/s) since the molniya contract transfers the vessel on completion.  You could also launch them both on the same vessel, separate the tundra before the molniya completes, then raise its orbit.  

Tundra Orbit Test Satellite and Geosynchronous Test Satellite could also be doubled up, but would require about 400 m/s and 2 restarts to make the change.  Geostationary Test Satellite is similar but would also involve an inclination change.  

I feel like all of these options involve enough extra work that it's reasonable to leave the option there, but I can also see making them mutually exclusive just for clarity's sake.  All of the other contracts either transfer on completion or are exclusive based on direct exclusions or prerequisites.

siimav:
350 to 400 dv really isn't a lot and is easily achievable with even a HTP thruster. I would prefer the optionals to be mutually exclusive.

siimav (approved):
LGTM

Issue #2288, opened 2023-10-27
Interplanetary contract issues


Test-Account-KSP:
- Venus Landing contract is bugged. It can be taken twice, but there is no confidence reward for the second one.  It's marked "CAPSTONE" and gives reputation again.
- Mars Landing has the same issue. The second contract is marked "Required" and also gives only reputation. 
- Crewed Targeted Moon Landing suffers from the same issue.
- Mercury, Pluto and Charon landings as well. 
- Titan Rover is an optional contract, but gives no confidence.

NathanKell:
Closed via #2290

PR #2290, opened 2023-10-30
Interplanetary contract issues

Ballatik:
Based on issue #2288 

- Split Venus landing into Capstone and optional repeatable, added optional to program definition
- Split Mars Landing and Mars Rover into required and optional repeatable, added optionals to program definition
- Split Mercury, Pluto, and Charon Landing into required and optional repeatables, added optionals to program definitions
- Crewed Targeted Moon Landing looks to be intentional.  The program requires 2 completions, and the contract is completable twice.
- Titan Rover was checking for SaturnObservation as the active program, but wasn't listed in the program.  It doesn't make sense there, so changing the check and optional listing to be SaturnMoonLandings.

### DIFF
--- a/GameData/RP-1/Contracts/Mars/Mars Landing.cfg
+++ b/GameData/RP-1/Contracts/Mars/Mars Landing.cfg
@@ -5,7 +5,7 @@ CONTRACT_TYPE
 	group = MartianSurfaceExploration
 	agent = Surveys
 
-	description = <b>Program: Mars Surface Exploration<br>Type: <color=green>Required</color></b><br><br>Design and successfully launch an uncrewed probe that will soft land on Mars and transmit a science report from the surface. This contract can be completed 2 times.
+	description = <b>Program: Mars Surface Exploration<br>Type: <color=green>Required</color></b><br><br>Design and successfully launch an uncrewed probe that will soft land on Mars and transmit a science report from the surface.
 
 	synopsis = Send an uncrewed probe to land on Mars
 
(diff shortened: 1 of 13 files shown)


NathanKell:
Thanks!

* Looks like there's an issue with the change to the Programs.cfg file? There's an awful lot of changes that seem unrelated to what you mention changing. Did you maybe use the web editor and paste some changes in, or something?
* I'm a bit concerned with unlimited repeats on the Venus Landing optional contract, because Venus landings are pretty easy and Venus windows are common. But I think we can try unlimited for now.

siimav:
> I'm a bit concerned with unlimited repeats on the Venus Landing optional contract, because Venus landings are pretty easy and Venus windows are common. But I think we can try unlimited for now.

Yeah, since the player is most likely already taking the Science Orbit repeatable then adding a landing on top of that is almost free. Maybe have just a small reward on this?

Ballatik:
> Did you maybe use the web editor and paste some changes in, or something?

Possibly yes, or possibly just my editor setup with line breaks.  I'll figure that out and fix it.

> Yeah, since the player is most likely already taking the Science Orbit repeatable then adding a landing on top of that is almost free. Maybe have just a small reward on this?

I was going on the idea that with an 8 year program you're likely only getting 4 windows meaning the required plus 3 repeats if all goes well.  I hadn't thought about piggybacking though, and really once you have a craft that works repeating the mission is extremely easy.  

Long term, I think these repeats should have increasing requirements, but that's a bigger change.  For now I think reducing the rewards for one or both is a good idea, or even giving diminishing rewards for each repeat.

Ballatik:
> There's an awful lot of changes that seem unrelated to what you mention changing.

Atom was stripping extraneous whitespace.  I found and killed that setting so the diff is much more readable.

Also lowered rewards on Venus repeatables.

PR #2295, opened 2023-11-11
Minor leader tweaks

NovemberOrWhatever:
Correct which award was won by Glushko
Make it clearer what the issue with Von Braun was
Flip an effect that should reduce SPH speed based on the name to reduce it instead of increasing it


GameData/RP-1/Strategies/Leaders/LeadersFlightDirectors.cfg
```diff
@@ -66,7 +66,7 @@ STRATEGY
 		effectTitle = No Planes at Baikonur
 		currency = Rate
 		effectDescription = for SPH integration
-		multiplier = 1.05
+		multiplier = 0.95
 		flipPositive = true
```
NathanKell, line 70: this should not be flipPositive then.

PR #2303, opened 2023-11-28
Fix Suborbital and Orbital Test Flight contracts

Test-Account-New:
Fixes https://github.com/KSP-RO/RP-1/issues/2294#issue-1984397638

screenshot5


Test-Account-New:
Flew the orbital test flight today and the fix didn't work.

PR #2317, opened 2023-12-24
Venus surface rework

Ballatik:
The goal is to leverage what makes Venus different (thick atmosphere and heat) to differentiate the contracts from Mars/Lunar landings, and to somewhat mirror real life missions (mostly the Venera series).  New version requires at least 3-4 craft to complete, but can technically be done with one launch.  Player must at least land once, and pick 3 of the following: land 2 more times, land and survive for a day, loiter at mid altitude, or 90 day high inclination orbit.  

Required Contracts (these can both be the same vessel):
probeVenus: (requires newVessel) 10 minutes of science from both high and low altitude, and survive down to 10km
venusLanding: (requires newVessel) Added specific science requirements to necessitate 1 hour survival on the surface

Required (3 of 5) Advanced Contracts:
venusLandingMulti: requires completion of surface science from 3 biomes.  
probeVenusClouds: requires a vastly slowed descent in the upper cloud layer.  Vessel is transferred on completion to avoid babysitting a craft in atmosphere.
landingVenusAdv: requires a day's worth of more advanced surface science.  This is long enough to require at least some power and heat considerations.  
orbitVenusAdv: requires 90 days and a highly inclined orbit.  
probeVenusAdv: requires 3 biomes of temp/baro high and low, or 20 minutes of MS03 flying low.

Optional Contracts:
Planned repeat landings


siimav:
The cloud layer probe is a neat idea.
Not sure what to think of the 1d duration lander though. I feel like that's a somewhat unrealistic requirement. Also in my experience 1 hour vs 1 day doesn't really have any meaningful gameplay impact. Most parts with insufficient heat tolerance already overheat on the way down or within minutes of landing.
For the multiprobe - I wonder if it would be possible to require them to be split off from the same vessel as IRL. On the other hand this might be somewhat problematic due to how long the descent takes and how KSP handles unloaded vessels.

Ballatik:
> Not sure what to think of the 1d duration lander though. I feel like that's a somewhat unrealistic requirement. 

I was looking for a way to require the player to iterate their lander design, but I agree that it strays pretty far from realism.  In practice it meant different antenna selection or a relay, and more batteries which in turn required more heatshield.  I'll see if I can find a better way to do this.

> For the multiprobe - I wonder if it would be possible to require them to be split off from the same vessel as IRL. On the other hand this might be somewhat problematic due to how long the descent takes and how KSP handles unloaded vessels.

It should be possible as long as you capture first, but you wouldn't be able to have multiple probes reentering at once.  Have orbit define the base vessel.  Rendezvous parameter (which would trigger when you stage off a probe), landing, and transmit defines vessel 1.  Repeat for vessels 2 and 3.

Clayell:
If this is ever taken out of draft, the prestige levels for this should be updated according to https://github.com/KSP-RO/RP-1/pull/2776. Notably, the script there will not work, as this PR uses text like `<color=green>Required (3 of 5 Advanced)</color>`.

PR #2319, opened 2024-01-05
Improve Soviet contractor descriptions

Tagir-1:
- OKB titles now feature not chief designers names, but modern proper names of OKBs.
- Descriptions were updated with more correct historical facts
- Logos that had generic OKB and number, now feature real, modern logos of organizations.


Capkirk123:
That was an intentional choice, since the path from design bureau to the modern state enterprises wasn't always linear, and many of them had extensive changes in structure between those times.

The current system is intended to represent the contractors as they existed during the space race. Contractor options changing over time to reflect real-world changes may be implemented in the future.

Tagir-1:
Oh, okay, got it. I would still like to fix some descriptions (all OKBs were reformed and renamed long before the fall of USSR), but I will exclude name and image changes

Tagir-1:
I've reverted names and logo changes, but I want to point out than most OKBs were renamed to modern names at around 60s-70s. So they had proper names instead of numbers long before the fall of Soviet Union.

PR #2325, opened 2024-01-23
Fix Von Braun description text in EN-us

jwvanderbeck:
Noticed a small error in the description.  Fixed for English.


hallucinogender:
This is not an error. The Army Ballistic Missile Agency was transferred to NASA and became the Marshall Spaceflight Center.

PR #2204, opened 2023-08-28
Earth Station Program Additions

Teykn:
Stations are heavily underdeveloped in RP-1 right now. Here's some additions that I've come up with.

What is added in this PR as of 8/28/23
- Commits contributed by Pap and Zephyr to the StationScience Branch
- New Experiments that apply to crew and science related Station parts in the first three station nodes (`spaceStationPrototypes`, `spaceStationDev`, `earlySpaceStations`)
- "Long Duration Space Habitation" experiment acts as both an abstraction of human health experiments for MOL and Salyuts 1-5, and double acts as an open-ended Duration tracker across all scenes. Consumes 1000 Units/1 Ton of Research Payload every 30 days. This will definitely need to be adjusted.
- Rest of the Experiments included are based on the MOL Proposal Document (Sections 18 and 20). These include:
  - Ocean Surveillance
  - Ultraviolet Radiation Orthicon
  - Airglow Horizon Spectroscopy
  - Narrow Beam Radio Communication
  - Narrow Beam Laser Communication
  - Vapor Distillation Water Purification
  - Solar X-Ray Detection
- Contract Changes: 
  - Added First Space Station Direct, a MOL style contract
  - Added Early Space Station, a 4-times repeatable contract that serves to complete the 180d experiment. After each completion, the requirements get progressively more difficult. This is not MOL compatible though, as it requires docking. (Should this be changed/MOL style contract be added?)
    - Requirements for each completion are listed below. They definitely will be changed.
      - 2 Crew Stay for 30d, requires positive power production
      - 2 Crew Stay for 30d, requires 100W / 100EC/s power production
      - 3 Crew Stay for 30d, requires 250W / 250EC/s power production
      - 3 Crew Stay for 60d, requires 500W / 500EC/s power production 
  - Above contracts and First Space Station modified in order to utilize Long Duration Space Habitation experiment instead of Duration Parameter
  - Disabled Improved Space Station. Will change it later
  
TO DO:
- Find sources or balance the experiments (Currently all of their science and transmission requirements, and some of the cost, mass, and EC requirements are all placeholder)
- Improve the text
- Modify Improved Space Station to emulate Salyut 6/7 and Skylab requirements. This will include a 1 or 2 year experiment that must be completed which completes the program
- Add Skylab and Salyut experiments (Sources on Salyut experiments ESPECIALLY would be much appreciated)

PR #2326, opened 2024-01-27
Earth Space Station Program Additions

Teykn:
This PR is a clone of #2204 , created due to merging issues. It is still incomplete.

You can read more about what this PR adds in the old one listed above.


Ballatik:
The contract path that Ultrafiltre and I have been going down with this was:

1. Combine first/first direct into one contract that is docking port agnostic.  Check for crew capacity on launch, then check for crew and orbit parameters, then check for habitation science progress. 
2. Change early space station contract to reflect Salyut style stations, requiring new launches instead of requiring docking.  habitation science progress is tracked, with the required completion level determined by $RepeatEarlySpaceStation_Count.  The effect there is that you can do the contract up to 4 times if you launch them for each 30d worth of science, but you can't complete 90% of the science on your first station and then launch 4 new stations for 1 day to get free rep.
3. Add auto-accepted science contract requiring completion of long term habitation 1, with optional rewards for other crewed science at that tier.  This contract also increments the $RepeatEarlySpaceStation_Count to keep the repeatable contracts up to date.
4. (Maybe) Combine all of the add module/supply/rotate crew contracts into one that doesn't use the performance breaking expressions in the data fields.  Pilot existing station to set target vessel, launch a new supply vessel, then check for optional parameters (crew, supplies, etc.) on rendezvous, check for docking, then complete on return of the supply vessel.

lighthea:
I think the disabled contracts should be removed for clarity sake 
I am also unsure of the need of an explicit recrew / resupply contract and I think it is part of the difficulty here to make a cheap launch vehicle and balance other rewards to make it possible

Teykn:
> I think the disabled contracts should be removed for clarity sake I am also unsure of the need of an explicit recrew / resupply contract and I think it is part of the difficulty here to make a cheap launch vehicle and balance other rewards to make it possible

Maybe those contracts could be moved to an ISS program? I feel that those contracts make sense for second gen and later stations (skylab and later), but not for earlier stations

One thing I just remembered is that instead of the “all vessels where” data node, a better solution would be to save applicable stations into a list. Vessel parameter groups have a defineList option iirc

lighthea:
> > I think the disabled contracts should be removed for clarity sake I am also unsure of the need of an explicit recrew / resupply contract and I think it is part of the difficulty here to make a cheap launch vehicle and balance other rewards to make it possible
> 
> Maybe those contracts could be moved to an ISS program? I feel that those contracts make sense for second gen and later stations (skylab and later), but not for earlier stations
> 
> One thing I just remembered is that instead of the “all vessels where” data node, a better solution would be to save applicable stations into a list. Vessel parameter groups have a defineList option iirc

Yes I tested the defineList method and that works. Also I approve of displacing them to a new folder for LateStations

lighthea:
I playtested the prototype and early space station contracts. They seem functional and should work as intended. I think we can try to release it to others in order to balance rewards now

Ballatik:
Adjusted and moved some rep rewards to bring it more inline with crewed lunar overall.  Slightly less rep if you do only the required missions, about the same if you take the more likely path of doing about half of the repeatables, and with about the same upper bound.
Initial testing of Proto tier missions: First station works for both direct and indirect launches.  Science contract optionals work, and habitation experiment tracking iterates the repeatables to keep them on track.  Repeatables correctly iterate based on science completion, though I haven't tested completion yet.

macluky:
I was looking for something like this. Do you need a tester?

Teykn:
> I was looking for something like this. Do you need a tester?

Any and all testers are welcome to try this PR out.

Kallmor:
Since the discord link is invalid I ask it here: How do I get this PR working? On the load screen it says that this is an invalid version of RP-1 and then the mod is disabled.

Teykn:
> Since the discord link is invalid I ask it here: How do I get this PR working? On the load screen it says that this is an invalid version of RP-1 and then the mod is disabled.

Encountered this issue as well. As a quick fix I just replaced the specific folders that had contents modified ("Contracts", "Science", and "Programs") and it worked fine.

Kallmor:
It worked. 

What I found so far: Ocean surveillance experiment has no sample amount. It says depleted on start.
Untitled

Water purification has the wrong description. 

These both seems to be coming from the main branch.

Ballatik:
Thanks for testing it out!
> What I found so far: Ocean surveillance experiment has no sample amount. It says depleted on start. 

Fixed, it shouldn't need or generate a sample now.
> Water purification has the wrong description.

Fixed.
> Can't complete the Collect Early Crewed Station Science contract. It says incomplete but in the save file the fractionComplete is 1 for the mandatory long duration experiment. 

Clarified the contract that it requires LDH2, which has not been completed in your quicksave.  Confirmed that that is a configurable experiment on early station parts.

Kallmor:
There are other 0 sample experiments:
UV Airglow Horizon Photography
Particle Collection
Composite Casting
Gallium Arsenide Crystal Growth

Ocean Surveillance adds 1.2 tons additional mass. I don't know whether that's intentional. 

Damn, I missed that in the Collect Early Crewed Station Science contract the experiment ID is LDH2. This is the prereq. contract for the Second Gen. Space station where you also have to complete a portion of LDH2. are you sure that the ID is right in Collect Early Crewed Station Science?

macluky:
complete install is not working, but copying the changed files over the current version works.
in earlyspacestationsupply the module to be included is rp0longdurationhabit2. we may want a more friendly name for that
rest seems ok. I'll go and try to build one and complete the contracts

macluky:
I can't add the resource payload to the service module, despite adding it in the config:

`@TANK_DEFINITION[Fuselage|ServiceModule]:FOR[RP-0]:NEEDS[RealFuels]
{
	addPayload = true
	addSoundingPayload = true
	addResearchPayload = true
	addProcessedPayload = true
}

// Add payloads to the new tanks
@TANK_DEFINITION:HAS[#addResearchPayload[true]]:FOR[RP-0]
{
	-addResearchPayload = DEL
	TANK
	{
		name = ResearchPayload
		utilization = 1
		fillable = True
		amount = 0.0
		maxAmount = 0.0
	}
}`

Ballatik:
> I can't add the resource payload to the service module, despite adding it in the config:
> 
Did you also define the resource further up in the RP0_Contract_Resources.cfg file?  I just checked all of my service module tanks and they are working, so something must be different with our installs.

GameData/RP-1/Contracts/RP0_Contract_Resources.cfg
macluky, line 1: why not add something like: 	

addResearchPayload = true
	addProcessedPayload = true 

here

macluky:
> > I can't add the resource payload to the service module, despite adding it in the config:
> 
> Did you also define the resource further up in the RP0_Contract_Resources.cfg file? I just checked all of my service module tanks and they are working, so something must be different with our installs.

Hmm, its now showing up in the pods but not in the procedural tanks, i'll double check the files

njits23:
I decided to give this PR a test. So far I've planned out all of the prototype stations and am halfway done with them, which leads me to ask if it's really necessary to make you do six months of the multi-week science. If you're launching small 11t prototype stations, you need to do three launches to complete all the regular science and 3 months of the multi-week science. I don't feel it adds much to gameplay after that to make the player do it three more times. Cutting it back to only four months (possibly with a reduction in funding) would make more sense to me. Is there a specific reason six months was chosen?

njits23:
I've completed all of the prototype stations, and have started on the second generation stations. The multi-month habitation experiment is misconfigured to require four crew, making it impossible to complete on something like Salyut 7, which only allows three crew. The issue lies in the associated part module, which specifies four crew. The experiment definition only specifies three.

GameData/RP-1/Science/Experiments/CrewScience/StationExperiments.cfg
```diff
@@ -0,0 +1,1013 @@
...
+{
+	MODULE
+	{
+		name = Experiment
+		experiment_id = RP0longDurationHabit2
+		ec_rate = 0.5
+		data_rate = 50 //Placeholder
+		@data_rate /= 63072000 //2 Years
+		requires = CrewMin:4
+		resources = ResearchPayload@0.0009645 //Consumes 2.5T of payload every 30 days. Placeholder
```
njits23, line 436: 4 crew should be changed to 3 crew.
2.5t of payload every 30 days is way too much. Resupplying 10 tons of payload every 120 days is entirely unreasonable. This amount should probably not be any higher than the prototype space stations. Even at 4 tons per 120 days it's still rather high.

A Progress resupply craft can only bring up 140d of supplies for three kerbals (oxygen, food, water) on a salyut 7. There would be no space left for any sort of research payload.

140d of every type of life support supply and research payload at a consumption rate of 1t of research payload per 30d would take just over 10kl, and weigh 8.21t in a SM-IV service module. Still higher than you could stuff in a Progress, meaning you'd need to be launching a progress about every 70 days to keep the station supplied, which seems excessive.

If you instead build a custom probe to supply this, and give it about 500m/s of dV, it'll weigh 10.5 tons, which doesn't seems more reasonable, but I still believe that the consumption rates, volumes, and masses of all consumables should be looked at to make supply via progress to a salyut-style station reasonable achievable.

CXG-2827:
some Typos:

Vapor Distillation Water Purification description text looks to be a copy/paste from "Narrow Beam Radio Communication".

First Space Station contract states as the objective "room for at least 2 crew" but the actual requirement to trigger needs 3.

njits23:
> some Typos:
> 
> Vapor Distillation Water Purification description text looks to be a copy/paste from "Narrow Beam Radio Communication".
> 
> First Space Station contract states as the objective "room for at least 2 crew" but the actual requirement to trigger needs 3.

Are you sure you are on the latest version of this PR? The text for the Vapor Distillation Water Purification experiment was fixed three months ago in [81ec592](https://github.com/KSP-RO/RP-1/pull/2326/commits/81ec592b2b9eeeb17d233a3725fcce7ffae5551f).

CXG-2827:
I downloaded the folders on 4/27. I've give it a redo.

njits23:
I've gotten most of the way through the second generation space station contracts and science now. I did end up modifiying my config files to only consume 1t of research payload for the multi-month habitation experiment. I also noticed that the second generation contract requires room for five astronauts, which I find strange considering the multi-month experiment (when fixed) only requires three. This means you can't complete the contract using only a single Salyut 6 or 7 module, which I don't think should happen. This is especially frustrating because the more advanced experiments are only available when using a Salyut or Skylab part. The experiments cannot go into something like the mobile processing lab, making it not possible to use a custom station either. 

It also appears to me that not using Salyut or Skylab and building a custom station would end up heavier, which is maybe something that should be looked at also.

Aside from that, the funding curve doesn't really fit well with the contract flow. When doing the prototype stations you are absolutely flooded with money, only to end up back in the negative income when you have to start flying the more expensive second generation stations. I think a two-hump funding curve would fit this program much better. Overall funding probably doesn't need to change, but a small hump for the first generation stations and a slightly bigger hump starting about 50% of the way into the program and peaking at 66% would likely help solve this imbalance.

I didn't have any trouble with any the second tier science experiments, and the science reward values are probably fine. However, the only confidence this program gives you comes from this science. There's zero contracts which award confidence. In my opinion there should probably be an additional source of confidence.

I did constantly have issues with the resupply contracts for the second generation stations not completing properly. They would get confused about what craft I was sending, incorrectly detect craft parameters, and not detect that I was landing with a different older craft that was sent up earlier. This means you can't send up a fresh crew and return the older crew in their original craft. The contract parameters also mean that uncrewed probes purely intended to send supplies up cannot complete these contracts, as the resupply contracts require returning to earth.

CXG-2827:
I managed to complete the First Station contract in less than 15 days, only consuming 418.5 Research Payload. I launched Salyut with the station on November 18th, and the Soyuz on the 21st. Contract flagged as complete on December 4th, which seems to align with the day the Salyut was put into orbit.

Screenshot attached, you can even see that the "Crewed duration record of 14 days" is still timing down, and the "multi-week space habitation" experiment has only collected ~12.5 days of science.
screenshot26

njits23:
I've fully completed the program now. Other than what I've already mentioned, I did run into one other issue, but that's not related to any of this PR's changes. If you timewarp too fast (faster than 1000x) in the space center view, Kerbalism gets confused about how much EC your station should have, lets it run out of power, and kills your crew. You can have twice as much charge rate in the sun as drain rate on the dark side and you'll still run out of power at high warp. Fixing that is not in the scope of this PR, but perhaps some of the contracts or program description should have a warning about not running high warp speeds when there's crew in a station.

Teykn:
Will look over the comments more closely later on. Putting a quick comment here. Very much appreciate the feedback on this, as my memory of post-moon boots needs to be refreshed.

Regarding the durations: the 180d duration was meant to be similar to the early saylut stations (1-5) and the 730d/2 yr was meant to be similar to later salyuts (6 and 7)
Will def adjust the Research Payload properties.

CXG-2827:
Is there a way the experiments can be configured to work in any of the station parts, much like how more science experiments become available in avionics when tech is unlocked? With the current capsule tier method, EarlyStation science can only be performed if the station is made using the parts in the Early Space Stations tech node. It is a bit limiting and doesn't give much flexibility in the design of the station.

Teykn:
> Is there a way the experiments can be configured to work in any of the station parts, much like how more science experiments become available in avionics when tech is unlocked? With the current capsule tier method, EarlyStation science can only be performed if the station is made using the parts in the Early Space Stations tech node. It is a bit limiting and doesn't give much flexibility in the design of the station.

It might be possible to? I remember trying to do so by using an `|` with the two tags, but that breaks the experiment. Haven't tried making a new tag that encompasses all of the station parts though.

njits23:
I tried reading through the documentation for ModuleManager, but the documentation isn't very clear about what you can do with the :HAS[] operator. Doing some experiments myself, it appears that it's not possible to match multiple capsuleTier tags with it at the same time. It seems to me that a new tag encompassing all crew-holding station parts would indeed be required.

CXG-2827:
Ive tested out changing the capsule tier to be the same for all station parts, and it correctly shows the crew sciences based on unlocks. This was checked in 2 save files before and after unlocking Early Space Stations. However it will need a compromise in the configure.cfg as both station tiers will have the same sample capacity and science slots. Maybe split the difference and have 5 experiment slots and +5 sample capacity?

Teykn:
> Ive tested out changing the capsule tier to be the same for all station parts, and it correctly shows the crew sciences based on unlocks. This was checked in 2 save files before and after unlocking Early Space Stations. However it will need a compromise in the configure.cfg as both station tiers will have the same sample capacity and science slots. Maybe split the difference and have 5 experiment slots and +5 sample capacity?

Multiple tags on a part should be possible I think, so it may be possible to have the "all stations" tag not modify sample capacity & still have the tier specific tags do their thing?

CXG-2827:
Some more balance comments. Fitting out a Progress using RN parts I can manage to fit enough supplies and research payload for 60d with almost no volume to spare, and tight delta-v margins to reach a station in 300km orbit. At a minimum, the second generation contract can be completed with 12 Progress launches if no other crew rotations/visits were launched. If the research payload is reduced to 500 units/30days the Progress can then fit supplies and research payload for 90 days.

CXG-2827:
I'm running into issues where the Early Station Resupply does not detect rendezvous or docking. If I switch between the station and supply craft while within rendezvous range it checks off the rendezvous parameter. Afterwards when I dock, it does recognize the completion of the docking. I have sent up resupplies twice so far. First time was when I noticed the rdvz and docking did not recognize days after docking, and I force complete the contract. The second resupply was when I noticed the switching between craft fixed the param checks. 

Also, the 30 day's worth of multi-month experiment did not check off as complete, even after using up 60 days worth of that research payload.

Ballatik:
> I'm running into issues where the Early Station Resupply does not detect rendezvous or docking. If I switch between the station and supply craft while within rendezvous range it checks off the rendezvous parameter. Afterwards when I dock, it does recognize the completion of the docking. I have sent up resupplies twice so far. First time was when I noticed the rdvz and docking did not recognize days after docking, and I force complete the contract. The second resupply was when I noticed the switching between craft fixed the param checks.

That situation sounds like it hadn't defined the target station before you did the rendezvous.  Had you piloted the station at any point after accepting the contract but before doing the rendezvous?

> Also, the 30 day's worth of multi-month experiment did not check off as complete, even after using up 60 days worth of that research payload.

That step is still using the placeholder values copied from the prototype station contract which would work out to 8 months.  Since we can't really track starting science completion, we should probably have that step check research payload depletion instead.

CXG-2827:
> 
> That situation sounds like it hadn't defined the target station before you did the rendezvous. Had you piloted the station at any point after accepting the contract but before doing the rendezvous?
> 
I have not piloted the station after the contract was accepted, nor before the rendezvous when this occurred. 

Sending up a new supply craft today and I confirm the above to still hold true. But, reloading the save before launch, I piloted the station before returning to the launch site. When going to the map vie, the station now has the strobing square indicator and the contract window now checks off "Control a space station with the required parameters".

Ballatik:
> Sending up a new supply craft today and I confirm the above to still hold true. But, reloading the save before launch, I piloted the station before returning to the launch site. When going to the map vie, the station now has the strobing square indicator and the contract window now checks off "Control a space station with the required parameters".

That's working as intended then.  The first contract parameter is to control the target station so that the contract knows what station you are supplying.

njits23:
I've found another downside to the way #capsuleTier is currently defined and used as the way to determine which parts should get access to the experiments. It means that only parts which are explicitly placed in the #capsuleTier tag can have experiments mounted in them. This means that a significant number of SSPX parts are excluded, like the 3.38m habitation module and others.

I tried replacing the explicit part list at the top where it adds the %capsuleTier tags with @PART[*]:HAS[#TechRequired[earlySpaceStations],#CrewCapacity[>0]]:BEFORE[RP-0-Kerbalism], which should add the experiments to any crewed part from that tech node. However, doing that, even when using only #TechRequired[earlySpaceStations] stops it from adding the experiments to any of those parts. I'm not seeing any errors in KSP.log, so I have no idea where it might be going wrong. I don't know how to check part tags once in game, and the ModuleManager wiki doesn't even say that it might be a possibility.

njits23:
I've put in a pull request (https://github.com/Teykn/RP-1/pull/1) that should resolve the experiment distribution issue I ran in to.

Teykn:
> I've put in a pull request ([Teykn#1](https://github.com/Teykn/RP-1/pull/1)) that should resolve the experiment distribution issue I ran in to.

Merged, definitely a lot less verbose and more convenient for when more parts get configs later on. Also gave you commit access

njits23:
It also looks like it's possible to integrate RP-1's main branch changes without any merge conflicts. Should we do that? It makes it easier to install this PR for anyone looking to do so.

njits23:
> Is there a way the experiments can be configured to work in any of the station parts, much like how more science experiments become available in avionics when tech is unlocked? With the current capsule tier method, EarlyStation science can only be performed if the station is made using the parts in the Early Space Stations tech node. It is a bit limiting and doesn't give much flexibility in the design of the station.

Regarding this, none of the current capsule crew experiments work this way. I also think it makes sense to have more advanced experiments require more advanced station parts. I don't think it's required to give newer stations access to the older experiments, as with some planning you can complete all of the prototype-tier experiments during the prototype station launches.

njits23:
> I also decided to build out a table to graph the spending of the program to compare against the funding curve. The data below is based on maintenance costs of the Space center facilities, a 325 and 720t LC, associated costs for mission trainings, and for the rockets. The red vertical line in the graph shows where I completed the Prototype capstone. I believe the current "BimodalBackloadedFundingCurve" might not work well, as compared against my spending curve. I think a flat funding curve might be a better choice.

Your graph of funding expenditures is quite interesting. In my opinion a flat funding curve isn't the best fit either, considering your expenditures were far from flat. I think the funding curve that would best fit to the stations program doesn't currently exist in RP-1, so I'm going to add a MildRampdownCurve specifically for this program. This would allow players to pick up the stations program instead of crewed lunar and still have sufficient funding to get all the required parts and build appropriately sized LCs.
image
The final funds/month curve could do with maybe a bit more smoothing, but it's really hard to get the funds/month curve to be smooth by configuring the %funding paid at a certain time curve, as the slope changes in that curve aren't very obvious.

siimav:
Some notes:
* 'Early' and 'Second Generation' appear to get used interchangeably which is a bit confusing.
image
* That's a mildly annoying unit. Maybe add units per day into experiment description.
image
* Some contract descriptions are missing the Required/Optional classifications.

njits23:
> * That's a mildly annoying unit. Maybe add units per day into experiment description.

Payload per 30 days is already in the description of both multi-week and multi-month space habitation science. I'm not entirely sure changing that to payload/day would make more sense.

njits23:
> * Some contract descriptions are missing the Required/Optional classifications.

This appears to only happen when looking at optional contracts that cannot be accepted yet because their requirements aren't met. This is no different from other optional contracts from other programs, and thus not something for this PR to fix:
image

And for the space stations, to prove it works:
image
image

siimav:
> and thus not something for this PR to fix

Actually it could be fixed. The logic is that if description contains variables, then CC will use genericDescription instead for not-yet-generated contracts. So in these cases genericDescription should have the same information minus the dynamic content.

> Payload per 30 days is already in the description of both multi-week and multi-month space habitation science. I'm not entirely sure changing that to payload/day would make more sense.

Right, it's in the contract description but not in experiment description. If you don't remember that value then you have to leave VAB to bring up the mission control overlay again. Showing for 30 days instead of per day is also fine.

njits23:
> Right, it's in the contract description but not in experiment description. If you don't remember that value then you have to leave VAB to bring up the mission control overlay again. Showing for 30 days instead of per day is also fine.

Ah, I see where it went wrong. The description shown in the configure crew science screen comes from a different place than I thought it did, and thus the descriptions are wrong/incomplete. Will fix.
image

siimav:
Most experiments have a warning about them taking 30 days and this doesn't appear to be true at all. Probably a copy-paste error?
image

njits23:
Do note that some experiments will take longer than their stated duration, such as airglow horizon spectroscopy, because their conditions to run are not met continuously. Will double-check other durations later.

GameData/RP-1/ProgramHandlerSettings.cfg
```diff
@@ -147,6 +148,17 @@ PROGRAMHANDLERSETTINGS
...
+			VALUE // MildRampdownCurve
+            {
+				key = 0 0 0.92747 0.92747
+				key = 0.02 0.0185494 1.033673 1.033673
+				key = 0.04 0.0413469 1.205534 1.205534
+				key = 0.2 0.2447376 1.233564 1.233564
+				key = 0.4 0.4839247 1.065012 1.065012
+				key = 0.7 0.7641512 0.8601255 0.8601255
+				key = 1 1 0.7861626 0.7861626
+				key = 2 1.2 0 0
```
siimav, line 161: Mixed tabs and spaces

GameData/RP-1/Contracts/RP0_Contract_Resources.cfg
```diff
@@ -79,6 +105,46 @@ RESOURCE_DEFINITION
...
+	{
+		name = ResearchPayload
+		utilization = 1
+		fillable = True
+		amount = 0.0
+		maxAmount = 0.0
+	}
+	TANK
+	{
+		name = ProcessedPayload
```
siimav, line 117: Cant see this used anywhere. Remove?

GameData/RP-1/Contracts/RP0_Contract_Resources.cfg
```diff
@@ -66,6 +66,32 @@ RESOURCE_DEFINITION
...
+  unitCost = 2.0
+  flowMode = ALL_VESSEL
+  transfer = PUMP
+  isTweakable = true
+  volume = 1
+}
+
+RESOURCE_DEFINITION
+{
+  name = ProcessedPayload
```
siimav, line 86: Cant see this used anywhere. Remove?

siimav (changes requested):

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -961,38 +961,45 @@ RP0_PROGRAM
...
 
-disabled_RP0_PROGRAM
+RP0_PROGRAM
 {
-	name = EarthSpaceStation
+	name = EarlyEarthSpaceStation
 	isHSF = true
-	title = Earth Space Station (WIP and Very Buggy)
-	description = An Earth space station will enable long-term human exploration of space and provide benefits to people of Earth. This Program tasks you with launching and maintaining Space Stations in Earth orbit and continuously crewing them with an increasingly longer stay durations. By completing the program, you will prove that long-term space habitation is possible which will open up crewed explorations of Venus, Mars, and beyond. NOTE: This program isn't functional enough to be completed and only has two unique contracts. Both of those contracts are also very buggy and may not complete even if the requirements should have been met.
+	title = Early Earth Space Stations (WIP)
```
siimav, line 968: WIP needs to be removed before shipping.

GameData/RP-1/Contracts/RP0_Contract_Resources.cfg
njits23: Are you specifically referring to the only ProcessedPayload part, or both Payload types? Because ResearchPayload is something that is used, but maybe you mean it doesn't need to be defined here?

siimav:
I consulted with smarter people than me and they thought it would be best to create 2 versions of this program.
1) Alternative to Crewed Lunar - becomes available after Manned Orbit and has similar timescale. Probably a bit less funding than MoonBoots though since the infrasctructure requirements are a lot lower. Funding curve should also be a bit smoother than Crewed Lunar since it requires less up-front infrastructure.
2) Taken after Crewed Lunar - the current duration is probably quite fitting for this but I think a completely flat funding curve is the best here since all the infrastructure is already in place.

The reasoning is that it's unlikely we can square the circle of balancing for both when the player needs to build launch infra and tech, and when they're reusing lunar hardware.

siimav:
It's been suggested that maybe we should instead implement a system that allows changing the funding and duration values of another contract when one gets accepted. 🤔

njd80:
Regardless of the choice of what to do with these Programs (which seems like it might take time), there are new crewed science experiments here - is there any way they can be decoupled from the program(s) and released as selectable crew science? A new source of crew science like this makes LEO stations desirable even without program(s)

siimav:
Here's my suggested way forward with this program:
* Convert it to be able to be accepted after completing Crewed Orbit.
* Remove the now-conflicting funding curve and replace it with `Mid`. IMO it should be less peaky that Crewed Lunar due to requiring less infrastructure.
* Change nominal duration to 13 years. It actually requires tech nodes that are a bit deeper in the tech tree. On the other hand progression with landing and engine tech isn't as important.
* Other values from the spreadsheet:
funding = 14300000
slots = 4
normal conf = 5250
fast conf = 10500
repDeltaOnCompletePerYearEarly = 920
repPenaltyPerYearLate = 920

njits23:
I think players should have completed Adv. Crewed Orbit before being able to accept the stations program, because stations require rendezvous and docking. This idea is reinforced by the fact that he non-direct first crewed lunar landing is locked behind the docking contract from that program.

Regarding the funding curve, Mid seems like a terrible choice:
image
The funding curve this PR adds is slightly less peaky, as well as more front-loaded, because that's when all the LC investments are being made. See the earlier comment with the graph: https://github.com/KSP-RO/RP-1/pull/2326#issuecomment-2211781132

Thirteen years nominal is rather a rather excessive amount of time. Even the current program duration is perfectly fine if you accept it sometime in 1963. Don't forget that the nominal duration is on fast, and thus on normal you have 10.5 years to complete the program. If you do choose to go faster, it requires a somewhat single-minded research effort, which is in line with doing other programs on high speeds. I could see the nominal duration being 8/9 years instead of 7, which would give you a normal speed duration of 12/13.5.

Regarding the other parameters, a funding increase for the program hasn't seemed required from any of my testing. Even on moderate everything is affordable. Unless you suggest a funding increase because launching empty stations now also uses the human-rated tag, increasing their cost.

siimav:
It's an alternative path to picking Crewed Lunar so it has to be available at the same time. I guess it could make more sense to require Adv Crew program being active instead of just having completed Crewed Orbit.
I'd like to point out that the numbers you based your funding curve on are from a career that had progressed to mid/late Crewed Lunar. As hinted before - there's gonna be another variant of this program when taken alongside or after Crewed Lunar which will have way shorter duration and probably also use tha flat funding curve.

siimav (approved):

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -961,38 +961,49 @@ RP0_PROGRAM
...
-	description = An Earth space station will enable long-term human exploration of space and provide benefits to people of Earth. This Program tasks you with launching and maintaining Space Stations in Earth orbit and continuously crewing them with an increasingly longer stay durations. By completing the program, you will prove that long-term space habitation is possible which will open up crewed explorations of Venus, Mars, and beyond. NOTE: This program isn't functional enough to be completed and only has two unique contracts. Both of those contracts are also very buggy and may not complete even if the requirements should have been met.
+	title = Early Earth Space Stations
+	description = An Earth space station will enable long-term human exploration of space and provide benefits to people of Earth. This Program tasks you with the exploration of prototypal and later, more mature solutions for low earth orbit long term space habitat. You will have to complete long term experiments, recrew and refill your station(s) or send new ones when needed.
 	requirementsPrettyText = Complete contracts X and Y
 	objectivesPrettyText = Do this and that
```
siimav, line 971: We probably should add something better here

GameData/RP-1/Contracts/Early Space Stations/Early Station Supply.cfg
```diff
@@ -0,0 +1,200 @@
...
+			minPeA = 250000
+			maxApA = 500000
+			targetBody = HomeWorld()
+			title = Keep the station in a stable orbit with a perigee greater than 250 km and an apogee less than 500 km
+		}
+		PARAMETER
+		{
+			name = Multi-Month Habitation Analysis
+			type = RP1CollectScience
+			title = Conduct Multi-Month Habitation Analysis of your astronauts for @/durationText days
```
siimav, line 191: This can be quite misleading. First of all, the contract doesn't check for any number of days at all. Instead it just looks that a predefined % of the experiment has been completed. The first level appears to complete somewhere around 243 days instead. For example if you send a resupply 120 days in then you end up waiting for another 123 days for it to tick off. Maybe it would be best to mention the experiment completion % instead? Or say that the experiment has to have been running for 243 days out of 730.

PR #2327, opened 2024-01-27
Add on-pad vessel repairs

siimav:
Creates a new option into the vessel recovery dialog that allows queueing up a repair of all failures. Only available for VAB-built vessels while in prelaunch state. This operation is blocking (i.e cannot build another vessel at the same time in that LC) and currently takes 1/7.5 of rollout time. No additional cost other than what is paid for engineer salary.

Things included in a *repair*:
* Reset all kinds of failures (except Explosion! for [hopefully] obvious reasons)
* Reset engine runtimes to 0
* Restore ignition counts to initial values

Things NOT included in a *repair*:
* Increase TF per-flight du cap
* Increase TF initial data so that the engine would become more reliable on attempt 2

**TODO**
* [x] 16x16 png wrench icon

Depends on https://github.com/KSP-RO/TestFlight/pull/284
Soft-depends on https://github.com/KSP-RO/RealFuels/pull/336


jwvanderbeck:
So now you get 0 data for pre-launch failures? That seems a bit harsh to me.  Should get t least some, even if it is less than the default, no?

siimav:
No, you get 1000 but it isn't possible to earn anything on top of that after repairing the failures and actually launching. There's a 1000 du per-flight cap that still applies here.

PR #2335, opened 2024-02-15
Fix vessel transfers

Ballatik:
Remove dissassociateVesselsOnContractCompletion on contracts that also have DestroyVessel on completion.  If the dissasociate happens first, there's nothing to destroy.


Ballatik:
Aimed at this issue here: https://discordapp.com/channels/319857228905447436/620690446540341261/1205648134148202506

PR #2337, opened 2024-02-25
Balance-neutral cheap astro

NovemberOrWhatever:
Just did this in the web UI, haven't tested yet, but it makes the base astronaut complex cheaper to maintain and increases maintenance costs for other facilities to keep balance neutral while annoying no planes people less

### DIFF
--- a/GameData/RP-1/CustomBarnKit.cfg
+++ b/GameData/RP-1/CustomBarnKit.cfg
@@ -3,7 +3,7 @@
 	@ASTRONAUTS
 	{
 		@levels = 5
-		@upgrades = 60000, 30000, 150000, 300000, 1000000
+		@upgrades = 30000, 30000, 150000, 300000, 1000000
 		@upgradesVisual = 1, 1, 2, 2, 3
 		
 		@recruitHireBaseCost = 5000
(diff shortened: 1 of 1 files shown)


siimav:
I might be wrong but doesn't this change upgrade costs and times for all the facilities involved?

NovemberOrWhatever:
> I might be wrong but doesn't this change upgrade costs and times for all the facilities involved?
Ah, yeah. The first cost I believe is the cost for the one you start with, so it only effects maintenance (or how expensive it would be if a building gets destroyed) and upgrade time, not upgrade cost. But I also might be wrong

siimav:
I vaguely remember the actual cost being the delta between 2 levels.

Clayell:
I suppose this makes sense for balance reasons to not throw anything off, makes it a little less annoying than seeing astronaut complex costing 1/4th of your infrastructure cost when its not in use.

Still think the hangar should be built only when needed for x-planes D:

siimav:
I had a look and this change has a rather significant effect on upgrade times.
2) 45d 21h -> 59d 19h
3) 158d 6h -> 177d 18h
4) 204d 3h -> 213 6h
Not sure that's the tradeoff we want to make.

siimav:
The idea is fine. It's just that the implementation has some issues.

PR #2346, opened 2024-04-04
Rebalance Optional Hypersonic Flight Contracts

njits23:
The current optional hypersonic flight contracts require speeds of 2300 m/s and 2600 m/s. The 2300 m/s contract is achievable, but only when using the non-inline X-15 cockpit. No parts exist with an aerodynamic shape, or that can be changed into an aerodynamic shape, which can withstand the temperatures encountered at those speeds.

In real life, no X-15 ever flew faster than 2020 m/s (Flight 188). That flight also essentially ruined the aircraft from thermal damage.
Because of this, believe that the differences in speed requirements should be dropped in favour of duration requirements.

I have changed the optional contracts to instead require the player to maintain 2000 m/s for longer, at three minutes for the first contract, and four minutes for the second. These are still difficult requirements to hit, but fall within the thermal limitations of something like the procedural nosecone.

I believe that these changes ensure that the confidence awarded by these contracts (150 on normal difficulty) is now more in-line with the craft required to perform them.

### DIFF
--- a/GameData/RP-1/Contracts/X-Plane Research/XPlanesHypersonic.cfg
+++ b/GameData/RP-1/Contracts/X-Plane Research/XPlanesHypersonic.cfg
@@ -5,7 +5,7 @@ CONTRACT_TYPE
 
 	title = X-Planes (Hypersonic)
 
-	description = <b>Program: X-Plane Research<br>Type: <color=green>Required</color></b><br><br>Design, build, and launch a crewed plane capable of hypersonic atmospheric flight of at least  @/minSpeedMPS km and return home safely. Optionally, landing back at the runway will award extra reputation. If playing from the Cape, the Space Shuttle runway provided by the RSS-CanaveralHD mod is also a valid landing target.
+	description = <b>Program: X-Plane Research<br>Type: <color=green>Required</color></b><br><br>Design, build, and launch a crewed plane capable of hypersonic atmospheric flight of at least @/minSpeedMPS m/s below 40km and return home safely. Optionally, landing back at the runway will award extra reputation. If playing from the Cape, the Space Shuttle runway provided by the RSS-CanaveralHD mod is also a valid landing target.
 	genericDescription = Design, build, and launch a crewed rocket or plane to put a person into high atmosphere above a specific altitude and return home safely.
 
 	synopsis = Launch a crewed vessel to @/minSpeedMPS m/s below 40km altitude.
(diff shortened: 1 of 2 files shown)


Capkirk123:
Looks good to me. The X-plane contracts should not require something so far beyond the limits of available parts.

lpgagnon:
> In real life, no X-15 ever flew faster than 2020 m/s (Flight 188). That flight also essentially ruined the aircraft from thermal damage.

it'd be useful to find data on _how long_ the x-15 maintained that top speed. intuitively, the "2 minutes" part seems more unrealistic than the speed part to me...

njits23:
I did not want to drop the speed on the non-optional contract, which requires 2000 m/s, because that contract seemed fairly well balanced. However, dropping the speed requirements of the non-optional contract would put it more in-line with the speeds experienced during typical X-15 flight. 

From my testing, the procedural nosecone will survive speeds of 2200 m/s indefinitely, so requiring a higher speed than the non-optional contract is possible. Therefore, a possible further modification to the optional contracts could be to first require 2100 m/s at two minutes, and then 2200 m/s at three minutes.

njits23:
From what I can find on how long flight 188 stayed above 2000 m/s, the answer can't be more than at most 10-20 seconds. The engine was not kept lit during this time according to mission reports.
https://www.thisdayinaviation.com/tag/flight-188/
https://www.nasa.gov/wp-content/uploads/2015/01/601242main_X15ExtendingFrontiersFlight-ebook.pdf

> The flight plan showed that X-15A-2 would weigh 52,117 pounds at separation, more than 50%
heavier than originally conceived in 1954. As the X-15 fell away, Knight lit the engine and set up
a 12-degree angle of attack resulting in about 1.5 g in longitudinal acceleration. As normal
acceleration built to 2 g, Knight had to hold considerable right deflection on the side stick to keep
X-15A-2 from rolling left due to the heavier liquid-oxygen tank. When the aircraft reached the
35-degree planned pitch angle, Knight began to fly a precise climb angle. The simulator had
predicted a maximum dynamic pressure of 540 psf, remarkably close to the 560 psf measured
during the rotation. Knight maintained the planned pitch angle within ±1 degree. [342]
Knight jettisoned the external tanks 67.4 seconds after launch at Mach 2.4 and 72,300 feet. Tank
separation was satisfactory, but Knight described it as "harder" than it had been on flight 2-50-89. The parachute system performed satisfactorily and the Air Force recovered the tanks in
repairable condition. Free of the extra weight and drag of the external tanks, the airplane began
to accelerate quickly, and Knight came level at 102,100 feet. As Knight later recalled, "We shut
down at 6,500 [fps] and I took careful note to see what the final got to. It went to 6,600 maximum
on the indicator." [343]

siimav:
Ehh, YOLO.

PR #2363, opened 2024-05-11
Note the format for putting in dates for simulations

Clayell:
**PLEASE CHECK THAT THIS IS ACTUALLY THE RIGHT FORMAT**

The format for dates varies around the world, so it is important to actually mention the format that this date is supposed to be in instead of just assuming.


siimav:
I'm not entirely sure this is necessary. Why not just change the date example to values that isn't ambiguous?

Clayell:
> I'm not entirely sure this is necessary. Why not just change the date example to values that isn't ambiguous?

Oh sure, that could also work.
Is it yyyy-mm-dd btw?

siimav:
> Is it yyyy-mm-dd btw?

Of course! Why would anyone order date components in anything other than ascending or descending?

Clayell:
Of course, why would anyone do that. 👀

PR #2372, opened 2024-06-01
Fix missing contract Sat upgrade in early avionics node

anmpereira:
EarlyAvionics upgrade for Com and Nav Sat's was missing from the tech tree due to a typo in the field techRequired in PayloadLevels.cfg

Tested this change in my local copy and it fixed the issue.


anmpereira:
Note that sats use 0.2W as described in simulation mode, but the KO window in the VAB still shows a drain of 30W. Maybe this could be fixed further

PR #2382, opened 2024-06-17
Forcefully reset RA CommNet when upgrading TS

siimav:
Depends on #2381, https://github.com/KSP-RO/RealAntennas/pull/19


Capkirk123:
Hope I didn't break anything with the conflict resolution...

siimav:
Looks correct to me.

PR #2385, opened 2024-06-20
Fixing descriptions

Clayell:
Typo in 3-satellite constellation


Clayell:
Add x10 multiplier for the com sat payload for molniya contracts to make up for the fact that about 10x payload can be put into a leo than a molniya orbit

siimav:
Adding different kinds of changes into the same PR is a very bad practice.

Clayell:
Woops, meant to make this into another branch, sorry. Should I revert this and make it into a different pull?

siimav:
Yes

PR #2388, opened 2024-06-25
Buff Molniya repeatable contract reward

Clayell:
Pretend to have 10x the payload for the reward, as you can approximately get 10x of com sat payload into a low earth orbit for the normal com sat repeatable.

Fixes the problem of reputation and confidence being remarkably low compared to the com sat repeatable. (40 conf compared to ~500 conf) https://discord.com/channels/319857228905447436/1115321536807718992/1255099197863231550


Capkirk123:
Sure

PR #2389, opened 2024-06-25
Preliminary part configs for giuliodondi's SSS fork

ThePuzzlemaker:
This adds preliminary part configs for some parts missing in @giuliodondi's [SSS fork](https://github.com/giuliodondi/Space-Shuttle-System-realistic-aerodynamics). Note that some of these parts (specifically the payload bay items) have names that could be non-unique and I'm not sure what would happen to them if a conflict occurred. It might be worth checking with Giulio to see if it's worth renaming them to be unique to the mod.

Feel free to offer balancing suggestions or anything I missed!


Capkirk123:
Space Shuttle System is already configured. I suspect this particular fork of SSS fixes a lot of the notoriously bad spelling present in the original SSS, which means the RP-1 configs are no longer applied. It should mostly just be a matter of matching the new parts to the old.

PR #2394, opened 2024-07-08
Make an independent cooldown time for EarlyNavSat, EarlyMolniyaSat, MolniyaSat, and TundraSat

Clayell:
It seems kinda silly that EarlyNavSat and EarlyMolniyaSat go off of the cooldown time for EarlyComSat, and same with MolniyaSat and TundraSat going off of the cooldown time for GEORepeatComSats. This will break those contracts off from the other unrelated repeatable contracts and allow for the player to properly switch between repeatables without having to repeat a contract that they just did.

If I recall correctly, the purpose of the cooldown is to prevent a player from spam-repeating one contract over and over again, so I believe this is still in line with that.


Capkirk123:
Has conflicts

Clayell:
Ah, thats probably due to me changing the description in 2385. Give me a second

Clayell:
Nevermind

Capkirk123:
I'm not sure if this violates the spirit of the repeatable contract cooldowns. Geostationary, Molniya and Tundra contracts have similar payload and deltaV requirements. Moving them onto independent timers just encourages launching triple the missions for each, instead of choosing which to complete for your 90-day allotment.

Clayell:
> I'm not sure if this violates the spirit of the repeatable contract cooldowns. Geostationary, Molniya and Tundra contracts have similar payload and deltaV requirements. Moving them onto independent timers just encourages launching triple the missions for each, instead of choosing which to complete for your 90-day allotment.

Hm, I disagree. I imagine the cooldown timer as the 'demand' for these types of satellites, where if you've just launched one type of that satellite then companies won't want it for awhile and will want other types.
I do see the potential problem you are bringing up though, where you could get a lot of confidence by building multiple LC's and running all 3 of those missions over and over. I was more thinking about it from a single LC standpoint, so I didn't see that. I'm not sure on how big a problem this would be, but a potential solution to it could be interlocking the timers with each other. (for example, resetting one of the timers to 0% from 100% would decrease the other timers by 50%)

siimav:
Yes, Kirk is correct.
We definitely do not want 3x as much grind.

PR #2398, opened 2024-07-13
Buff reward of far side landing optional and sample return contracts

Clayell:
Currently, the reward for the lunar landing optional contracts are 200 reputation each. However, the sample return contract, which is far more difficult, gives the exact same reward, and furthermore the far side landing optional contract gives less of a reward?
I've fixed this problem by scaling the rewards roughly on difficulty. (for reference, Luna 16 (sample return) weighed 5.7t wet, while Luna 13 (lander) weighed 1.6t wet)

### DIFF
--- a/GameData/RP-1/Contracts/Lunar Surface Probes/MoonLandingFarSide.cfg
+++ b/GameData/RP-1/Contracts/Lunar Surface Probes/MoonLandingFarSide.cfg
@@ -29,7 +29,7 @@ CONTRACT_TYPE
 	rewardScience = 0
 	rewardFunds = 0
 	failureFunds = 0
-	rewardReputation = 150
+	rewardReputation = 250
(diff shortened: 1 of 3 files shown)


siimav:
400 for Return contract sounds good.
Not so sure about the 300 for optional far side landing though. Especially considering that you left the required contract at the same 150. Honestly I'd just set both of those at 200.

Clayell:
> Honestly I'd just set both of those at 200.

Perhaps 250 would be more accurate? It was only accomplished in 2019, and it requires a relay and partial orbit before landing. Setting it to the same reward as the normal lunar landing contract doesn't make sense in my opinion.

> Especially considering that you left the required contract at the same 150.

Woops, that's my bad.

siimav:
Ehhhh, whatever.

PR #2422, opened 2024-08-13
Crewed Mars Program

CXG-2827:
Submitting this program for feedback.

A Crewed Mars Program which unlocks after the completion of the Crewed Lunar Program, Mars rover contract, and the 2nd Generation Science contract (includes the 2yr space habitation experiment).

**Program Contract Structure:**
The idea is to allow the program to be complete one of two ways. 

Option1: Crewed Mars Landing
- Mars orbit.
- Mars crewed landing with a surface stay of a specified duration.
- Collection of Mars Surface samples, and Deep Surface Samples.

Option2: Similar to Project Red Rocks/Mars Base Camp
- Mars Orbit
- Robotic mission to Mars to send back up a surface sample to take back home.
- A crewed mission to Phobos for collection of Surface samples, and Deep Surface Samples. 

Phobos crewed landing will be an optional choice if the crewed Mars landing contract is chosen, but will not allow double-dipping if Option2 path is taken. Deimos Crewed Landing contract will be an optional.

**Funding and Timeline:**
- Program funding on Moderate difficulty is $40M. Normal is $48, and Hard is $32M.
- Timelines: Normal: 18 Years, Fast: 12 years, Breakneck 9 years
- Pricing modeled around moderate difficulty with mission architecture placing 2,000T to LEO and the construction of a 5k-7kT LC with a second pad. Assuming a design using propulsive capture and chemical propulsion. 
- Assumption of 3,600 researchers and 5,000 engineers.
- Researchers account for ~$2.3M/yr
- Net cost is estimated to be ~$2.9M/yr with maxed out subsidy
- New LC will be at roughly half engineering capacity and roll out crafts in ~60 days for non-crew rated payloads.
- Front Loaded Funding Curve seems like a good start, but might be a bit heavy in the beginning and light at the second half. 

**To-Do:**
- Fix contract to perform certain crew experiments while in Mars Orbit. Possibly use of research payload from the Early Stations Program.
- Need to figure out how to write the contract for the EVA surface samples, if even possible. 
- Fix Mars Robotic sample return contract.
- Playtest after contracts are ironed out.
- Adjust Funding Curve.


siimav:
In case you haven't noticed yet - the cfg validator found some errors.

CXG-2827:
> In case you haven't noticed yet - the cfg validator found some errors.

Thanks. found the missing brackets.

CXG-2827:
Sharing my thoughts on the funding and timeline for this program.

Facilities and staffing in the post-moon landing era will need $3.5M/year just to stay afloat. Ballpark estimates for launching the Mars mission is based on 9x200T uncrewed payloads and 1x200T crewed payload/habitat totaling $3.5-4M accounting for engineering and rollout. Building a new LC and extra pad will be another $2.5M. I would think at this stage of a playthrough, the player should have 12 slots in the admin building to allow for 2 additional programs to be running concurrent with the Crewed Mars program. Assume $400k/year from the additional programs are contributing to the facilities and staffing budget, and $1.2M/year of maximum subsidy contributing to the researchers salaries.

To run the program on fast, you will need to begin construction on the new LC during the first year, as it will take ~3.75 years to complete. After the LC is complete, it will take another 2 years to construct and launch the 10 payloads, putting you at the end of the 6th program year. With 6 years remaining, it gives flexibility with missing a launch window and having enough time for the 2-2.5yr mission.
image

the current front-loaded curve will need for you to bank almost $6M by the 4th year, as you will be operating in the negatives for the remainder of the program.

image

Regarding the funding curve, if having such a large aggressive funding peak goes against simulating a slower post-Apollo era, then I believe Fast will need to be stretched to 20 years, budget changed to ~$50M, and use a Mid or flat funding curve. With Crewed Lunar being a 12 year program, maybe 14 years for Mars is a bit short.

CXG-2827:
Contracts have been tested in sims and a test save. I've checked the pricing against my current Mars design and increased funding a bit. Cost to launch the Mars mission estimated closer to $6.5M. From a spending estimate 12 years still works ok, but has a slightly shorted buffer for the mars window to complete and hit the program deadline. Build period for the infrastructure is now about 5 years on the Mid curve.
marsspend1

marsspend2

CXG-2827:
Files have been cleaned up with formatting.

byt3c:
@ cents on my end - this program is made with assumption that everything would need to be built from scratch. I found out that unless speedrunning heavily there is a lot of surplus money - too large funding removes incentive for optimization - thinking ahead of reusing pads/ architectures. I've used the same LV / diameter since crewed moon missions in all of my program. 

My suggestion would be to slightly underfunding the whole program would incentivise smarter management. Crewed mars program in late 70s is still remarkable achievement with having tons of surplus money and research.

CXG-2827:
> -The program needs more programs before it to "feel good" progression wise, going from 2 years in LEO with crew rotations to 3 years in interplanetary space is a bit of a jump. Lunar habs or long duration heo/lunar stations should do the job.
> 

I agree with this. Gating it behind an Early Lunar habitation program, or a 6-9 month contract from that program would be best.

> -The program is also balanced around building a new LC according to @CXG-2827 (correct me if im wrong) but it feels more natural to perform orbital assembly from the Moonboots LC since its already huge and doesnt perform any other task.
> 

RiS modified the radiation shielding density so it was a lot lighter, correct? Like you mentioned, the program funding was based around the existing kerbalism settings for the shielding mass. You might not have build a new LC for the Mars program, but that shielding change was likely a contributing factor.  

The idea with the funding was to give an opportunity for the player to build a UR-700/ SaturnV-25(S)U  LC. Assuming a Kerbalism shielding revamp is a North Star goal, it might be best to keep funding to allow these new LC for the meantime. a new 5-7kT LC plus pad is ~2.5M, so really not a huge portion of the program funding anyway.

ezsnackeur:
As i said the radiation system is bad so if your total travel time is shorter than 2y you dont need any shielding at all, if you take more than that then the shielding becomes the bigger fraction of your crafts mass with default settings, i think the setting that were used for RIS were too generous in foresight but they still dont address the fact that the system itself is flawed

byt3c:
"he idea with the funding was to give an opportunity for the player to build a UR-700/ SaturnV-25(S)U LC. "  it is there already with lunar program. If you play on fast with normal difficulty there is so much money that you can build enterprise and fly to the pluto and back with those money by 80s - obviously that is when you try to aggregate a lot of programs at the same time. We are not the average player. My rocket had 50t leo capability which was 1k1t pad. Taking into consderation most players would need 2kt pad it is still way enough for mars mission. 5-7kt is a speedrun pad basically as it is not needed in any other setup than large interplanetary missions built quickly or with absurd shield sized like Kargath 12m frontal one.

Obviously this is my opinion, so i'll be fine with whatever is decided at the end.

About radiation - i really miss the option to :
* make shielding lighter with tech, 
* Adding magnetic shielding at the later nodes
* Adding possibility to remove radiation from kerbal (like scrubber that has capacity to "heal" / process kerbal radiation out of it.) or have a cryostat in near future nodes that prevents it from impacting the kerbal or  make it slower. in exchange for some wake up stress penalty.

jimmymcgoochie:
In my experience in RIS2024, having to complete the full 2 years of multi-month habitation in the stations program before even starting crewed Mars was a bit unnecessary. One year would probably be sufficient to start the Program, with the contracts gated behind the full 2 years maybe?

I can't comment on funding as I fired all my researchers after starting the crewed Mars program and built my (slightly overkill) 6000t launch complex using the funding from stations, completing it before starting crewed Mars.

Not-a-flying-brick:
- **Targeted landing on mars is too hard to be compulsory**
I agree with byte's idea, can be made not needed but strongly recommended(even targeted landing on duna is hard, weak atmospheric landings are not suitable for spaceplanes nor actual rockets)
- **4 is a smart number**, means that all early capsules cannot be cavemanned with 1 voskhod ball goto everywhere. It forces realistic designs while keeping life support not a mess.
_Another new part idea unrelated until merged_ :make shielding IRSU(light and small, early in tech) so that ppl can mine from phobos or mars for lighter stage overall.

siimav:
With those new proposed durations, I feel like you're purely balancing this program for speedrunners. The durations are just barely longer than the expedited versions of MoonBoots/Stations.
Targeted landings are fine if you add a sufficiently large margin of error.
Btw, did everyone in the race opt for landing on the Potats instead of proper MarsBoots?

byt3c:
> With those new proposed durations, I feel like you're purely balancing this program for speedrunners. The durations are just barely longer than the expedited versions of MoonBoots/Stations.

Agree. It should accommodate failed mission and reply for all those who are not super experienced.

> Actual Mars landings were mandatory. No-one potato-landed to my knowledge.

We were targeting mars boots specifically. And also potato mission is much easier

ezsnackeur:
In the race all 5 finishers landed on Mars, it was mandatory, thats why im also saying that potato landing shouldn't be enough to complete the program as a ship for that is barely heavier than one for a flyby

siimav:
What's the expected budget difference (both in terms of $ and mass) for a mission that orbits Mars vs actually lands on it?

njits23:
Don't forget that both routes require some sort of ascent back from the surface of Mars. For the crewed potato landing you need to do a robotic sample return.

In the end I don't actually expect a very big difference in the total mission cost between the two routes. Depending on how  you do it landing crew on the surface of Mars has lots of ways to reduce cost, as you can just plunge straight into the atmosphere from the interplanetary transfer, rather than braking all the supplies into Mars orbit with an engine if you don't want to aerocapture. You do still have to keep a return vehicle somewhere, but that can be much lighter.

The robotic sample return is also pretty restrictive on where the big crewed ship should be for the sample return, a 200 to 600km orbit, so you actually need more vehicles, a robotic lander for the surface sample, a crewed expedition ship to go to a potato, and a return vehicle. You wouldn't even be able to mine any shielding efficiently (if my proposed shielding changes get merged), because you'd have to lug the supplies for a full window stay to a potato directly and then move the empty 4-crew ship to low Mars orbit to go pick up your samples to return, whilst a surface landing barely needs to spend dV to get its supplies for the full mission duration if it just plunges directly into the atmosphere.

byt3c:
well, in my case it takes 35tish less on LEO and 13t less on LMO. compared to 285t LEO and 83t LMO respectively.

CXG-2827:
I believe it is comparable. A round trip from low Mars orbit to Phobos is ~2.5km/s. And as mentioned, you also need to plan for a robotic mars sample return and rdvz for the potato path. It's been a while since I play-tested, but I remember it was not a trivial task.

What about bumping the current Phobos stay from 7 days to 30 so it matches the Mars duration? Though an issue I remember from playtesting, is the low gravity means the craft will jump when easing out of timewarp, and will cause the "landed" timer to reset. Might be less problematic if the Grabbing unit is used to latch to the roid.

njits23:
Speaking of the grabbing unit, the robotic sample return currently requires the presence of docking ports. Why? There's multiple other ways of solving the sample transfer. Restricting the player doesn't seem very nice. IIRC one of the loading screen images is even of a grabber being used to recover a Mars sample.

jimmymcgoochie:
Why does the sample return contract require crew at all? The current MSR architecture with Perseverance etc. is all probes. Is there a separate robotic sample return contract I'm just forgetting?

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -1688,35 +1688,56 @@ disabled_RP0_PROGRAM
...
-	repPenaltyPerYearLate = 2585
+	title = Crewed Mars Exploration (WIP)
+	description = After long-term testing of the mental and physiological impact of micro-gravity on the human body in low earth orbit, we can pursue the challenge of exploring our neighboring planet: Mars. This program will task you with the challenge of sending the first humans on Mars. The program allows for the choice to either land crew for a Martian surface stay, or to send a crewed expedition to Phobos and obtain a Mars surface sample via a robotic survey.  
+	requirementsPrettyText = Complete Crewed Lunar Program, and Collect Second Generation Station Science from the Early Station Program.
+	objectivesPrettyText = Send crew to orbit Mars and land crew on either Phobos or Mars.
+	nominalDurationYears = 12
+	baseFunding = 40000000
+	fundingCurve = Mid
+	repDeltaOnCompletePerYearEarly = 1500
+	repPenaltyPerYearLate = 1500
```
siimav, line 1703: Change these to 2000

PR #2423, opened 2024-08-15
Add support for Program Modifiers

siimav:
These will change the funding, duration etc values on a program if another gets accepted.

### DIFF
--- a/Source/RP0/Programs/Program.cs
+++ b/Source/RP0/Programs/Program.cs
@@ -74,7 +74,7 @@ namespace RP0.Programs
         [Persistent(isPersistant = false)]
         public double baseFunding;
 
-        [Persistent(isPersistant = false)]
+        [Persistent]
         public string fundingCurve;
 
         [Persistent]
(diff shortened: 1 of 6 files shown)


Source/RP0/Programs/Program.cs
```diff
@@ -569,5 +558,36 @@ public void SetBestAllowableSpeed()
...
+        /// This should NOT get called on template programs (i.e the ones in ProgramHandler.Programs) since there is currently no logic that resets them to original state.
+        /// </summary>
+        public void ApplyProgramModifiers()
+        {
+            var pms = ProgramHandler.ProgramModifiers.FindAll(pm => pm.tgtProgram == name);
+            foreach (ProgramModifier pm in pms)
+            {
+                if (ProgramHandler.Instance.IsProgramActiveOrCompleted(pm.srcProgram))
+                {
+                    pm.Apply(this);
```
Karhgath, line 573: What if you define two PM config with different source programs that modifies the same target program, you'd apply both and I am not sure the result would be deterministic here or be the intended effect?

siimav: PMs get applied in the order they are configured and thus the effects should be deterministic.

Karhgath:
Looking at the code, it might be possible to do an improvement where you could define a comma-separated list of srcProgram and have a AreActiveOrCompleted function instead. If would be considered a AND. It could enable scenarios like boosting both Early Sat programs if you have both Rocket Dev and Suborbital Research active or completed.

Source/RP0/Programs/Program.cs
Karhgath: So since "modifiers" just a new value configuration, the last one configured is the only one used at the end right? or could we do partial PM configs that could be combined? (One defines nee confidence, one defines new fundingcurve)

siimav: They would combine.

siimav:
> Looking at the code, it might be possible to do an improvement where you could define a comma-separated list of srcProgram and have a AreActiveOrCompleted function instead. If would be considered a AND. It could enable scenarios like boosting both Early Sat programs if you have both Rocket Dev and Suborbital Research active or completed.

Honestly that's not what I'm interested in right now. This should be tackled separately once we decide that such a feature is needed.

PR #2425, opened 2024-08-17
Make targeted moon landing easier

uriyuzhui:
Current `Crewed Targeted Moon Landing` requires landing within 1km of target locations, which is very difficult due to in-game limitations, also also does not make sense that the entire mission becomes a failure if you miss the site by just 1km.

The modified version：
- Relax distance requirement 1km -> 10km for completion (updated)
- Bonus science if you land within 3km, which is roughly [how far off Apollo did historically](https://space.stackexchange.com/questions/37488/how-far-off-did-apollo-11-land)
- Total reputation is unchanged (25% moved from base to the optional part)
- Check waypoint distance and landing status concurrently to prevent cheesing it by landing and then fly-by the waypoints

Also fixing some texts along the way.

### DIFF
--- a/GameData/RP-1/Contracts/Lunar Crewed/CrewedMoonLandingTargeted.cfg
+++ b/GameData/RP-1/Contracts/Lunar Crewed/CrewedMoonLandingTargeted.cfg
@@ -8,7 +8,7 @@ CONTRACT_TYPE
 
 	description = <b>Program: Crewed Lunar Exploration<br>Type: <color=red>CAPSTONE</color></b><br><br>Design and launch a spacecraft with at least one crew member to land at a specific location on the Moon. Explore the area for at least @/LandDur and then return safely to Earth.&br;&br;<b>Number of Contracts Completed: @index / @maxCompletions</b>
 
-	genericDescription = Launch a crewed single-person spacecraft and land it on a specific lunar biome. Explore the area for the specified amount of time and then return safely to Earth.
+	genericDescription = Launch a spacecraft with at least one crew member and land it on a specific lunar biome. Explore the area for the specified amount of time and then return safely to Earth.
 
 	synopsis = Land a crew on a specific biome on the Moon and return safely to Earth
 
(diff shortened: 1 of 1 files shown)


GameData/RP-1/Contracts/Lunar Crewed/CrewedMoonLandingTargeted.cfg
```diff
@@ -221,14 +221,15 @@ CONTRACT_TYPE
 			{
 				name = TargetedLanding
 				type = Any
+				completeInSequence = true
```
siimav, line 224: I don't get the addition of this. Doesn't make any sense to me.

GameData/RP-1/Contracts/Lunar Crewed/CrewedMoonLandingTargeted.cfg
```diff
@@ -306,6 +307,138 @@ CONTRACT_TYPE
 		}
 	}
 	
+	PARAMETER
+	{
+		name = OptVesselGroup
+		type = VesselParameterGroup
+		title = Demonstrate precision landing and land within 3km of target locations
```
siimav, line 314: Need to land within 1 target location, not multiple.

GameData/RP-1/Contracts/Lunar Crewed/CrewedMoonLandingTargeted.cfg
```diff
@@ -306,6 +307,138 @@ CONTRACT_TYPE
...
+		}
+
+		
+		PARAMETER
+		{
+			title = Land within 3km of target locations
+			name = TargetedLandingOptional
+			type = Any
+			hideChildren = true
+			completeInSequence = true
```
siimav, line 357: Or this

siimav (changes requested):

siimav:
On a separate note: all rover contracts need updating too to follow the same pattern as #2418

GameData/RP-1/Contracts/Lunar Crewed/CrewedMoonLandingTargeted.cfg
uriyuzhui: It's to seal a possible loophole that, you fly past the target area at low altitude, and then land at somewhere else

uriyuzhui:
- Fix the texts referring to plural "target locations"
- Lower the threshold 25km -> 10km for completion
- Move the VisitWayPoint blocks to be in the same ALL block as landing state checking, to seal the loophole of flying-past waypoints before/after landing

siimav:
There's something funky going on with the indentation now.

siimav (approved):

uriyuzhui:
image

Passed the test with an end-to-end out-of-simulation run.

PR #2426, opened 2024-08-30
Rebalancing Spaceplanes and Stations

Its-Just-Luci:
Makes space stations reliant on docking and materials science, rather than capsules. This allows non-capsule routes to doing stations, such as the X-20 Dynasoar. Ultimately, they are pressure hulls, and materials science and survivability serves as the backbone for pressure hulls regardless of their use-case, so I feel this is the most logical route. Additionally this doesn't make them significantly easier to unlock, just less tied to capsules.

Additionally, the Spaceplane line was changed, optionally divorcing it from capsules through a similarly challenging hypersonic research tree consisting of "Prototype Spaceplanes" and "Reusable Thermal Protection Systems", the former preparing for the BDB X-15A, the latter giving a place for TPS related items (namely spaceplane wings and shielded tanks) to reside. 

I feel the new tree both preserves spaceplanes as a more difficult route to orbit, while keeping them seperate from capsules so that players aren't required to do capsules first. That said, going from Mercury to Dynasoar is a faster way to orbit, but it takes more time to reach reuse, and will cost more. This leaves players with choices, which I feel is more fun and balanced compared to the previous requirements of needing to unlock Gemini prior to spaceplane research.


Its-Just-Luci:
In the process of redoing the Aeronautics node line as well, adding new airlaunch levels and enhancing spaceplanes. Some of these changes have been pushed to my fork.

PR #2428, opened 2024-08-31
Fix PolarSat Heavy to require FO

NovemberOrWhatever:
This puts it in line with first biosat and means early satellites heavy can't be completed in one launch anymore


siimav:
> Doing this probably stops heavy sats being anywhere near viable for speedrunning.

I don't care. Single launch programs shouldn't be a thing.

PR #2429, opened 2024-08-31
Remove downsides from subcontractors, buff research rate modifiers

NovemberOrWhatever:
Players will still have to pick carefully due to only having two subcontractor slots, but this removes the downsides from these less influential leaders, and should hopefully make the capitalist ones more practical. Feedback welcome

### DIFF
--- a/GameData/RP-1/Strategies/Leaders/LeadersSubcontractors.cfg
+++ b/GameData/RP-1/Strategies/Leaders/LeadersSubcontractors.cfg
@@ -71,7 +71,7 @@ STRATEGY
 		name = ResearchRateModifier
 		effectTitle = Crewed Spaceflight Visionary
 		effectDescription = command module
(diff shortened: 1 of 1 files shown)


GameData/RP-1/Strategies/Leaders/LeadersSubcontractors.cfg
```diff
@@ -221,37 +149,13 @@ STRATEGY
 		name = ResearchRateModifier //TODO: should be a integration speed buff to staged combustion
 									//TODO: leaving this like this now until I can target SC parts
 		effectTitle = Jet Engine Proficiency
-		multiplier = 1.1
+		multiplier = 1.05
```
siimav, line 152: Maybe instead of lowering the multiplier, we should bring all of them to +10%?

PR #2430, opened 2024-09-01
Edit leader balance

NovemberOrWhatever:
Using usage data from https://discord.com/channels/319857228905447436/1118681527199465503/1173498779316527155 and my own intuition, I rebalanced some leaders. Probably not perfect, but probably closer to perfect than it was. I also edited Qian's description while I was at it


NovemberOrWhatever:
For reference, I marked this as a draft as I feel like it should probably get a decent bit of feedback before its merged, not because it's necessarily unfinished

NovemberOrWhatever:
I can do it again better

Issue #2416, opened 2024-08-07
Earth-specific Capsule sciences work at other planets


CXG-2827:
It appears the capsule and WIP station experiments which have "celestialBodies = Earth" still work on Deimos and Phobos

![screenshot2](https://github.com/user-attachments/assets/b49cdb8d-7a43-45cf-bff0-e5ba1ab7b1fc)
![screenshot3](https://github.com/user-attachments/assets/fd129d7d-8b09-4a81-a12b-f0e8de1b9253)

Capkirk123:
This is probably a consequence of deleting all the Phobos/Deimos biomes, but also _incredibly_ niche.

Clayell:
> This is probably a consequence of deleting all the Phobos/Deimos biomes

I imagine this was done due to the minuscule size of them (?), but surely there could be a fake 'biome' that covers all of Phobos and a similar one for Deimos. This issue doesn't seem super niche to me, there are times where you would bring kerbals to mar's moons.

CXG-2827:
I noticed they also work on other planets, and not just limited to Deimos and Phobos. These experiments work in Mars and Venus orbit as well:
![screenshot7](https://github.com/user-attachments/assets/551ef5cb-23b7-48fb-ad48-164980753d45)
![screenshot8](https://github.com/user-attachments/assets/7d77cb91-5396-49da-b713-4983cc6642fd)

PR #2437, opened 2024-09-25
Enforce body restrictions for crewed experiments

TwistedGiraffe:
Solves #2416.
Added `BodyAllowed = HomeBody` for experiments with `celestialBodies = Earth`, and added `BodyAllowed = HomeBodyAndMoons` for experiments with `celestialBodies = Earth;Moon`. 
Also trimmed the whitespace.


TwistedGiraffe:
I discovered a very worrying mix of four-spaces and tabs among RP-1 configs that makes the whitespace appear not trimmed after all. Fixing this, however, is much beyond the scope of this PR.

Issue #2438, opened 2024-09-25
Bell Aircraft Company's hiring penalty is not applied


ezsnackeur:
![image](https://github.com/user-attachments/assets/8335731b-d029-4902-b060-a03872981481)

PR #2439, opened 2024-09-26
Fix Bell's researchers hiring penalty

ezsnackeur:
just copied Yvonne's cfg over as suggested, works correctly in game


siimav:
Fixes #2438

PR #2466, opened 2024-12-25
Add ablative airbrake

Clayell:
https://github.com/KSP-RO/RealismOverhaul/pull/3027
Still working on this so its just a draft


pap1723:
Any update on this?

Clayell:
Haven't looked at it in ages. Maybe I will someday (although maybe it would be better to add different configs for the existing airbrakes...)

PR #2473, opened 2025-01-05
Add additional orbital parameters to sim UI

siimav:
* Separate Ap and Pe (including negative values to easily yeet stuff into the atmosphere!)
* LAN


Clayell:
Might be useful to have the options for argument of periapsis (for molniya stuff) and true anomaly (not as important, but still helpful) too.

PR #2478, opened 2025-01-13
Add optional reward to docking contract for rendezvousing with a different craft first

Clayell:
~~As per discussion in https://discord.com/channels/319857228905447436/1295137142762242139/1328487821220118682~~
Additionally, add extra reputation/confidence for rendezvousing with a different craft before docking (instead of just re-docking to your upper stage, for example)

### DIFF
--- a/GameData/RP-1/Contracts/Earth Crewed Adv/Docking.cfg
+++ b/GameData/RP-1/Contracts/Earth Crewed Adv/Docking.cfg
@@ -80,6 +80,37 @@ CONTRACT_TYPE
 		tag = CrewedOrbitOptional
 		invertRequirement = true
 		title = Don't have an active optional crewed orbit contract.
+  	}
+
+	PARAMETER
+	{
+		name = Rendezvous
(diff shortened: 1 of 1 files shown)


Clayell:
As @uriyuzhui has pointed out, [Kosmos 186 and 188](https://en.wikipedia.org/wiki/Kosmos_186_and_Kosmos_188) docked autonomously in 1967 without any crew, so this PR doesn't make a ton of sense. I'll keep the optional reward for rendezvousing first though.

ezsnackeur:
Although uncrewed dockings have been performed the contract is the capstone of the advanced crewed program, it doesnt make sense that it can be done uncrewed can it?

ezsnackeur:
the same would be true for the rendezsvous then, thats not what the program is about

njits23:
Are you going to tell India it's doing it wrong? https://apnews.com/article/india-isro-space-docking-spadex-sriharikota-mission-b9c3d4c11468c1ea009b8c6ad45e05a7

Issue #2401, opened 2024-07-17
Landing on runway bonus confidence


uriyuzhui:
For optional x-plane contracts that award confidence:
- Mission Control shows no bonus confidence for landing on runway
- When completing the contract with landing on runway, pop-up windows shows bonus confidence is awarded
- However no extra confidence is actually awarded despite the display in pop-up window

uriyuzhui:
![20240622154715_1](https://github.com/user-attachments/assets/c92595fd-2a3c-4ec3-b3b9-bf0fe0ba847c)

For example, this one shows 47 bonus confidence is awarded, but it is not true

GraithTiger:
A similar issue is that when a game is loaded mid-flight after completing an altitude x-plane contract, the contract will not complete after landing.

PR #2479, opened 2025-01-13
Actually award reputation for landing on the runway correctly

Clayell:
As far as I know, "@/rewardReputation" is not a valid name.

Potentially fix https://github.com/KSP-RO/RP-1/issues/2401

### DIFF
--- a/GameData/RP-1/Contracts/Early Space Stations/Early Station Supply.cfg
+++ b/GameData/RP-1/Contracts/Early Space Stations/Early Station Supply.cfg
@@ -33,7 +33,7 @@ CONTRACT_TYPE
 	rewardScience = 0
 	rewardReputation = 200  
 	rewardFunds =  0
-	failureReputation = 0 // was @/rewardReputation
+	failureReputation = 0 // was @rewardReputation
 	failureFunds = 0
 
 	// ************ REQUIREMENTS ************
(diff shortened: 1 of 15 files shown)


Capkirk123:
Has this been tested?

Clayell:
No, but I can't think of any reason why it would be "@/rewardReputation" instead of "@rewardReputation", right?
I will try to test if I have some time.

njits23:
Landing _reputation_ gets paid out correctly. The problem in issue #2401 is that _confidence_ also shows up as a reward in the contract completion screen. This is unexpected, as confidence is not listed as a reward in mission control when accepting the contract. It only shows reputation. What issue #2401 actually requires to fix it is that confidence no longer shows up in the contract completion screen. Unless it is actually intended that landing on the runway also awards confidence, which this PR wouldn't fix anyways. The relevant CC documentation on the current configs being a valid syntax is over here: https://github.com/jrossignol/ContractConfigurator/wiki/Expressions#special-identifiers

Clayell:
> Unless it is actually intended that landing on the runway also awards confidence

I believe that the contracts are only meant to provide reputation, but these contracts were originally written long before confidence was added. In my opinion, they should give some extra confidence, as reputation only serves to mildly increase your subsidy.

Where is confidence awarded in contracts anyway? Is it just an automatic calculation based on the amount of reputation/science that the contract gives?

https://github.com/KSP-RO/RP-1/blob/f725e772d8e1758a0530be9cc6c04d97983c9bcd/GameData/RP-1/Contracts/X-Plane%20Research/RocketPlaneDevelopmentOptional.cfg#L8
> Optionally, landing back at the runway will award extra reputation. If playing from the Cape, the Space Shuttle runway provided by the RSS-CanaveralHD mod is also a valid landing target.

PR #2486, opened 2025-01-17
Add Lunar Habitation Program

CXG-2827:
@SierraHotel84 I noticed your upload was missing the original contracts and group modification, so those have been re-added.

Something for discussion is whether the 28+ day contracts should be locked behind the first station contract. From reading the following AAP document https://ntrs.nasa.gov/citations/19660022516 pages 53-54, it looks like long duration lunar orbits and lunar surface activities would be performed after long-duration earth orbit missions.  

From the same document (pg 56), I added an optional contract for a 28 day 3-crew lunar polar orbit. At the moment the objective is just the orbital stay, but I'll be testing to see how to add the return of lunar film samples.

I also added a 90 day lunar stay based on the following https://ntrs.nasa.gov/citations/19690008396 for an Extended Lunar Orbital Rendezvous (ELOR). It still needs some changes but would require rendezvousing back with the command module in lunar orbit as a contract parameter. For this one I think it would be a good idea to also require a certain amount of shielding resource on the landed shelter, or if there is a way to validate the crew return under a certain radiation %.


SierraHotel84:
Thanks, need to just blow away my old branch apparently, might explain the errors I kept getting.

Good call on the station contract gate for the longer missions.  I wonder then if we leave something like a 14-day stay as the longest required one, a la the original AAS program, and do a 30-day optional along with the 90-day one like ALSS/LESA.  Could use an experiment and higher confidence payout to make them enticing and worth doing.  Perhaps a crossover experiment from the stations program that could also be accomplished in this one?  

I saw the polar orbit mission as well but couldn't figure out how to work it in and have it be useful at the moment without an added experiment, so I definitely like the idea of lunar film samples, 

Good question on the radiation amount as a validation, I like it, but I've never looked into the parameters to see if kerbal characteristics can be used that way.  Would be a great check to figure out for any long duration mission. 

My big original intent was to also do a nuclear-based program, but I couldn't figure out a way to make it a standalone program on its own and it not just be a grind through a few launches, so I thought about having an optional mission here using a nuclear tug as a proof-of-concept since most of the Mars Design Reference docs I've seen all wanted a nuclear option.  Could be a gate for an over-arching nuclear engine program, but that's a different discussion.

SierraHotel84:
- [ ] Finalize required and optional missions
- [ ] Create science experiments for lunar film samples, surface exploration experiments, any additionals
- [ ] Determine funding curve
- [ ] Balance reputation/confidence payouts

CXG-2827:
Reviewing MMeridian's lunar habitation program, I copied over some of the contract texts and created a 180-day contract to instead be the capstone. With having a 90-day mission being the capstone, this might end up being too short of a program.

One thing I am thinking about is how the "orbit warper" acts if you were to send some crew out for some EVAs mid mission, as I believe it resets the timer when any of the crew depart from the craft. For the 30, 90, and 180d contracts, it might worth to add a crew parameter to allow the departure of some nauts during the lunar stay, but I wonder if there can be a loophole that would make this easily cheesed.

Tying into the comment above, I am looking into an optional MOLAB contract, which would have 2 nauts depart for the roving exploration. An Idea would be to travel X kilometers, remain at the location for 1 day (parameter to check velocity less than 1m/s), then return to the shelter. This would be a repeatable that would increase the travel distance and remote stay duration each time it is completed.

CXG-2827:
Closing this as progress has gone stale, and @WithCrow has submitted https://github.com/KSP-RO/RP-1/pull/2582

PR #2489, opened 2025-01-22
Set max tracking station level to 11

Capkirk123:
Set the max tracking station level in Skopos contracts to 11, so the contracts aren't accidentally hidden if you happen to start them very late.

### DIFF
--- a/GameData/RP-1/Contracts/Skopos Commercial Applications 0/FirstComSat-SCA.cfg
+++ b/GameData/RP-1/Contracts/Skopos Commercial Applications 0/FirstComSat-SCA.cfg
@@ -53,7 +53,7 @@ CONTRACT_TYPE
 			type = Facility
 			facility = TrackingStation
 			minLevel = 4
-			maxLevel = 10
+			maxLevel = 11
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


Clayell:
Just out of curiously, does maxLevel need to be set at all?

njd80:
As well as the 'intermittent TV' contracts, there are 5 'Firsts' in Commercial Applications 0 that also have MaxLevel = 10

https://github.com/KSP-RO/RP-1/blob/master/GameData/RP-1/Contracts/Skopos%20Commercial%20Applications%200/FirstComSat-SCA.cfg
https://github.com/KSP-RO/RP-1/blob/master/GameData/RP-1/Contracts/Skopos%20Commercial%20Applications%200/FirstGEOSat-SCA.cfg
https://github.com/KSP-RO/RP-1/blob/master/GameData/RP-1/Contracts/Skopos%20Commercial%20Applications%200/FirstGEOSat-SCA.cfg
https://github.com/KSP-RO/RP-1/blob/master/GameData/RP-1/Contracts/Skopos%20Commercial%20Applications%200/FirstMolniyaSat-SCA.cfg
https://github.com/KSP-RO/RP-1/blob/master/GameData/RP-1/Contracts/Skopos%20Commercial%20Applications%200/FirstTundraSat-SCA.cfg

Clayell (commented):
Can we omit specifying maxLevel completely, or will that break things? If a 12th level is ever added, we'll have to update this again.

Capkirk123:
> Can we omit specifying maxLevel completely, or will that break things? If a 12th level is ever added, we'll have to update this again.

I have no idea

PR #2492, opened 2025-02-02
Rename KDU-414 to S5.19

Clayell:
Fixes https://github.com/KSP-RO/RP-1/issues/2490.
Please let me know if I have configured anything here incorrectly.
Related: https://github.com/KSP-RO/RealismOverhaul/pull/3036


Capkirk123:
The "do not edit directly" label on the parts browser config files is there for a reason

PR #2494, opened 2025-02-03
Fix a timewarp kraken, add MNA and ArgPe to simulation GUI options, and some other various simulation GUI changes

Clayell:
Fix https://github.com/KSP-RO/RP-1/issues/2364.
Currently, only inputting Hours:minutes:seconds into the simulation window passes through [TryParseTimeString](https://github.com/KSP-RO/ROUtils/blob/master/Source/ROUtils/Utils/DTUtils.cs#L68) without any failures as expected, but this leads to some problem later down the line where the game starts timewarping rapidly (2364). This PR makes it so if the user is trying to specify a specific time, they must include the YYYY-MM-DD as well to avoid this error. (only putting in YYYY-MM-DD and not the HH:mm:ss is currently fine, as it just sets the time to midnight) All other formats should still work.

Also remind the user that they are allowed to specify seconds in that format, as it can be parsed perfectly fine.

RIP analemma

### DIFF
--- a/Source/RP0/UI/KCT/GUI_Simulation.cs
+++ b/Source/RP0/UI/KCT/GUI_Simulation.cs
@@ -10,7 +10,7 @@ namespace RP0
         private static Rect _simulationConfigPosition = new Rect((Screen.width / 2) - 150, (Screen.height / 4), 300, 1);
         private static Vector2 _bodyChooserScrollPos;
 
-        private static string _sOrbitAlt = "", _sOrbitPe = "", _sOrbitAp = "", _sOrbitInc = "", _sOrbitLAN = "", _UTString = "", _sDelay = "0";
+        private static string _sOrbitAlt = "", _sOrbitPe = "", _sOrbitAp = "", _sOrbitInc = "", _sOrbitLAN = "", _sOrbitMNA = "", _sOrbitArgPE = "", _UTString = "", _sDelay = "0";
         private static bool _fromCurrentUT = true;
         private static bool _circOrbit = true;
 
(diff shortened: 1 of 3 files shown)


Clayell:
Also also remind the user that they can just specify time in seconds.
Not actually sure if this works, or if I even put in the right amount of seconds for the example (I used 1960-12-31 23:59:59 and changed it to seconds).

Clayell:
Hm, on second thought, I think the format that is allowed should changed based on whether or not "From now" is selected.
If "From now" is selected, then it will only accept values in a format like "1y 2d 3h 4m 5s". (and the tooltip will change as well)
If "From now" is not selected, then it will only accept values in a format like "1960-12-31 23:59:59" or "61883248319". (and the tooltip will change as well) (still disallowing "23:59:59", but allowing "1960-12-31")

Clayell:
Also added 10 to the minimum safe altitude for bodies with atmosphere, as if you're a little too close you might accidentally immediately be in the atmosphere (especially with principia)
image

Clayell:
Also added mean anomaly at epoch and argument of periapsis to the simulation gui.

Clayell:
This PR should be ready to merge, testing is complete.
Sidenote: I tried to make this already-present tooltip smaller to make it fit within the GUIContent description,
image

but it ended up just shrinking the GUIContent description box for some reason.
image

This bug was already present before this PR, so if someone could figure out how to increase the size of the GUIContent description box that would be great.

Clayell:
> ...but this leads to some problem later down the line where the game starts timewarping rapidly (https://github.com/KSP-RO/RP-1/issues/2364)

At some point, it would be better to figure out the actual problem where inputting HH:mm:ss causes the game to freak out, so we can fix that bug and implement the more user-friendly solution where:

> putting in a time puts you at the next viable date (either today or the next day) for that time?

Source/RP0/UI/KCT/GUI_Simulation.cs
```diff
@@ -310,7 +344,7 @@ private static double EnsureSafeMaxAltitude(double altitudeMeters, CelestialBody
 
         private static double GetDefaultAltitudeForBody(CelestialBody body)
         {
-            return body.atmosphere ? body.atmosphereDepth + 20000 : 20000;
+            return body.atmosphere ? body.atmosphereDepth + 5000 : 20000;
```
Clayell, line 347: Strangely, principia seems to ignore the +5000 entirely, and just throws you at 140km if you don't specify an altitude. If you do specify an altitude, it will put you at the correct altitude.

PR #2501, opened 2025-02-06
Add KAC alarm for leader retirement and program deadline

Clayell:
Fix https://github.com/KSP-RO/RP-1/issues/2073
I honestly have no idea if this is set up remotely correctly, as I pretty much just copy and pasted from [here](https://github.com/KSP-RO/RP-1/blob/0a7d57a8038295a2be002089f5aa4b27fe48e456/Source/RP0/UI/TrainingGUI.cs#L425). If this is set up, this should automatically create a KAC alarm that says "Retirement: [leader name]", with the time set to their retirement date. I did not do this for normal astronauts because their retirement dates can move. (leaders' retirement dates can't move, right?)

### DIFF
--- a/Source/RP0/Leaders/StrategyRP0.cs
+++ b/Source/RP0/Leaders/StrategyRP0.cs
@@ -152,6 +152,10 @@ namespace RP0
                 CareerLog.Instance?.AddLeaderEvent(Config.Name, true, 0d);
                 KACWrapper.KAC.CreateAlarm(KACWrapper.KACAPI.AlarmTypeEnum.Crew, $"Retirement: {ConfigRP0.Title}", DateActivated + LongestDuration);
             }
+            else if (this is Programs.ProgramStrategy)
+            {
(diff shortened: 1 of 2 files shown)


Clayell:
image
Hm, that's not right...

Clayell:
image
Leader alarm worked, but program alarm did not.

Clayell:
I'll have to figure out a way to make an alarm for program retirement in another pr, several methods did not work for some reason.

Clayell:
(reposting here since I'm not sure if the ping went through and this is the relevant PR)
@siimav In your [recent commit](https://github.com/KSP-RO/RP-1/commit/3bb3c264369fa50f428ac5960deb9bb1d7a06f82), shouldn't the first line be removed? I believe this would create the alarm twice, and an exception would still be made.
https://github.com/KSP-RO/RP-1/blob/3bb3c264369fa50f428ac5960deb9bb1d7a06f82/Source/RP0/Leaders/StrategyRP0.cs#L153-L154

PR #2502, opened 2025-02-06
Add cost difference from previous vessel(s) while editing a vessel

Clayell:
If this actually works, it will show how much more the new edited vessel will cost, compared to the previous unedited vessel(s).
Fixes https://github.com/KSP-RO/RP-1/issues/2072
(this is kinda dumb tbh)


siimav:
What is this? Vessel merge feature is no longer used.

Clayell:
Oh yea, you're right, I totally forgot. I tried taking this from https://github.com/KSP-RO/RP-1/blob/0a7d57a8038295a2be002089f5aa4b27fe48e456/Source/RP0/Utilities/KCTUtilities.cs#L456, is that code redundant now? How would I get the cost of the vessel before edits? (assuming that editedVessel.effectiveCost is the cost of the vessel after edits)

PR #2504, opened 2025-02-07
Move CostExporter.csproj export path to \bin\Release\ to prevent unnecessary User folder being made

Clayell:
While CostExporter does not need to be built usually, there is no need for its .csproj to lead to a custom Users folder (I think its NK's?) if someone builds it by accident.
Let me know if the output path should be set to something other than \bin\Release\, I just set it to that to get it out of the way. (and so KSP wouldn't try to load its dll if it was built by mistake)


Clayell:
Should probably change this too. https://github.com/KSP-RO/RP-1/blob/37ec8ab9f5ea619770be6d0f501cd296321080d6/Source/Tech%20Tree/Parts%20Browser/CostExporter.runtimeconfig.dev.json#L4-L5

PR #2509, opened 2025-02-10
Early magnetometer rebalance

njits23:
I accidentally deleted the old PR for this change by updating my RP-1 fork's master branch. The original PR was here: https://github.com/KSP-RO/RP-1/pull/2449 and was apparently referencing the wrong branch.


njits23:
As far as I've been able to tell from looking at early probes, Mariner 2 (1962) and Mariner 4 (1964) weren't even really using boom-mounted magnetometers, but they were mounted away from the main bus.
Mariner 2, with the magnetometer below the omnidirectional antenna: 
Diagram_of_Mariner_1
Mariner 3/4, with the Magnetometer on the high-gain antenna waveguide:
800px-Mariner_3_or_4_diagram

Considering how much earlier MG01 shows up in the tech tree, I do not consider it appropriate to model it as anything other than what spacecraft of that era like Sputnik 3 and Explorer VI were using: bus-mounted magnetometers.

PR #2519, opened 2025-02-17
Fix various typos, add eras to kerbalism configs

Clayell:
Moved from https://github.com/KSP-RO/RP-1/pull/2483


Capkirk123:
Does this actually do anything?

Clayell:
It fixes typos, updates the Jupiter Observation description, and adds some eras to parts that were missing them.

PR #2520, opened 2025-02-17
Add "[Procedural]" to all ROSolar panels

Clayell:
Done in a similar vein to "Avionics [Procedural]".
Fix confusion in https://discord.com/channels/319857228905447436/331813459417235457/1336143787675750504
> To be fair, it doesn't actually say it's procedural...

Fix https://github.com/KSP-RO/ROSolar/issues/21

Should be done with https://github.com/KSP-RO/ROSolar/pull/22
Moved from https://github.com/KSP-RO/RP-1/pull/2483


Clayell:
Can you clarify what you mean?

Clayell:
@Capkirk123 I ran it through the Parts Browser, but all of the files are exactly the same compared to the master. (which should make sense, considering this is only a display name change?) This should be ready to merge?

PR #2528, opened 2025-03-03
Add suppress notifications to Near Future mods

Clayell:
According to https://discord.com/channels/319857228905447436/620690446540341261/1345817564428767343, there is a strange issue where selecting community tree with RP-1 (non-express I think) [causes ckan to pick an older version of RP-1 without the community tech tree conflict](https://github.com/KSP-CKAN/CKAN/issues/4325). This PR removes recommendations from near future mods to prevent community tech tree from being suggested, although this is not a solution to the problem.

### DIFF
--- a/RP-1.netkan
+++ b/RP-1.netkan
@@ -92,9 +92,13 @@ suggests:
   - name: HullcamVDSContinued
   - name: InternalRCS
   - name: NearFutureConstruction
+    suppress_recommendations: true
   - name: NearFutureElectrical
+    suppress_recommendations: true
   - name: NearFutureExploration
+    suppress_recommendations: true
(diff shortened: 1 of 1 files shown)


Clayell:
Getting RP-1 through the express install doesn't suppress the RP-1 recommendations though, so I'm not sure why this doesn't appear with express. I anticipate that CKAN has a hard cutoff to the levels of relations it can do? Express->RP-1->NF Mods->CTT is probably one too many.
https://github.com/KSP-RO/RP-1-ExpressInstall/blob/cc6216493340f5a3602f6d86195137823e3f111f/RP-1-ExpressInstall.netkan#L74-L77

PR #2537, opened 2025-03-10
Fix disconnect between avionics masses in procedural avionics window

Clayell:
Reported by Krydax (https://discord.com/channels/319857228905447436/845457276641738782/1348458469983916082)
The top value was getting mass from GetShieldedAvionicsMass, while the bottom value was getting mass from GetAvionicsMass.

image

Actually, what even is 'shielding mass'? It seems to have been added all at once, with no real explanation for what it was, and then never used again. It might even be throwing off GetModuleMass and RefreshDisplays. Would it be safe to [remove all traces of this "shielding mass"](https://github.com/Clayell/RP-1/pull/2)?
Commits where it was added:
https://github.com/KSP-RO/RP-1/commit/a107763f0abe51d1ada9de388c690e3ed1919b2c
https://github.com/KSP-RO/RP-1/commit/08df9c823d79fba93b862f9e6f7ca57ea74bbcbb

Only present in these 3 files:
[ModuleProceduralAvionics.cs](https://github.com/KSP-RO/RP-1/blob/master/Source/RP0/Avionics/ModuleProceduralAvionics.cs)
[ProceduralAvionicsTechNode.cs](https://github.com/KSP-RO/RP-1/blob/36d2b3e566038d2bbd23844d7790b339a181cb0c/Source/RP0/Avionics/ProceduralAvionicsTechNode.cs#L27)
[ProceduralAvionics.cfg](https://github.com/KSP-RO/RP-1/blob/master/GameData/RP-1/Tree/ProceduralAvionics.cfg)

### DIFF
--- a/GameData/RP-1/Tree/ProceduralAvionics.cfg
+++ b/GameData/RP-1/Tree/ProceduralAvionics.cfg
@@ -105,7 +105,7 @@ AVIONICSCONFIGS
 	AVIONICSCONFIG
 	{
 		name = Near-Earth
-		description = This part provides full control of the craft, instead of the limited control from Science core. The mass and volume of avionics necessary depend on tech level and the mass of the whole craft. It is functional only when close to Earth (up to 2x GEO distance). Uses: LV ascent guidance and upper stages that operate around the Earth.
+		description = This part provides full control of the craft, instead of the limited control from Science core. The mass and volume of avionics necessary depend on tech level and the mass of the whole craft. It is functional only when close to Earth (up to 2x GEO distance) due to a lack of shielding. Uses: LV ascent guidance and upper stages that operate around the Earth.
 
 		TECHLIMIT
 		{
@@ -447,7 +447,7 @@ AVIONICSCONFIGS
 	AVIONICSCONFIG
 	{
 		name = Deep-Space
-		description = This is the most advanced (and thus expensive) procedural avionics. Compared to to Near-Earth avionics, this works and offers control at any distance from Earth, and can be put on hibernation mode to save power. Uses: lunar and interplanetary probes, satellite buses when hibernation is desired.
+		description = This is the most advanced (and thus expensive) procedural avionics. Compared to to Near-Earth avionics, this works and offers control at any distance from Earth due to added TR (Thermal/Radiation) Shielding, and can be put on hibernation mode to save power. Uses: lunar and interplanetary probes, satellite buses when hibernation is desired.
 			
 		TECHLIMIT
 		{
--- a/GameData/RP-1/Tree/ProceduralAvionics.cfg
+++ b/GameData/RP-1/Tree/ProceduralAvionics.cfg
@@ -105,7 +105,7 @@ AVIONICSCONFIGS
 	AVIONICSCONFIG
 	{
 		name = Near-Earth
-		description = This part provides full control of the craft, instead of the limited control from Science core. The mass and volume of avionics necessary depend on tech level and the mass of the whole craft. It is functional only when close to Earth (up to 2x GEO distance) due to a lack of shielding. Uses: LV ascent guidance and upper stages that operate around the Earth.
+		description = This part provides full control of the craft, instead of the limited control from Science core. The mass and volume of avionics necessary depend on tech level and the mass of the whole craft. It is functional only when close to Earth (up to 2x GEO distance). Uses: LV ascent guidance and upper stages that operate around the Earth.
 
 		TECHLIMIT
 		{
@@ -447,7 +447,7 @@ AVIONICSCONFIGS
 	AVIONICSCONFIG
 	{
 		name = Deep-Space
-		description = This is the most advanced (and thus expensive) procedural avionics. Compared to to Near-Earth avionics, this works and offers control at any distance from Earth due to added TR (Thermal/Radiation) Shielding, and can be put on hibernation mode to save power. Uses: lunar and interplanetary probes, satellite buses when hibernation is desired.
+		description = This is the most advanced (and thus expensive) procedural avionics. Compared to to Near-Earth avionics, this works and offers control at any distance from Earth due to added complexity (Thermal/Radiation Shielding, Star Trackers, etc.), and can be put on hibernation mode to save power. Uses: lunar and interplanetary probes, satellite buses when hibernation is desired.
 			
 		TECHLIMIT
 		{
--- a/Source/RP0/Avionics/ModuleProceduralAvionics.cs
+++ b/Source/RP0/Avionics/ModuleProceduralAvionics.cs
@@ -130,7 +130,7 @@ namespace RP0.ProceduralAvionics
             return avionicsMass + GetShieldingMass(avionicsMass);
         }
 
-        private float GetShieldingMass(float avionicsMass) => Mathf.Pow(avionicsMass, 2f / 3) * CurrentProceduralAvionicsTechNode.shieldingMassFactor;
+        private float GetShieldingMass(float avionicsMass) => Mathf.Pow(avionicsMass, 2f / 3) * CurrentProceduralAvionicsTechNode.shieldingMassFactor; // use ^(2/3) to convert volume into surface
         private static float GetShieldingMass(ProceduralAvionicsTechNode techNode, float avionicsMass) => Mathf.Pow(avionicsMass, 2f / 3) * techNode.shieldingMassFactor;
 
         protected override float GetEnabledkW() => GetEnabledkW(CurrentProceduralAvionicsTechNode, GetInternalMassLimit());
--- a/Source/RP0/Avionics/ModuleProceduralAvionics.cs
+++ b/Source/RP0/Avionics/ModuleProceduralAvionics.cs
@@ -676,7 +676,7 @@ namespace RP0.ProceduralAvionics
             else if (controllableMass <= 0)
                 controllableMass = techNode.interplanetary ? 0.5f : 100f;
 
-            massKG = GetAvionicsMass(techNode, controllableMass) * 1000;
+            massKG = GetShieldedAvionicsMass(controllableMass) * 1000;
             cost = GetAvionicsCost(controllableMass, techNode);
             powerWatts = GetEnabledkW(techNode, controllableMass) * 1000;
             return controllableMass;
--- a/Source/RP0/Avionics/ModuleProceduralAvionics.cs
+++ b/Source/RP0/Avionics/ModuleProceduralAvionics.cs
@@ -591,11 +591,10 @@ namespace RP0.ProceduralAvionics
         public static string BuildPowerString(float enabledkW, float disabledkW)
         {
             var powerConsumptionBuilder = StringBuilderCache.Acquire();
-            powerConsumptionBuilder.Append("Online: ");
             powerConsumptionBuilder.Append(KERBALISM.Lib.HumanOrSIRate(enabledkW, KERBALISM.Lib.ECResID));
             if (disabledkW > 0)
             {
-                powerConsumptionBuilder.Append(" / Hibernated: ");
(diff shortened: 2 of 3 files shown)


siimav:
I don't understand your "fix". It appears to me that you made both numbers not agree with the *actual* avionics mass now.

Clayell:
Wow, I even made a mental note for myself that the top value was using GetAvionicsMass, while the bottom value was using GetShieldedAvionicsMass. I guess I switched them up halfway through.

On further inspection, this problem seems to be different than I thought. The disconnect only shows up for Deep Space probes not Near Space probes, so this seems to be a problem with GetStatsForTechNode or GetAvionicsMass. I will check it out in a bit.
image
image

The shielding mass seems to be a different issue entirely. The 'shielded' mass is used for displaying mass and the actual mass of the avionics (and therefore part), but I have yet to find a reason for why this 'shielding mass' was added. @siimav Do you know what the the shieldingMassFactor represents? It seems weird that there's a hidden mechanic adding a tiny bit of mass to every procedural avionics part, is this intentional? It's especially strange that the shieldingMassFactor does not change at all with increasing tech nodes, yet is re-specified for every single avionics tech node. (in [ProceduralAvionics.cfg](https://github.com/KSP-RO/RP-1/blob/36d2b3e566038d2bbd23844d7790b339a181cb0c/GameData/RP-1/Tree/ProceduralAvionics.cfg))

siimav:
I have zero reason to believe that shielding mass isn't intentionally added to the code.

Clayell:
Sorry, I misworded my question. Do you think that the shielding mass is worth keeping? I can't tell how it is making anything more  realistic or balanced. (and if it is worth keeping, shouldn't the mass factor decrease with increasing tech?)

siimav:
You don't just go and rip out everything you don't understand. If you think deep space avionics mass figures are wrong then come up with a better proposal and have some arguments to back that up.
As the shielding mass is based on avionics mass, you already get a mass reduction purely as a function of the other.

Clayell:
What? The deep space avionics top number is currently bugged, that's the main point of this PR. I haven't yet got the chance to look into it, but I imagine it should be pretty easy to fix.
The point about shielded mass is an entirely different issue that I discovered while looking into this that affects the mass of all types of avionics slightly but noticeably. I am just wondering if the 'shielding' is realistic or balanced, and what it would represent in the real world. It appears to me that no explanation, reasoning, or source was ever given for the mechanic being added, and so it almost looks unfinished to me. Sorry for asking you, I thought you would know what it was for.
Perhaps I can ask mike what their purpose was with the shielding mass? It was quite a while ago though, I wonder if they even remember.

Clayell:
> I imagine it should be pretty easy to fix.

I should not have jinxed myself, yet the answer seems so obvious now. It's multilayered, so I'll lay it out now:
- Only the Deep-Space AVIONICSCONFIG has a shieldingMassFactor setting, Near-Earth and Science-Core do not
- The top value uses GetAvionicsMass
- The bottom value uses GetShieldedAvionicsMass, which is the sum of GetAvionicsMass and GetShieldingMass
- GetShieldingMass is 0 when shieldingMassFactor is 0
Therefore:
- When we're using Near-Earth, GetShieldedAvionicsMass is the same as GetAvionicsMass, because shieldingMassFactor is 0. This means that the top and bottom values are the same.
- When we're using Deep-Space, GetShieldedAvionicsMass is not the same as GetAvionicsMass because shieldingMassFactor is not 0. This means that the top and bottom values differ.

Therefore, the simple solution is to modify the top value to use GetShieldedAvionicsMass instead. I am still unclear on the purpose of shielding mass, so I will probably ask mike and see what they say it is for.

Clayell:
Made shielding mass actually visible if it is defined.
image
image
Still running into https://github.com/KSP-RO/RP-1/issues/2516, which is sorta annoying.

MikeOnTea:
I've added some documentation and real world comparisons of the shielding mass to [the original issue](https://github.com/KSP-RO/RP-1/issues/968). If anything, the numbers are probably a bit on the low side, i think.

Source/RP0/Avionics/ModuleProceduralAvionics.cs
```diff
@@ -131,6 +131,7 @@ internal float GetShieldedAvionicsMass(float controllableMass)
         }
 
         private float GetShieldingMass(float avionicsMass) => Mathf.Pow(avionicsMass, 2f / 3) * CurrentProceduralAvionicsTechNode.shieldingMassFactor;
+        private static float GetShieldingMass(ProceduralAvionicsTechNode techNode, float avionicsMass) => Mathf.Pow(avionicsMass, 2f / 3) * techNode.shieldingMassFactor;
```
MikeOnTea, line 134: While it's not a big formula, i'd still prefer to not repeat it, one method could just call the other method.

Clayell:
Thanks for the commentary mike. I'll rename the shielding mass tooltip I added to "TR Shielding Mass". (TR = Thermal/Radiation according to https://ntrs.nasa.gov/api/citations/20070014633/downloads/20070014633.pdf) I'll also add a bit to the wiki that at least mentions this shielding.

Source/RP0/Avionics/ModuleProceduralAvionics.cs
Clayell: A similar thing is done with GetAvionicsMass a few lines up. This also allows for the tech node to be specified directly, instead of requiring the current tech node. (honestly, I tried and failed to get that technique to work before, so just gave up and did this, IIWIW)
https://github.com/KSP-RO/RP-1/blob/5d6515649a1f3a25998a36ec3bfda6747924243d/Source/RP0/Avionics/ModuleProceduralAvionics.cs#L120-L122

Clayell:
Made the tooltip a bit more clear and condensed. This PR is ready for merge.
image
image

siimav:
I'm sorry but I don't understand why you insist on explicitly mentioning the shielding mass. This has no relevance to the end user and is just blasting them with (IMO unnecessary) extra information.

Clayell:
There are a few reasons why it is fine to mention the shielding mass:

- It alerts a new player that Deep Space avionics cores are heavier than Near-Earth by telling them about a mechanic exclusive to Deep Space that will increase its mass considerably. It is more difficult to notice that Deep-Space is considerably heavier than Near-Earth if you are not aware that Deep-Space adds shielding, while Near-Earth does not.
  - I would also argue that showing a new mechanic to a new player will encourage them to look up what it means. More knowledge = more better?
- This tooltip is already full of unnecessary information. All of the info in it is repeated elsewhere within the Procedural Avionics window or the PAW. However, there is nothing else that tells the user about the distinction between the avionics mass and the shielding of the avionics mass, as the other results simply just combine them.
  - I would especially argue that the Axial control, can hibernate, and sample container fields should be removed from that tooltip considering that they are literally shown again right below it. 
- Time was expended to add the shielding mechanic, so I think it's fair that users get to appreciate the extra work (and realism) of procedural avionics cores.

siimav:
You have completely misunderstood the purpose of the tooltip. The idea is that you can just hover over the various TLs and quickly get an overview of what they bring to the table.
DS cores aren't heavier than NE because of the shielding. They already are even before the shielding mass is applied.

Clayell:
> DS cores aren't heavier than NE because of the shielding. They already are even before the shielding mass is applied.

Most of the additional mass is from the shielding, only some of it is from the increased mass exponents/constants/factors/etc.

> The idea is that you can just hover over the various TLs and quickly get an overview of what they bring to the table.

After thinking about this for a few days, I have decided to discontinue this idea of trying to show the shielding in the avionics window for now. I tried a couple of different approaches, but they didn't look too great:
image

I will add some notes to ProceduralAvionics.cfg to alert the user of this shielding though, since it is extra points for realism.

GameData/RP-1/Tree/ProceduralAvionics.cfg
```diff
@@ -105,7 +105,7 @@ AVIONICSCONFIGS
 	AVIONICSCONFIG
 	{
 		name = Near-Earth
-		description = This part provides full control of the craft, instead of the limited control from Science core. The mass and volume of avionics necessary depend on tech level and the mass of the whole craft. It is functional only when close to Earth (up to 2x GEO distance). Uses: LV ascent guidance and upper stages that operate around the Earth.
+		description = This part provides full control of the craft, instead of the limited control from Science core. The mass and volume of avionics necessary depend on tech level and the mass of the whole craft. It is functional only when close to Earth (up to 2x GEO distance) due to a lack of shielding. Uses: LV ascent guidance and upper stages that operate around the Earth.
```
siimav, line 108: Shielding is only part of the reason. There are others like star trackers etc so IMO shielding shouldn't singled out here.

GameData/RP-1/Tree/ProceduralAvionics.cfg
```diff
@@ -447,7 +447,7 @@ AVIONICSCONFIGS
 	AVIONICSCONFIG
 	{
 		name = Deep-Space
-		description = This is the most advanced (and thus expensive) procedural avionics. Compared to to Near-Earth avionics, this works and offers control at any distance from Earth, and can be put on hibernation mode to save power. Uses: lunar and interplanetary probes, satellite buses when hibernation is desired.
+		description = This is the most advanced (and thus expensive) procedural avionics. Compared to to Near-Earth avionics, this works and offers control at any distance from Earth due to added TR (Thermal/Radiation) Shielding, and can be put on hibernation mode to save power. Uses: lunar and interplanetary probes, satellite buses when hibernation is desired.
```
siimav, line 450: Same as above.

Source/RP0/Avionics/ModuleProceduralAvionics.cs
```diff
@@ -119,7 +119,7 @@ private float GetControllableMass(float avionicsMass)
 
         private float GetAvionicsMass() => GetAvionicsMass(GetInternalMassLimit());
         private float GetAvionicsMass(float controllableMass) => GetPolynomial(controllableMass, CurrentProceduralAvionicsTechNode.massExponent, CurrentProceduralAvionicsTechNode.massConstant, CurrentProceduralAvionicsTechNode.massFactor) / 1000f;
-        private static float GetAvionicsMass(ProceduralAvionicsTechNode techNode, float controllableMass) => GetPolynomial(controllableMass, techNode.massExponent, techNode.massConstant, techNode.massFactor) / 1000f;
+        internal static float GetAvionicsMass(ProceduralAvionicsTechNode techNode, float controllableMass) => GetPolynomial(controllableMass, techNode.massExponent, techNode.massConstant, techNode.massFactor) / 1000f;
```
siimav, line 122: Why change the access modifier if it isn't used outside of the same class?

Source/RP0/Avionics/ModuleProceduralAvionics.cs
```diff
@@ -676,7 +677,8 @@ internal static float GetStatsForTechNode(ProceduralAvionicsTechNode techNode, f
             else if (controllableMass <= 0)
                 controllableMass = techNode.interplanetary ? 0.5f : 100f;
 
-            massKG = GetAvionicsMass(techNode, controllableMass) * 1000;
+            var avionicsMass = GetAvionicsMass(techNode, controllableMass);
```
siimav, line 680: Should not omit type if it isn't apparent from the right side of the assignment.

siimav (changes requested):

GameData/RP-1/Tree/ProceduralAvionics.cfg
Clayell: I've updated this to not specifically single out shielding, but I'm still including it because I want there to be some mention of shielding *somewhere* in RP-1. Additionally, while not the only reason, shielding is the biggest factor as to why deep-space avionics are heavier.

PR #2546, opened 2025-03-26
Confidence and Reputation changes for some programs.

ezsnackeur:
All the changes were done using the provided sheet, which i updated, adding Skopos programs to it and separating disabled and deprecated programs from the working ones, as well as double checking all the working programs with the programs.cfg to make sure they match.

Overall the change increases consistency across the board for Confidence costs and fixes some values that were probably missed in the last big update, mostly regarding Reputation gain/loss for completing early/late.

What it doesn't do is any sort of targeted balance change, the sheet doesn't weight the type of curve in the Confidence calculations, in my opinion it should, waiting for feedback.

Heres the updated sheet:
[RP1 2.1 Program_Contract Associations.xlsx](https://github.com/user-attachments/files/19474230/RP1.2.1.Program_Contract.Associations.xlsx)

I also moved the deprecated programs to the bottom of the list in programs.cfg in a separate commit so it can be reverted if it wasnt a good idea.


ezsnackeur (commented):
this is to reserve the "disabled_RP0_PROGRAM" type for programs that are yet in developement. differentiating them from old deprecated ones

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -1145,8 +796,8 @@ RP0_PROGRAM
 	
 	CONFIDENCECOSTS
 	{
-		Normal = 1300
-		Fast = 2600
+		Normal = 1400
+		Fast = 2800
```
siimav, line 800: Leave confidence mult as 0.8 here

ezsnackeur: at .8 it would be 1100 2200, at .9 1250 2500, so previously it was at 0.93, do you want that or actual .8

PR #2548, opened 2025-03-27
Fix Yangel bonus, but for real

### LINKED ISSUE #2540: Yangel funds reduction bonus doesn't work at all
Image
Image
As seen in the screenshots after hiring Yangel the funding stays the exact same even tho i have 1 craft being integrated and 1 being recovered, the integration time is affected as expected.

ezsnackeur:
Yangel was supposed to grant -15% from rollout, recovery and integration costs but:
Recovery cost are not an actual thing apparently, recovering costs 0 excluding the staff salary.
It's not clear to me what Integration costs are, i would imagine its the EffectiveCost of the rocket, since you pay for it over time, but no transactionReason affects it. (it would need to affect the engineer's salary only while they are working)
To fix the problem i opted for reducing the purchase cost instead, also maintaining the reduction to rollout cost.
The reasoning is that EC is often more than 2x the purchase cost so instead of 15% less EC it will reduce purchase cost and rollout by 25%.

I also removed a few reduntant commas from some leaders text.

Fixes #2540

### DIFF
--- a/GameData/RP-1/Strategies/Leaders/LeadersEngineers.cfg
+++ b/GameData/RP-1/Strategies/Leaders/LeadersEngineers.cfg
@@ -152,7 +152,7 @@ STRATEGY
 		name = CurrencyModifier
 		effectTitle = Practiced Efficiency
 		currency = Funds
-		effectDescription = of vessel purchase and rollout
+		effectDescription = for vessel purchase and rollout
 		flipPositive = true
 		multiplier = 0.75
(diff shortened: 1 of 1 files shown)


Clayell:
> of vessel integration, rollout and recovery

Are you not a fan of the serial comma? It can be useful for removing ambiguity.

ezsnackeur:
Eeh theres not any ambiguity here, it means the same any way you read it

PR #2551, opened 2025-03-28
Add heatshield requirement for early uncrewed orbital and sub-orbital reentry contracts

ezsnackeur:
No more cheesing the reentry using a large fairing with high stringer mass

### DIFF
--- a/GameData/RP-1/Contracts/Earth Crewed/FirstOrbitRecovery.cfg
+++ b/GameData/RP-1/Contracts/Earth Crewed/FirstOrbitRecovery.cfg
@@ -96,6 +96,15 @@ CONTRACT_TYPE
 			title = Uncrewed
 			hideChildren = true
 		}
+    		PARAMETER
+		{
+			name = HasHeatshield
+			type = PartValidation
+			title = Include a heatsink or heatshield for reentry
+			part = ROH-AdjustableHS
+			part = ROC-MercuryHS			
+			hideChildren = true
+		}
 		PARAMETER
 		{
 			name = OrbitalVelocity
--- a/GameData/RP-1/Contracts/Earth Crewed/OrbitRecovery.cfg
+++ b/GameData/RP-1/Contracts/Earth Crewed/OrbitRecovery.cfg
@@ -115,6 +115,15 @@ CONTRACT_TYPE
 			title = Uncrewed
 			hideChildren = true
 		}
+  		PARAMETER
+		{
+			name = HasHeatshield
+			type = PartValidation
+			title = Include a heatsink or heatshield for reentry
+			part = ROH-AdjustableHS
+			part = ROC-MercuryHS			
+			hideChildren = true
+		}
 		PARAMETER
 		{
 			name = Orbit


ezsnackeur:
image

the part check doesnt work for the mercury heatshield but it works fine for the adjustable one, any pointers?
the mercury shield check could also be removed i guess.

ec429:
Would this also break using lifting RVs (i.e. unmanned spaceplanes)?  I feel like if you can make one of those work it should be allowed.

ezsnackeur:
It would break them but if a check for b9 spaceplane wings is implemented it opens the possibility to just add a minuscule one to pass it, defeating the point of the PR of forcing heatshield tech.
Tbf the other way around is true as well, an umanned spaceplane can have a small heatsink just to pass the check, so maybe the PR is fine as is.

Clayell:
Perhaps a more robust way to do this could be to detect if the part has the ablator resource. (although obv that wouldn't detect heat sinks, but you could keep the part check for procedural heat shields)

Capkirk123:
I agree that this is too restrictive, there's various probably valid ways to complete this program without heat shields

ezsnackeur:
Do you mean that its intended for the contract to be completed with a retro burn or that it should include spaceplanes in the requirements? if so then i guess i can just add the requirement for b9 wings as an alternative to a heatshield.
While still cheesable at least the contract would be hard locked between one of these 2 costly nodes

Clayell:
I should note that test_account specifically [pointed this problem](https://discord.com/channels/319857228905447436/1313442218870050818/1313442796824166432) out in the "Lessons learned" thread for RIS 2024.

> 2. Two first orbital return contracts in Crewed Orbit program can be completed before researching EDL and/or not using heatshields on the vessel. That is not right, not because it's physically impossible, but because the space program does these contracts specifically to test the new heatshield technology.  There should be a contract requirement to have a heatshield on the vessel.


I agree with his point that the program is meant specifically for these heatshields.

Capkirk123:
Fair enough I suppose. Does this work with Vostok?

ezsnackeur:
No? this is an uncrewed contract, its required that theres no crew on board, i dont think you should expect anyone to try to complete this with an empty Vostok, if anything it could have an alternative check for B9 spaceplane wings to allow completion with a drone spaceplane

Capkirk123:
> Lets ask ChatGPT and see what it says 

Let's not. Most of that is nonsense or just plain wrong.

I agree that this contract (and RP-1 as a whole) should allow player freedom in finding a solution. I'll look into seeing if decreasing fairing temperature limits is viable instead od mandating heat shields, although that does still leave very aggressive retro-braking I suppose.

NovemberOrWhatever:
I'll add that the point is to test the viability of heatshields for reentry of a crewed spacecraft (or, IRL, a nuclear warhead). Putting a heavy fairing around it is a perfectly viable option for returning a small payload and seems fairly realistic, but it's not exactly something that'd be practical for returning a crew capsule from orbit, and so doing that wouldn't actually be accomplishing the point of the contract. Also, yeah, an LLM-style noise generator is, as usual, not useful

ezsnackeur:
1.32t needs 700kg of heatsink, to be lifted with 56-57 tech, are you out lf your mind?

PR #2552, opened 2025-03-30
Make CostExporter.csproj use $(ProjectDir) to prevent unnecessary User folder being made

Clayell:
Related: https://github.com/KSP-RO/RP-1/pull/2504
This should hopefully solve the build issues that were present in https://github.com/KSP-RO/RP-1/pull/2536, since it'll actually be sending it to the place that it expects. I mean, hopefully, assuming that github actions doesn't freak out again. If it does, then this is probably a lost cause 🤷.

Worked on my version of Visual Studio 2022:
image

### DIFF
--- a/Source/CostExporter/CostExporter.csproj
+++ b/Source/CostExporter/CostExporter.csproj
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-    <OutputPath>C:\Users\nderaney\Documents\GitHub\RP-0\Source\Tech Tree\Parts Browser\</OutputPath>
+    <OutputPath>\bin\Release\</OutputPath>
     <PlatformTarget>x64</PlatformTarget>
   </PropertyGroup>
   
--- a/Source/CostExporter/CostExporter.csproj
+++ b/Source/CostExporter/CostExporter.csproj
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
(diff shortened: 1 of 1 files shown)


Clayell:
Additionally update:
https://github.com/KSP-RO/RP-1/blob/37ec8ab9f5ea619770be6d0f501cd296321080d6/Source/Tech%20Tree/Parts%20Browser/CostExporter.runtimeconfig.dev.json#L4-L5

Not 100% sure that using %USERPROFILE% will work here, since I don't really have a way of triggering this file to run, but I believe it should work? %USERPROFILE% would give something like "C:\Users\nderaney", but the extra backslashes should just be to escape any special characters, so I think it'll be fine. This file wasn't causing any problems anyway though.

Source/CostExporter/CostExporter.csproj
```diff
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-    <OutputPath>C:\Users\nderaney\Documents\GitHub\RP-0\Source\Tech Tree\Parts Browser\</OutputPath>
+    <OutputPath>$(ProjectDir)Source\Tech Tree\Parts Browser\</OutputPath>
```
siimav, line 9: This should just be `..\Tech Tree\Parts Browser\`

Clayell: image
Hm, yea that seems to work too. Is there a particular reason for using one over the other?

siimav: Simple is better?

Clayell: Sure, I guess. I find this to be a bit more descriptive, but I've seen the ".." used in other csproj files around here, so I'll change it to that to be consistent.

PR #2555, opened 2025-03-31
Allow for commas in autohire GUI, allow for both errors to show up simultaneously

Clayell:
Fix https://github.com/KSP-RO/RP-1/issues/2547.

Strips any commas from the strings before using them. (ReserveFunds doesn't even need to be stripped since it seems that double.TryParse can handle commas anyway, but there's no harm in doing it)
Additionally, previously only the "reserve funds" error was able to show up, even if the error was within parsing "crew count". This is because the "staff count" error would only show if the staff count had successfully parsed, which doesn't make any sense. I have changed this logic to allow for both error messages to show up at the same time if necessary.

image
image
image


ezsnackeur:
the correct word is not crew but rather staff

PR #2606, opened 2025-08-07
Add methods related to alarms with and without KAC

Clayell:
Fix https://github.com/KSP-RO/RP-1/issues/2517 (both parts)

CreateAlarm adds a way to make an alarm with a description that defaults to KAC if installed, but otherwise creates a stock alarm. I've made this trigger for:
- Program deadlines
- Leader retirement (although I don't think we use it anymore since https://github.com/KSP-RO/RP-1/pull/2569)
- Leader penalty fee expiration
- Leader reactivation cooldown
- Leader deactivation cooldown

ClearAlarms adds a way to remove all alarms (again accounting for KAC) with a certain string in the title. (to prevent duplicate alarms confusing people)
I've made this trigger when:
- Activating a leader
- Deactivating a leader
- Completing a program (no need to do it when activating the program, since the program can't have been activated before?)

Tested in-game to be working.

I'm not sure if "[program] must be completed at this time to avoid penalties." is the best choice of words, but I couldn't think of anything better.

### DIFF
--- a/Source/RP0/Harmony/Administration.cs
+++ b/Source/RP0/Harmony/Administration.cs
@@ -378,6 +378,11 @@ namespace RP0.Harmony
             if (!Administration.Instance.SelectedWrapper.strategy.Deactivate())
                 return;
 
+            if (Administration.Instance.SelectedWrapper.strategy is ProgramStrategy ps)
+            {
+                ps.ClearAlarms(ps.Title);
+            }
+
             Administration.Instance.UnselectStrategy();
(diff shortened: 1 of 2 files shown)


Clayell:
Screenshots:
KAC:
image
image

Adding leader:
image
image
image

Removing leader:
image
image
image

Completing program:
image
image
image

Stock: (I hate this UI, KAC is way better)
image

Removing leader:
image
image
image

Completing program:
image
image
image

PR #2608, opened 2025-08-07
Stop flickering of main KCT window when Build Plans Window is open

paulo-santana:
When the build plan window is open, the main window from KCT gets it's size set to 1 and back constantly whenever the mouse is button is being pressed.

I believe this line of code is the a remnant of some refactors done some years ago and it never got touched again. I did some local testing in the space center and in the VAB and didn't notice any loss of functionality, hence why i believe it's just left over code. More testing or some explanation would be appreciated though.


Clayell:
Can confirm this works.

PR #2609, opened 2025-08-09
Add separate S5.4 part

njits23:
Adds a separate S5.4 part to 1959OR (Vostok, Voskhod, and Zenit deorbit engine).

Needs https://github.com/Capkirk123/ROEngines-Extended/pull/5


njits23:
Also discounts the VostokSM after buying the S5.4 engine.

PR #2612, opened 2025-08-13
Fix Science Config Indentation

WithCrow:
Most of the science configs are not consistent with their indentation  - this PR aims to fix that.


siimav:
I'd suggest enabling this option and going over all your changes again. There's more to it than just unifying whitespaces.
image

PR #2616, opened 2025-08-17
Revert "Add methods related to alarms with and without KAC"

Capkirk123:
Reverts KSP-RO/RP-1#2606
@Clayell it doesn't compile and I'm not debugging it


Clayell:
@Capkirk123 What was the compile error? Seemed to work fine on my end. Perhaps something with adding more "using" lines?

PR #2619, opened 2025-08-17
Add Kerbalism hard drive configs

Capkirk123:
Very WIP. Actually add hard drive configs for kerbalism based on real probe data storage systems. Most of the work is actually already done for this, it just needs sensible values to be set for hard drive capacity, mass and cost.


Capkirk123:
Well it functions and my eyes glaze over whenever I try to do anything else so I'm declaring this probably ready

Clayell:
Someone should update https://github.com/KSP-RO/ROKerbalism/wiki/Science#storage.

PR #2626, opened 2025-08-23
Increase the technology and contract requirements for the crewed lunar missions

njits23:
Currently, the crewed lunar exploration program has no limitations on what sort of capsules are acceptable for use in the cislunar environment. This leads to strange situations where a Vostok lunar landing is an entirely workable concept.

This PR adds a new experiment, Lunar Landing Television Broadcast, which unlocks once the lunar landing tech node is unlocked and adds it to lunar orbiter, mature and later capsules, and lunar landing capsules. This experiment is required to be completed to complete the first crewed moon landing of either type.

The preexisting experiment for TV broadcast on the way to the moon is now required to complete the crewed lunar orbit contract.

To achieve this assignment of experiments, the way crew science was assigned to capsules had to be reworked as well. I've also added the 2nd gen experiments to mature capsules to enhance player choice.

Hopefully we can figure out some way to require near-live transmission of the TV broadcasts by requiring a sufficient bandwidth during it. Perhaps via what Skopos implements. I don't know how to achieve that, though. For now it just requires all the data to be transmitted before recovery.


njits23:
The Kerbalism version required to make the live TV experiments work as intended has released. https://github.com/Kerbalism/Kerbalism/releases/tag/3.23
The PR should now be ready for review.

Capkirk123:
I don't love adding a whole new antenna band just for a contract

siimav:
> I don't love adding a whole new antenna band just for a contract

This is somewhat problematic for sure. I've seen people try to configure the Skopos bands on their antennas for regular comms.

njits23:
Well, that part won't be a problem here. Analog Video S-band band just acts like regular S-band with a much bigger bandwidth but less bandwidth at range due to not having an encoder. Maybe the name needs to be Wide S-band instead, to better reflect what it actually is.

Capkirk123:
Ok did some testing, saturating a wideband S-band channel to a DSN station using the Apollo HGA at the moon with TL5 antennas and ground stations requires 0.05 Watts, while with default S-band it requires 0.01 Watts. Turns out the DSN stations are pretty chunky, you don't need much of an antenna at all to talk to them.

In light of this information, this is not worth adding a new antenna band at all. The data rate requirement is functionally just an S-band check, any directional antenna will trivially saturate the band if the DSN stations are listening, so just use normal S-band.

It's mildly ahistorical but after Skopos, I really don't want to add more bands, especially as a "gotcha" for completing a lunar landing contract. Just force the player to saturate S-band (150 kbps for encoder margin).

njits23:
> Ok did some testing, saturating a wideband S-band channel to a DSN station using the Apollo HGA at the moon with TL5 antennas and ground stations requires 0.05 Watts, while with default S-band it requires 0.01 Watts. Turns out the DSN stations are pretty chunky, you don't need much of an antenna at all to talk to them.

Correct, but the main thing here is that you only tested with a dish antenna. Reducing the bandwidth requirements suddenly makes omnidirectional S-bands also viable, which I find undesirable.

Here is what you need to hit 500 Kbps with a 3 dBi omni on the wideband:
image
Whilst 165Kbps is only a little over half the watts (and also 5kg lighter):
image

Mass-wise and cost-wise a 500 Kbps 19.2 dBi 24 dBm dish and a 165 Kbps 3.0 dBi 33 dBm omni are basically the same, though, so I suppose it doesn't really matter. And yeah, a 165 Kbps dish and a 500 Kbps dish aren't much different either. I still don't see the problem with an extra band, but I'll change it.

PR #2632, opened 2025-08-29
Remove landing training from Apollo capsule and split Gemini Lunar training

Karhgath:
Removed landing from the Apollo capsule training requirement (unlocking training at AC4 instead of AC5) while keeping the LEM as AC5. Since you still need LEM training if included in the LV, no training time is lost.

This would allow early Apollo 8 missions with Apollo or even use Apollo for Adv Crew if you are coming late to the Crew Program and skip Gemini altogether, without needing AC5 yet, which is focused on landing operations.

Also split off Gemini Lunar into extra lunar operation's training in line with the proposed changes to crew lunar experiment being a requirement in lunar explo, allowing Gemini Lunar to be usable at the cost of extra training. Also lumped in Big Gemini as potential lunar ferry as it is a later variant.

There is weird training progression where Gemini Lunar is more training than D2 and more training than Apollo, but that can account for more advanced tech making it easier to train.

### DIFF
--- a/GameData/RP-1/CrewTrainingTimes.cfg
+++ b/GameData/RP-1/CrewTrainingTimes.cfg
@@ -134,7 +134,7 @@ TRAININGTIMES
 		Mk2Pod = Gemini
 	Gemini-Mission = 120
 
-	// Advanced Gemini (Lunar), also include Big Gemini as potential lunar ferry
+	// Advanced Gemini (Lunar)
 	GeminiLunar = 50, Gemini									// 700
 		ROC-GeminiLCMBDB = GeminiLunar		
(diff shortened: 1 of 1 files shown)


njits23:
I do not think the changes as proposes here make sense. Right now Gemini L takes more time than D2, and Apollo takes less time than D2 and Gemini L. D2 and Gemini L being the same would make sense, and considering as lunar orbiter capsules their purpose is the same as Apollo, making them all 700 days seems the most sensible.

Karhgath:
@siimav I removed Big Gemini.

As for Lunar Gemini, is your point that we should not have dedicated training for it and keep it as normal Gemini?

PR #2636, opened 2025-09-02
Reduce TL0 FlyingLow and FlyingHigh Science

CXG-2827:
Temperature, Telemetry, and Pressure Scan are a bit powerful, and may discourage using later tech experiments for gathering science. Given that these only require 5-10 minutes of operation, it doesn't seem appropriate for having these provide science for biome-specific areas under FlyingLow and FlyingHigh conditions. This PR is to change these to only provide science for the situation rather than each biome.

To provide some data, the below is what can be gathered with Telemetry, Temp, and Pressure if you look at Venus and Mars. This excludes space high and low.

Venus (8 biomes): 2,033 Science
Landed science: 704
FlyingLow/High: 1,329

Mars (19 Biomes): 3,406 Science
Landed science: 1,463
FlyingLow/High: 1,943

Total for Venus and Mars: 5,439

This alone could get you to 1972 (6,069 total science points), accounting for all the tech already researched to take on Early Inner Planet Probes. With the proposed change, the above scenario would be reduced to 2,519.

From a gameplay perspective, this will not impact the science gathered with the first probe sent to the bodies, but subsequent missions would have diminished returns. This change will also impact the sounding rocket era by removing FlyingHigh temperature-farming from biomes adjacent to the LC. To compensate, the multiplier was increased.

RP-1 Tech Tree - Total Science

### DIFF
--- a/GameData/RP-1/Science/BodyScienceParams.cfg
+++ b/GameData/RP-1/Science/BodyScienceParams.cfg
@@ -13,7 +13,7 @@
 				%landedDataValue = 1
 				%splashedDataValue = 1
 				%flyingLowDataValue = 1
-				%flyingHighDataValue = 0.6
+				%flyingHighDataValue = 1
 				%inSpaceLowDataValue = 1
 				%inSpaceHighDataValue = 1.25
 				%recoveryValue = 1.0
--- a/GameData/RP-1/Science/Experiments/PressureScan.cfg
+++ b/GameData/RP-1/Science/Experiments/PressureScan.cfg
@@ -11,7 +11,7 @@
 	@dataScale /= #$baseValue$
 	@requireAtmosphere = False
 	@situationMask = 61
-	@biomeMask = 5
+	@biomeMask = 1
 	
 	%RESULTS
 	{
--- a/GameData/RP-1/Science/Experiments/TelemetryAnalysis.cfg
+++ b/GameData/RP-1/Science/Experiments/TelemetryAnalysis.cfg
@@ -12,7 +12,7 @@ EXPERIMENT_DEFINITION
 	dataScale = 0.00005625
 	requireAtmosphere = False
 	situationMask = 63
-	biomeMask = 7	
+	biomeMask = 3	
 	
 	RESULTS
 	{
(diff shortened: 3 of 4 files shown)


ezsnackeur:
This is a bad way to fix this imo, the actual way would be rebalancing rhe science body multipliers which i plan on trying to do xd

njits23:
I also disagree with this change, but more on the basis of how these kinds of experiments were conducted IRL. The Venus balloons from the Vega program launched by the Soviet Unions specifically investigated atmospheric properties (including temperature and pressure) across what in KSP would be considered biomes.

Changing the body science multipliers doesn't make sense either, because then you would have to start doing multiprobe missions to match current science income. Personally, I just don't see the problem with Venus and Mars being able to supply so much science. 1972 isn't all that far, after all, and its a lot of work to get all biomes even with simple landers, so a suitable reward is to be expected. The main reward isn't even the ability to select nodes in R&D, but the research speed boost you get from gathering science.

ezsnackeur:
The way to address that would be calculate the multipliers for the planet first, then divide each situation by the amount of biomes it can be collected in, this way more biomes becomes a negative attribute

siimav:
> This makes sense for imaging type experiments

No, it doesn't. Large amount of tiny biomes doesn't play nice with Kerbalism at high timewarp rates. The gameplay value of biomes is giving the player some motivation to go for higher inclination orbits and rewarding them for landing at various locations on the planet. We definitely do not need 10+ biomes for that.

CXG-2827:
Understood. I was not justifying the 19 biomes for mars, simply commenting on imaging experiments having biome-specifics

Capkirk123:
Converted to draft since this is not in a state where we want to merge

pap1723:
A better way to have this work might be to do something similar to how DMagic originally had the situations setup for the Spy Cameras.

Essentially you had northern hemisphere and southern hemisphere.

Maybe expanding temperatures to be different altitudes would make more sense. I am going to get all of the names wrong (and don't care enough to look right now), but having troposphere, stratosphere, mesosphere, etc would make the most sense for barometer and temperature readings.

For something like Venus, this could be important. Because of the greenhouse effect, there is very little temperature variance, but that wasn't known before it was tested.

PR #2646, opened 2025-09-28
Do not complete proficiency trainings more than once

Clayell:
Fix https://github.com/KSP-RO/RP-1/issues/2571

Still not sure what causes the problem to only show up in the tracking station.

[FixedUpdate](https://github.com/KSP-RO/RP-1/blob/master/Source/RP0/Maintenance/MaintenanceHandler.cs#L468) -> [Process](https://github.com/KSP-RO/RP-1/blob/master/Source/RP0/Crew/CrewHandler.cs#L172-L183) -> [ProcessCourses](https://github.com/KSP-RO/RP-1/blob/master/Source/RP0/Crew/CrewHandler.cs#L804-L822) -> [IncrementProgress](https://github.com/KSP-RO/RP-1/blob/master/Source/RP0/Crew/TrainingCourse.cs#L310-L322) -> [CompleteCourse](https://github.com/KSP-RO/RP-1/blob/master/Source/RP0/Crew/TrainingCourse.cs#L158-L234)

Due to CompleteCourse being called every 1 second in FixedUpdate, this bug should have been present everywhere, not just the tracking station? Maybe there's some weirdness with FixedUpdate.


Clayell:
I wasn't able to decipher what's causing the other errors to show up.
image

Clayell:
Managed to get rid of the NRE at least.
image

Clayell:
> Did you not check whether fixing the nullref already fixes the issue? Looks like it probably does.

Hm, it does seem to fix the issue? I can't possibly imagine why though. The DontDestroyOnLoad error still appears though.

Is it worth it to keep the extra check to make sure we haven't completed proficiency training before? It's only running upon course completion so I don't see much of a performance impact.

siimav:
Couple of levels up: https://github.com/KSP-RO/RP-1/blob/3c41cf2549ee3c13fa067c42ba2351e44c85383f/Source/RP0/Crew/CrewHandler.cs#L813

No point in having code that isn't supposed to get triggered.

Clayell:
> Couple of levels up:
> 
> https://github.com/KSP-RO/RP-1/blob/3c41cf2549ee3c13fa067c42ba2351e44c85383f/Source/RP0/Crew/CrewHandler.cs#L813
> 
> No point in having code that isn't supposed to get triggered.

Can you specify what you're pointing out here? I'm not seeing how this is related.

siimav:
> Is it worth it to keep the extra check to make sure we haven't completed proficiency training before?

Was a reply to this.

Clayell:
> No point in having code that isn't supposed to get triggered.

Well, I mean clearly it was getting triggered in some way, an astronaut could have like 20 proficiency trainings of the same target. I have no idea how adding a null check to a seemingly unrelated popup fixed this, and I imagine the bug could still pop up under different circumstances.

Clayell:
Hm, hadn't thought of it that way.


Wouldn't this imply some deeper problem with popup dialogs in the tracking station?

PR #2665, opened 2025-10-24
Make error messages for install checker more specific in what they are missing

Clayell:
Suggested by [njit](https://discord.com/channels/319857228905447436/1115321536807718992/1431293350606278748)


Clayell:
I'll assume you're referencing this:

https://github.com/KSP-RO/RP-1/blob/c22876b88d142cd3e1bfdf0345d10fd5e25fdc04/Source/InstallChecker/InstallChecker.cs#L30

So, if a `KSPAssemblyDependency` isn't satisfied, then `AssemblyLoader.loadedAssemblies` won't contain that dll? That makes sense I suppose. I can just change "missing" to be "not loaded", since that would be more accurate.

PR #2676, opened 2025-11-21
Give Sputnik an empty hard drive so it actually works

Capkirk123:
resolves https://github.com/KSP-RO/RP-1/issues/2675

Apparently if you delete a hard drive from a part completely it bugs out and can't transmit science. Give Sputnik an empty hard drive instead so it works properly-ish.


Capkirk123:
I mean if your best probe core is Sputnik, none of your probe cores have hard drives so what does it matter?

PR #2679, opened 2025-12-04
Give starting parts 1 Byte of data, not zero

### LINKED ISSUE #2678: v4.1.2.0 New hard drive mechanics potentially functioning incorrectly
Happy to post log files and screenshots, although it may take awhile, my current save has been modified since initial discovery of the issue. I did verify that the issue remained consistent between my modded instance and my initial, fresh, express install.

When setting difficulty upon original setup, unchecking the "Automatically flag files for transmission" box in Kerbalism settings prevents any data transmission. Leaving the setting checked, as it is at default, allows data transmission but displays hard drive value of "%NaN%" similar to the issue in #2652. I'm unsure if this is intended behavior or a bug.

On the one hand, the way the .cfg is written it seems like the hard drive slot is meant to be removed until the earliest avionics node that unlocks an upgrade. On the other, if it is intended behavior, you can't collect any early science until the early tracking node with the bio sample because the temperature and pressure scans both don't have a sample value. I'm not very familiar with .cfg creation or the intricacy of KSP modding, so forgive me if this comes from a place of ignorance, but would it be possible to change the base data value for memory from 0 to an arbitrary nonzero number? I'm assuming based off what I know that the base value for ProcAvionics gets multiplied by the upgrade modifier when you unlock the next tier of HDD. It would seem to me that since this value is set to 0 that the upgrades are multiplying their values by 0 and returning the %NaN% result. Again, my knowledge is cursory so apologies if I'm making a simple mistake in understanding how this is supposed to function.

I would also like to add, if it is intended behavior and temperature and pressure scans data transmission is supposed to be locked behind the first HDD upgrade, I think some documentation should be added to the wiki reflecting this particular aspect.

Capkirk123:
Apparently having 0 data storage capacity on a craft causes the Kerbalism data storage UI to get a bit buggy (displaying NaN for data capacity) so set everything to 1 Byte instead. Still too little to be useful, but makes the UI display data correctly.

resolves https://github.com/KSP-RO/RP-1/issues/2678

### DIFF
--- a/GameData/RP-1/Science/HardDriveConfigs.cfg
+++ b/GameData/RP-1/Science/HardDriveConfigs.cfg
@@ -717,7 +717,7 @@ KERBALISM_HDD_SIZES
 {
 	@MODULE[HardDrive]
 	{
-		%dataCapacity = 0				// will be upgraded with partupgrades later
+		%dataCapacity = 0.000001		// will be upgraded with partupgrades later. 1 byte (UI gets buggy with 0 data capacity)
 		%sampleCapacity = 1
 		%maxDataCapacityFactor = 10		// Sure, just add more tape
 		%dataCapacityCost = 50			// You can't un-upgrade, so this has to be universal. Be conservative and set it fairly low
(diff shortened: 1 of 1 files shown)


Capkirk123:
Nevermind, closing this since https://github.com/Kerbalism/Kerbalism/pull/971 fixes it at the source

Issue #2414, opened 2024-08-01
Space station parts too cheap


siimav:
As an example, a fully kitted early space station costs about 150k all-in-all to launch into LEO. If I used the same launcher but replaced the station with a D2 then it would cost about 200k. The gap would be even more wild with something like Apollo CSM.

CXG-2827:
The x1.5 Reentry-rating of descent modules seems to be a noticeable contributor to the cost increase. 

Here are some numbers to compare with capsules and station modules. Right now it looks like most station parts are around the $2,500/seat pricing. I would wonder if it is better to balance them from a cost per volume rather than cost per seat? Seeing as living volume is a more important factor to reduce stress build up with Kerbalism. Or some formula to also account for total fuel/SM volume included with the part.

maybe align the cost/volume of the station modules to be similar to the descent module cost/volume pricing (Voskhod and Apollo being outliers).
![Capture](https://github.com/user-attachments/assets/a606fbab-d697-4ca2-98ae-91f567506524)

siimav:
Skylab cost estimates are easy enough to find: https://www.gao.gov/assets/b-172192.pdf
![image](https://github.com/user-attachments/assets/9e06d9e6-1933-4452-aece-a9b35443caf0)

PR #2699, opened 2026-02-12
Space Telescopes Program restructure

wyllowthewisp:
A continuation and rework of #2442. The program now uses a wrapper contract that requires and tracks the telescope experiment completion, and offers optional contracts to support individual launches. This gives the player much more freedom to approach the program objectives, and should have room for multiple strategies (very much inspired by Early Space Stations). 

The servicing contracts (save GOCrewedServicing) have also been disabled for now, since they didn't really work and caused bugs. I also abandoned the effort to include an L2 space telescope contract, although principia users may find it is worth attempting for the transmission data rate benefits.

Overall, it should be much more playable, and includes a configurable start for testing.

To-do:
-Configure final parts from various telescope mods
-Housekeeping pass (contract descriptions, etc.)
-Second balance pass on program funding and telescope part costs (which should probably be much higher, similar to #2414).


Clayell:
Just noting that prestige levels for this should be updated according to https://github.com/KSP-RO/RP-1/pull/2776.

pap1723:
Small issue found here with the disconnect between the Description and resource required
image

hayghostin:
is this still being worked on?

GameData/RP-1/Contracts/Great Observatories/GOCrewedServicing.cfg
```diff
@@ -0,0 +1,230 @@
...
+			VALIDATE
+			{
+				MODULE
+				{
+					name = Experiment
+					experiment_id = RP0BasicSpaceTelescope2
+				}
+			}
+			title = Craft must have an Improved Space Telescope
+		}
```
periodically-makes-puns, line 112: As noted in #2766, this VALIDATE check does not actually check anything. This is most notable with the AtMost parameters in some contracts here, which auto-fail since both VALIDATE sub-parameters are immediately fulfilled due to being vacuously true.

See https://github.com/KSP-RO/RP-1/pull/2766/changes for the new contract parameters, and note that changing VALIDATE to FILTER also appeared to work [while testing](https://discord.com/channels/319857228905447436/840034852261593109/1498333113909121135).

PR #2703, opened 2026-02-27
Leader rebalance

njits23:
Currently designer leaders have effects on research speeds and scientist leaders have effects on integration speeds.
This PR aims to make designers no longer affect research and make scientists no longer affect integration.

A further goal of the changes is that I want leaders to set the direction of your space program, rather than them encouraging leader shopping to get short-term buffs. Directions such as staged combustion focus vs hydrolox focus, solid rocket motor focus, crewed vs uncrewed focus, etc.

Changes are not yet complete. The PR is being submitted as a draft to gather some preliminary feedback.
Changes to designers are already complete.


njits23:
I would like to note I'm not happy about the effects the Administration leaders have, but I do not have any better ideas.

pap1723:
I have been looking at the leader rebalancing as well. It was one of the features that I had been wanting to add to the game for a long time. Unfortunately, we only partially completed it and it has fallen flat.

I like a lot of the ideas that you have tried to accomplish here. My problem with the overall adjustments are they will seemingly allow players to accomplish even more milestones very early. That is a bad thing in my opinion.

I think it might be better to have some weaker leaders early. Even if that means the most famous people like Von Braun, Korolev and Glushko do not come around until later. Then, like pmpuns said, have the major milestones ACTUALLY be really rewarding by unlocking leaders with more significant buffs. The rewards a player earns for those milestones are a little flat right now in-game compared to historical significance.

Another way to approach that could be to have different cooldown times depending on the position. Changing the Administrator should be a bigger deal than swapping out some contractors, which was done all the time in real life.

I also really like the idea of ramping up effectiveness of leaders. To me that would be a more natural way of incentivizing not swapping.

I also agree that the Administrators rewards seem uninteresting. Now, I am not a spreadsheet player. I do not try to min-max, and those of you that do, might REALLY understand the benefits of 10% more confidence or reputation, or penalties against them. I would guess for the majority of the players, that doesn't mean much to them. I don't know a good answer for the administrators, but they should probably be the most impactful, and players should be able to understand their impact without having to know all paths their future space program will have.

One idea that was talked about at one time was having the Administrators be more directional steering than anything else. Essentially you'd have something like Korolev with a -20% research modifier (longer research) and a -20% integration costs (cheaper). So it would mimic the Soviet type of find 1 design that works really well and make a lot of them faster than the US could.

I don't know if any of the ideas are good, bad or otherwise. I think this is a complicated problem and needs a lot of discussion to try and nail down correctly.

njits23:
In the end everything boils down to how you acquire and spend your money. A leader that reduces research rates by 20% but makes integration 20% cheaper is a net-negative on your cashflow. Research is the single biggest expense of a space program. If you make that 20% more expensive but make another thing that cost less to begin with 20% cheaper you are hurting yourself. A leader like that only becomes interesting once your bottleneck is launches and not research, but that just never happens in both normal and speedrun progression.

As for these changes speeding up the achieving of milestones, there is some kernel of truth to that, but it would at best gain you a few months on your moonboots date whilst it still steers your research priority in a specific direction. There is no way to avoid speeding up milestones if you want leaders to give interesting research boosts.

WarlordFTW:
I agree with a lot of statements that have been made here. I feel like Admins should have the biggest impact on what you do but not in a negative direction. I think Korolev in the current system is maybe the best example of what's good and bad in the current state. I think the science reduction in exchange for cheaper everything is great. It forces the player to chose the direction of their space program but I don't think it's too impactful. it doesn't fundamentally change the way I play. at the same time if the sacrifice was any higher, I would NEVER pick korolev. I think administrators should fundamentally change how you manage programs and facilities and nothing else. Von braun kinda does that but again doesn't feel impactful. I think, starting with separating different playstyles and going from there is a good idea. Here's my suggestion:

One playstyle might be someone who runs a bunch of LCs with very narrow focuses (i think this is what von braun wanted to be). For this, I suggest -50% LC construction and modification cost and time, +100% LC efficiency gain, -75% LC adjustment range (The mass limits for an LC. I remembered this being recently added as a configurable parameter). The point is LCs build up quickly and cheaply but are much less versatile.

Another might be a tech advanced playstyle where you really push tech. For this I'd say +10% science, +20% research rate, -40% integration speed, -25% Unlock credit. So this one focuses on pushing deep into tech and taking it slow.

I think the speedrunning rosen is pretty good as is.

Then there's the One LC to do it all playstyle. -50% LC efficiency gain, +100% LC adjustment range, -50% untooled penalty (So using untooled parts is not nearly as bad.)

Then there's a launch launch launch style. +100% program slots, -50% conf for program, -20% research speed, -5% science gained.

These are just some suggestions where I think they are forcing you to pick a playstyle and stick with it. Another cool mechanic can be that the impact of the leader is dependent on milestones. so each milestone increases the impact of your leader further. And I would honestly consider making it almost impossible to remove an admin. like go down to 0 rep 0 conf or smth.

PR #2710, opened 2026-03-10
Asteroid Exploration Revamp

CXG-2827:
Revised the asteroid program to require 1 landing on either Vesta or Ceres. Additionally added 2 new optional contracts for sample returns.

Given the Delta-v requirements to enter into an orbit of these asteroids, I have changes the $/year to be on par with the Venus, Mars, and Mercury programs.

Nominal duration reduced to 10 years, and slots increased to 3.


uriyuzhui:
1. From a gameplay perspective, I would suggest keeping Ceres/Vesta program at 2-slot. All other late-game programs are 3 slots at least, so having one (or more) 2-slot program would create more strategic decision-makings (e.g. upgrade Admin vs not), rather than "just pick the best-paying one, they are all 3-slot anyway". "2-slot" is also consistent with IRL Dawn mission being Discovery-class at NASA.
2. The difficulty of Ceres/Vesta missions would be underestimated if using dV as the sole metric. They are outer than Mars, which means one would need power system prepared for longer duration. Either use advanced solar panels (typically one-tier higher than what we use at Mars), or use RTG which would also bump-up the mission cost. This would be even more consequential if a round-trip (sample return) is required.

TL;DR: keep it 2-slot, jack up total payment

CXG-2827:
From my testing thus far, I can launch both Ceres and Vesta landers via a 640T Proton LV, using Al Gridded for the transfer, capture and landing stages (23.2T in LEO). I'll be taking this program in 1975 on my current save, so these designs are working around the '75 launch windows.

I haven't yet designed a sample return mission, but I am guessing it will need a 1000T+ LV

WithCrow:
Having playtest it in an organic save, I have to say it was quite a lot of fun. As I mentioned before, going from orbit to landing is not really too challenging, but it was fun to do.

CXG-2827:
Did you try any of the optional sample returns?

I also noticed while running this program that the orbit contracts do not require a new vessel. I'm wondering if this should be changed to enforce the new vessel requirement. Otherwise, there is no bearing on Mission control level/contract slots, since you can just accept it mid-mission.

WithCrow:
I have not attempted sample returns - I would not enforce the orbit require a new vessel. I think it's fine to require it for the lander

CXG-2827:
Something I was wondering. Should the program require landing on both asteroids given how little extra delta V is needed once you are in orbit? or leave the program so you can just land on one?

WithCrow:
I am a strong advocate for requiring a landing on both. Like you said, it's trivial to land once you're already there. I must have misread things - I had assumed that was already the case and sent two lander missions

GameData/RP-1/Contracts/Asteroid Exploration/Ceres Landing.cfg
siimav, line 33: All such silly comments have to get removed.

siimav (changes requested):

siimav:
I wonder if it would make sense to tie KSP 🥔roids into this program somehow as well.

GameData/RP-1/Contracts/Asteroid Exploration/Ceres Landing.cfg
Clayell: They're gone already. They just haven't rebased their PR recently.

CXG-2827: I rebased. I might have missed some new updates, but are all contracts suppose to be "Trivial"? I notice all but two of the asteroid ones are Trivial. The exception being that the orbits are Significant

gabo463: if its the issue. there has been a change where all optional are trivial and all required are significant.

Clayell: Optional contracts should be trivial, required contracts should be significant, capstone contracts should be exceptional.

WithCrow:
Weird that it required you to apply the prestige levels again - I took care of that a while ago🤔

PR #2713, opened 2026-03-18
Add Knes configs

WithCrow:
This PR aims to add configs for Knes, which adds more European spacecraft. Currently, only Viking & Hermes are configured. 

Neither are balanced in how they were assigned, but Viking is added in the same node as Gemini, whereas Hermes is added in the Shuttle's node. Both are ahistorically early - Viking was proposed for ISS, Hermes for Ariane 5 - but are moved up in similar fashion to other European parts in the tech tree. 

Viking, as a single seat capsule, is not more capable than Gemini. It does have independent RPOD capability, but at 3 tons, it is heavier than what a player could assemble with a Mercury capsule.

[RO](https://github.com/KSP-RO/RealismOverhaul/pull/3127)
[ROKerbalism](https://github.com/KSP-RO/ROKerbalism/pull/191)


WithCrow:
As a single seater capsule without EVA capability is useless for the Gemini era, I've moved it up to Mercury. This would leave the European path unable to complete adv crewed until they get MRC or use the Voskhod EVA airlock.

WithCrow:
Training times for Viking (Mercury + Docking), MRC (Apollo) and both Hermes variants (Between DynaSoar & Shuttle) have been added

WithCrow:
Entry Cost modifiers have been added to align with what is in the mod currently. Viking is a cheaper 2nd gen capsules, MRC a slightly cheaper, but cruder Apollo. Hermes & Callisto are both using spaceCockpit.

WithCrow:
The ECM has finally been fixed by tweaking the part names in the RO patch. I've also added the ATV & STEAM parts, as both were already configured for RO.

WithCrow:
image
Kliper (shown on the left) and three additional smaller station parts have also been added. The former is a real proposal by Energia to replace Soyuz, the latter an extrapolation of 'What if they wanted to do Columbus Man-Tended Free Flyers on Ariane 4?'

Kliper is placed into the Improved Spaceplanes node (for now), with the station modules placed into the Station Development node.

WithCrow:
Upon thinking about it for a few days, I think I'll move Kliper to the node after Improved spaceplanes & also (in yet another set of PRs) place the Lockheed CEV there. 

That a) gives soviet players a nice progression from Buran (Improved Spaceplanes) and b) means US & USSR/(in this case)Russian playthroughs have the same tech at the same time (and when it would have flown IRL)

WithCrow:
I've attempted to configure the solids once more, but can't seem to get them to work. This might be a problem on the [RO side](https://github.com/KSP-RO/RealismOverhaul/pull/3127), or it might be a skill issue on mine - I would greatly appreciate if someone could take a look at this. It must be something obvious, because while Zefiro 23 & 40, as well as P80 and P120 don't work, EAP and P120C *do* work for some reason.

WithCrow:
I can't believe I never noticed this, but Viking can just EVA. this means it's a perfect fit for the Gemini era.

PR #2723, opened 2026-03-23
Fill in _serverUrl with the RP1 career log one by default

Clayell:
This should save a copy and paste step for anyone using the career log, while still leaving the option open for other APIs (are there any?)

image


Clayell:
How many years has it been this domain? If that day comes, its a 1 line fix to change it over. It's not like it's the only place we'd need to update this link anyway.

PR #2735, opened 2026-04-02
Add sun angle requirements to photography experiments

### LINKED ISSUE #2669: Sun Angle Requirement for film cameras
Adding the already existing requirement from TV cameras to film cameras as well. This is only an issue instead of a PR because I'm not actually sure there were no night-time film photos - but it would sync them & make everything a bit more cohesive.

sparr:
Fixes #2669 which points out that most film/still camera satellites required sunlight to work up until there were infrared and radar satellites that came later than our last current experiments (e.g. late KH-11) . Up to KH-9 / Photography 4 I have added the same 85 degree max sun angle as the early video experiment, since they used relatively arbitrary orbits and took occasional photos throughout the local day. For early KH-11 / Photography 5 I added a min sun angle similar to the later video experiments which mirrors early KH-11's real use of sun-synchronous orbits to always take photos at local not-noon to get consistent useful shadows.

### DIFF
--- a/GameData/RP-1/Science/Experiments/Photography.cfg
+++ b/GameData/RP-1/Science/Experiments/Photography.cfg
@@ -58,9 +58,9 @@ EXPERIMENT_DEFINITION
 		%ec_rate = 0.005
 		%data_rate = 20
 		@data_rate /=  600 //10 minutes
-		%requires = 
+		%requires = SunAngleMax:85 // Can capture with almost any amount of sunlight, like RP0visibleImaging1
 		%resources = 
-		%experiment_desc = Early film flown on V2 rockets, taking the first photos from space. Can also be used as a reconnaissance camera. Requires physical recovery.
+		%experiment_desc = Early film flown on V2 rockets, taking the first photos from space. Can also be used as a reconnaissance camera. Requires daylight on the planet below the camera while running the experiment, and physical recovery to collect the science.
 		allow_shrouded = false
 	}
 }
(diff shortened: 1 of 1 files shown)


siimav:
Still not convinced this is a good idea. While realistic, it sounds more like a *gotcha* than something that adds gameplay value. Or makes you design something differently.

pap1723:
I would probably agree with siimav on this one. Hopefully my fixes for SCANsat make it able to be used with RSS and then these contracts would be replaced by required SCANsat coverage.

WithCrow:
> While realistic, it sounds more like a _gotcha_ than something that adds gameplay value.

I disagree here - it does add value in when you launch, as well as further things to consider for the orbits you send them in. But it also makes me wonder: what distinguishes the requirement here as a gotcha, but does not for the video experiments?

siimav:
Sorry, I meant specifically for cameras launched on sounding rockets. Either you warp to the correct time of day and get the science or you don't. As for orbital film - I guess it could serve the purpose of directing the player to launch into SSO (I assume this satisfies the 85 degrees angle?). Otherwise it just makes the experiment take roughly twice as long.

siimav:
Anyone willing to check whether this affect the suborbital film camera experiment or if it's only for the experiments meant to go into orbit?

periodically-makes-puns:
RP0photos1 has the SunAngle requirement, so it affects the suborbital camera.

sparr:
Repeating my comment from #2669:

Getting players used to choosing their launch time intentionally seems like a good thing to do early on. It took me far too long in vanilla KSP to realize I could do polar orbit contracts by waiting on the ground rather than doing huge normal burns.

PR #2742, opened 2026-04-09
Add effective cost breakdown tab in editor

periodically-makes-puns:
image

[An untooled version of a Moonboots vehicle, because I needed something realistic with a 6-digit cost.]

The current implementation works, but it feels a little hacky and there's more logic duplication than I'm happy with. It also removes the debug log line in `GetEffectiveCostInternal` because otherwise we would start spamming the log.

This does also mean we recompute the cost of the vessel every 0.5s while the vessel is open, which could be computationally nasty for large vessels on weak computers.

Leaving as a draft for now to gather feedback on UI/UX and implementation details. It is fully functional, though.

### DIFF
--- a/Source/RP0/SpaceCenter/Projects/VesselProject.cs
+++ b/Source/RP0/SpaceCenter/Projects/VesselProject.cs
@@ -116,6 +116,15 @@ namespace RP0
         private double _buildRate = -1d;
         private double _leaderEffect = -1d;
         public double LeaderEffect => _leaderEffect < 0 ? UpdateLeaderEffect() : _leaderEffect;
+        private double _modifiedEC = -1d;
+        public double ModifiedEC 
+        { 
+            get
+            {
+                if (_modifiedEC < 0) UpdateLeaderEffect();
+                return _modifiedEC;
+            } 
+        }
 
         internal ShipConstruct _ship;
 
@@ -1287,15 +1296,15 @@ namespace RP0
 
         public double UpdateLeaderEffect()
         {
-            double modifiedEC = effectiveCost;
+            _modifiedEC = effectiveCost;
             double globalMult = ApplyGlobalCostModifiers();
             foreach (var t in tagEffectiveCosts)
             {
                 double ec = t.ec * Leaders.LeaderUtils.GetPartEffectiveCostEffect(t.tags);
-                modifiedEC += (ec - t.ec) * globalMult;
+                _modifiedEC += (ec - t.ec) * globalMult;
             }
-            modifiedEC *= Leaders.LeaderUtils.GetGlobalEffectiveCostEffect(globalTags, resourceAmounts);
-            double modifiedBP = Formula.GetVesselBuildPoints(modifiedEC);
+            _modifiedEC *= Leaders.LeaderUtils.GetGlobalEffectiveCostEffect(globalTags, resourceAmounts);
+            double modifiedBP = Formula.GetVesselBuildPoints(_modifiedEC);
             if (modifiedBP < 1d)
                 modifiedBP = 1d;
 
--- a/Source/RP0/UI/CostBreakdownGUI.cs
+++ b/Source/RP0/UI/CostBreakdownGUI.cs
@@ -17,9 +17,9 @@ namespace RP0.UI
             public double effectiveCost;
             public double effectiveCostModifier;
 
-            public override bool Equals(object o)
+            public override bool Equals(object obj)
             {
-                if (o is CostEntry other)
+                if (obj is CostEntry other)
                     return Equals(other);
                 else
                     return false;
--- a/Source/RP0/UI/CostBreakdownGUI.cs
+++ b/Source/RP0/UI/CostBreakdownGUI.cs
@@ -11,11 +11,29 @@ namespace RP0.UI
 {
     public class CostBreakdownGUI : UIBase
     {
-        private record struct CostEntry
+        private struct CostEntry
         {
             public string name;
             public double effectiveCost;
             public double effectiveCostModifier;
+
+            public override bool Equals(object o)
+            {
+                if (o is CostEntry other)
+                    return Equals(other);
+                else
+                    return false;
+            }
+
+            private bool Equals(CostEntry other)
+            {
+                return effectiveCost == other.effectiveCost && effectiveCostModifier == other.effectiveCostModifier && name == other.name;
+            }
+
+            public override int GetHashCode()
+            {
+                return (name, effectiveCost, effectiveCostModifier).GetHashCode();
+            }
         }
         
         private float _lastUpdateTrigger = 0f;
--- a/Source/RP0/UI/CostBreakdownGUI.cs
+++ b/Source/RP0/UI/CostBreakdownGUI.cs
@@ -1,0 +1,186 @@
+﻿using KSPCommunityFixes;
+using RealFuels;
+using Smooth.Slinq;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+
+namespace RP0.UI
+{
+    public class CostBreakdownGUI : UIBase
+    {
+        private record struct CostEntry
+        {
+            public string name;
+            public double effectiveCost;
+            public double effectiveCostModifier;
+        }
+        
+        private float _lastUpdateTrigger = 0f;
+        private readonly Dictionary<CostEntry, int> _partCosts = new Dictionary<CostEntry, int>();
+        private readonly List<Part> _singlePart = new List<Part>(); // dummy list for calling GetEffectiveCost, to avoid extra allocations
+        private Vector2 _partCostsScroll = new Vector2();
+        private bool _isToolingTempDisabled = false;
+        private IEnumerator _activeUpdate = null; 
+
+        protected override void OnStart()
+        {
+            GameEvents.onEditorShipModified.Add(ShipModifiedEvent);
+        }
+
+        protected override void OnDestroy()
+        {
+            GameEvents.onEditorShipModified.Remove(ShipModifiedEvent);
+        }
+
+        public UITab RenderCostBreakdownTab()
+        {
+            VesselProject vessel = SpaceCenterManagement.Instance?.EditorVessel;
+            if (vessel == null)
+            {
+                GUILayout.BeginHorizontal();
+                GUILayout.Label("No vessel currently detected.", HighLogic.Skin.label, GUILayout.Width(500));
+                GUILayout.EndHorizontal();
+                return UITab.CostBreakdown;
+            }
+
+            GUILayout.BeginHorizontal();
+            GUILayout.Label(new GUIContent("Total Effective Cost", "Effective cost is a metric that includes the various cost multiplier tags attached to each part and resource.\n" +
+                "This is usually the preferred metric for cost comparison between different rockets."), HighLogic.Skin.label, GUILayout.Width(312));
+            GUILayout.Label($"{vessel.effectiveCost:N2}", RightLabel, GUILayout.Width(144));
+            GUILayout.EndHorizontal();
+
+            GUILayout.BeginHorizontal();
+            GUILayout.Label(new GUIContent("Effective Cost after Leader effects (conditional only)", "Unconditional integration speed buffs/debuffs are ignored."), HighLogic.Skin.label, GUILayout.Width(312));
+            GUILayout.Label($"{vessel.ModifiedEC:N2}", RightLabel, GUILayout.Width(144));
+            GUILayout.EndHorizontal();
+
+            DrawHorizontalSeparator();
+
+            GUILayout.BeginHorizontal();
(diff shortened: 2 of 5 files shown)


Source/RP0/SpaceCenter/Projects/VesselProject.cs
```diff
@@ -1164,7 +1164,7 @@ private double GetEffectiveCostInternal(object o)
 
             UpdateTagECs(effectiveCost);
 
-            RP0Debug.Log($"Eff cost for {name}: {effectiveCost} (cost: {cost}; dryCost: {dryCost}; wetMass: {wetMass}; dryMass: {dryMass}; partMultiplier: {partMultiplier}; moduleMultiplier: {moduleMultiplier})");
+            // RP0Debug.Log($"Eff cost for {name}: {effectiveCost} (cost: {cost}; dryCost: {dryCost}; wetMass: {wetMass}; dryMass: {dryMass}; partMultiplier: {partMultiplier}; moduleMultiplier: {moduleMultiplier})");
```
siimav: RP0Debug doesn't actually log when built in release mode

Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -0,0 +1,189 @@
...
+        {
+            
+            var toolingModule = p.FindModuleImplementing<ModuleTooling>();
+            if (!(toolingModule is null)) 
+                // for single-tooling module parts we can insert this info to help disambiguate
+                // TODO: make this consider only tooled cost if we can
+                return $"{p.partInfo?.title} ({toolingModule.ToolingTypeTitle}) {toolingModule.GetToolingParameterInfo()}";
+            
+            var engineModule = p.FindModuleImplementing<ModuleEngineConfigsBase>();
+            if (!(engineModule is null))
```
siimav: Not a fan of this syntax. Just use `!= null`. Using `is` could also be one of those ways that bypass Unity null checks, I can never remember.

Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -0,0 +1,189 @@
...
+        private string GetPartDisplayName(Part p)
+        {
+            
+            var toolingModule = p.FindModuleImplementing<ModuleTooling>();
+            if (!(toolingModule is null)) 
+                // for single-tooling module parts we can insert this info to help disambiguate
+                // TODO: make this consider only tooled cost if we can
+                return $"{p.partInfo?.title} ({toolingModule.ToolingTypeTitle}) {toolingModule.GetToolingParameterInfo()}";
+            
+            var engineModule = p.FindModuleImplementing<ModuleEngineConfigsBase>();
```
siimav: Better to start using `FindModuleImplementingFast` from KSPCF. The KSP method is kinda ass and doesn't build the cache when a PM isn't found which is nearly always here.

Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -0,0 +1,189 @@
...
+        private record struct CostEntry
+        {
+            public string name;
+            public double effectiveCost;
+            public double effectiveCostModifier;
+        }
+        
+        private float _lastUpdateTrigger = 0f;
+        private Dictionary<CostEntry, int> _partCosts = new Dictionary<CostEntry, int>();
+        private double _globalModifier = 1;
```
siimav: Sonar has valid suggestions for once. _lastUpdateTrigger is even completely unused?

Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -0,0 +1,189 @@
...
+            if (_lastUpdateTrigger < now)
+            {
+                if (!(_activeUpdate is null))
+                    UIHolder.Instance?.StopCoroutine(_activeUpdate);
+                _activeUpdate = Update(ship, now);
+                UIHolder.Instance?.StartCoroutine(_activeUpdate);
+            }
+        }
+
+        private IEnumerator Update(ShipConstruct ship, float triggerTime) {
```
siimav: Check the Sonar suggestion. Also, formatting.

Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -0,0 +1,189 @@
...
+            }
+            else if (_isToolingTempDisabled && Event.current.type == EventType.Repaint)   // button events are handled on the Repaint pass
+            {
+                ToolingManager.Instance.toolingEnabled = HighLogic.CurrentGame.Mode == Game.Modes.CAREER;
+                _isToolingTempDisabled = false;
+                GameEvents.onEditorShipModified.Fire(EditorLogic.fetch.ship);
+            }
+        }
+
+        private string FormatCostAndMultiplier(double cost, double multiplier)
```
siimav: What multiplier? Appears to be from leaders?

siimav (changes requested):

Source/RP0/UI/CostBreakdownGUI.cs
periodically-makes-puns: Yeah, the multiplier is the product of all leader cost multipliers for that given part/vessel.

For the global multiplier (in the second line), it currently only includes global conditional modifiers. Per-part modifiers are not displayed in the multiplier there.

e.x. in the image above, it displays (-9%) from Max Faget's effect on human-rated vehicles, but the actual total reduced cost is much higher because of Douglas's effect on isogrids, which is displayed next to the cost for each isogrid tank.

Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -0,0 +1,191 @@
...
+            // I would *love* to use EditorVessel, but it appears the per-part effective-cost methods there have side-effects.
+            // So instead we build the vessel from scratch, hooray?
+            // I'm really not a fan of this logic duplication. 
+            yield return null;
+            // Stall a frame.
+
+            var parts = ship?.Parts; 
+            if (parts?.Count > 0)
+            {
+                VesselProject vessel = new VesselProject();
```
siimav: OK, this is really confusing indeed. What side effects did you observe?

Source/RP0/UI/CostBreakdownGUI.cs
siimav: I'd still prefer to get rid of those is null checks. They're honestly ugly when negated and bypass Unity's overloaded operator.

periodically-makes-puns: Those should be != now instead of `is null` in the latest commit

Source/RP0/UI/CostBreakdownGUI.cs
periodically-makes-puns: I tried using GetEffectiveCost to grab the effective cost for each part, but that resets the tracked global tags and stored resource amounts every time it’s called (since it usually appears to be called with the list of all parts on the vessel? but I need specifically the cost of one part)

Source/RP0/UI/CostBreakdownGUI.cs
siimav: There's apparently more

periodically-makes-puns: should all be gone now, did a find/replace just to double check

Source/RP0/UI/CostBreakdownGUI.cs
siimav: I tried refactoring the EC code to be more suitable for this purpose. It's completely untested though.
https://github.com/KSP-RO/RP-1/commit/aedf389d944f759abf902e271396cd520990b8f9

periodically-makes-puns: I'll test it a bit and rewrite this code for that refactor if it works. Looking through the refactored code, I think I can swap this code over to report total BP (and leader-adjusted BP) instead by reading off `LeaderEffect`. Alternatively, if `VesselProject.UpdateLeaderEffect` reported the modified EC somewhere that would also work, it's just that the modified EC is currently hidden (and BP -> EC is not a conversion I want to do).

periodically-makes-puns:
I'm leaning towards removing the displayed percentage for the leader-adjusted effective cost (the second line). It's mismatched, since the percentage only accounts for global reduction effects, while the actual amount displayed next to it accounts for all conditional leader effects. Also, it would remove some logic from here and rely more on the actual VesselProject code.

I think the only remaining duplicated logic after rewriting for the refactor should be the application of leader effects to the EC of each part. The bottom list displays the leader effect as a percentage cost reduction/increase next to the leader-modified EC (but not global-tag modified, because that applies later).

siimav:
I'm not quite sure what you need with leader effects but you're welcome to propose changes to the existing code.

periodically-makes-puns:
EC calculation refactor appears to work, or at least nothing is horribly broken. I updated the GUI code to work with the refactor (and realised that I could in fact just pull some variables out of EditorVessel instead of manually calculating them)

siimav:
Merged the branch so this needs a rebase now.

periodically-makes-puns:
Squashed and rebased.

Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -0,0 +1,186 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+
+namespace RP0.UI
+{
+    public class CostBreakdownGUI : UIBase
+    {
+        private record struct CostEntry
```
siimav: Bot no likey: Error: /home/runner/work/RP-1/RP-1/Source/RP0/UI/CostBreakdownGUI.cs(14,24): error CS1001: Identifier expected [/home/runner/work/RP-1/RP-1/Source/RP0/RP0.csproj]

periodically-makes-puns: Ah. I'm guessing it doesn't like the `record struct`, which I think is a C# 10 thing?

periodically-makes-puns:
Works locally, and is now `record`-less.

Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -0,0 +1,204 @@
...
+            }
+
+            private bool Equals(CostEntry other)
+            {
+                return effectiveCost == other.effectiveCost && effectiveCostModifier == other.effectiveCostModifier && name == other.name;
+            }
+
+            public override int GetHashCode()
+            {
+                return (name, effectiveCost, effectiveCostModifier).GetHashCode();
```
siimav: What arcane sorcery is this? :starebeale:

Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -0,0 +1,204 @@
...
+
+        private IEnumerator Update(ShipConstruct ship) 
+        {
+            yield return null;
+            // Stall a frame.
+
+            var parts = ship?.Parts; 
+            if (parts?.Count > 0)
+            {
+                _partCosts.Clear();
```
siimav: Shouldn't this be outside the `if`? So that the list is empty after resetting the editor.

Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -0,0 +1,204 @@
...
+
+            DrawHorizontalSeparator();
+
+            GUILayout.BeginHorizontal();
+            GUILayout.Label("Part Name", BoldLabel, GUILayout.Width(312));
+            GUILayout.Label(new GUIContent("Ind. Effective Cost", "Global effective-cost modifiers are not included unless they originate from the part."), BoldRightLabel, GUILayout.Width(144));
+            GUILayout.EndHorizontal();
+
+            _partCostsScroll = GUILayout.BeginScrollView(_partCostsScroll, GUILayout.Height(300), GUILayout.Width(500));
+            foreach (var kvp in _partCosts.OrderByDescending(kvp => kvp.Key.effectiveCost * kvp.Key.effectiveCostModifier * kvp.Value))
```
siimav: Better to sort at the end of updating the part list. Unless... there's some kind of switchable sort mode coming.

Source/RP0/UI/CostBreakdownGUI.cs
periodically-makes-puns: I wasn’t sure if the base GetHashCode implementation for structs worked, but I know that using tuples as keys in a Dictionary works as expected. So I just borrow that behaviour here.

I should probably test if the base implementation works first before assuming that it doesn’t…

Source/RP0/UI/CostBreakdownGUI.cs
periodically-makes-puns: If someone complains about the default sort order (descending order of total cost), I’ll implement that. For now I’m just going to move the sort.

Source/RP0/UI/CostBreakdownGUI.cs
periodically-makes-puns: Base implementation works, no arcane sorcery required.

PR #2747, opened 2026-04-16
Remove apoapsis check for 3 crewed orbital

Clayell:
It doesn't make sense to have a constraint for periapsis, apoapsis, AND eccentricity. Additionally, the way that the periapsis and apoapsis are set up means that they can rarely intersect with each other, as seen [here](https://discord.com/channels/319857228905447436/512556137380315139/1494155775126405273) (periapsis above 300km and apoapsis below 300km)

### DIFF
--- a/GameData/RP-1/Contracts/Earth Crewed Adv/OrbitalFlight3.cfg
+++ b/GameData/RP-1/Contracts/Earth Crewed Adv/OrbitalFlight3.cfg
@@ -112,12 +112,6 @@ CONTRACT_TYPE
 	}
 	DATA
 	{
-		type = int
-		startApA = 300000 + Round(Random(0, 200000), 25000)
-		title = Apogee
-	}
-	DATA
-	{
 		type = double
 		Eccentricity = 0.0034
 		title = Eccentricity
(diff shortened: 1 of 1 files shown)


pap1723:
The idea behind this originally was a holdover from the original RP0 contracts and was so that you didn't have the same contract and sometimes it would be harder. I am good removing this as it is very arbitrary.

GameData/RP-1/Contracts/Earth Crewed Adv/OrbitalFlight3.cfg
```diff
@@ -111,12 +111,6 @@ CONTRACT_TYPE
 		title = Perigee
 	}
 	DATA
-	{
-		type = int
-		startApA = 300000 + Round(Random(0, 200000), 25000)
```
siimav, line 116: Yeah but now you have a contract parameter that still references this value.

siimav (changes requested):

GameData/RP-1/Contracts/Earth Crewed Adv/OrbitalFlight3.cfg
Clayell: Didn't see that, thanks. Should be good now.

PR #2749, opened 2026-04-19
Add Early Spaceplanes Program

WithCrow:
This PR aims to reactivate and modernise [1935](https://github.com/KSP-RO/RP-1/pull/1935). The program is intended for something like an X-20 or equivalent light spaceplane, with capabilities similar to Gemini. The contracts include:

- Suborbital demonstration flight
- First orbital flight
- Rendezvous
- Docking

As well as optional additional flights to mimic the recon flights proposed for X-20. The program is currently set up as a side grade to Gemini, but could be taken in parallel with Crewed Orbit if one so desired. 

Credit for the contract architecture & amost all descriptions goes to Sound & Fury.


WithCrow:
image
In a test career, it takes a ~200 ton LC with 780 engineers to launch a DynaSoar. With 1000 researchers and 780 engineers and the program on Fast, the player gets a decent surplus. At least in my eyes, this sounds about right for a program of this era. 

DynaSoar's 100k EC, though, is quite excessive when compared to Gemini's 28k EC, but that is outside the scope of this PR.

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -735,6 +735,60 @@ RP0_PROGRAM
...
+	repToConfidence = 3
+	slots = 3
+	
+	REQUIREMENTS
+	{
+		complete_program = EarlyXPlanes
+		ANY
+		{
+			complete_program = CrewedOrbitEarly
+			active_program = CrewedOrbitearly
```
panic-7700:
```
			active_program = CrewedOrbitEarly
```
resolves the attached:

Image

WithCrow: What a silly mistake! Fixed.

billy54:
I'm not sure if requiring completion of X-Planes to accept this program is a good idea. A Soviet player wanting to do the Spiral spaceplane may not have done planes. But on the other hand xplanes is logically the first step in the airplane progression, and we have other ahistorical things in RP-1 like requiring lunar orbit before landing or allowing the player to orbit venus or mars on the first interplanetary mission.

WithCrow:
What prevents a soviet player from completing x-planes?

A reason for requiring x-planes is that a player should have the experience of managing a plane's thermals & suborbital flight before attempting this program, so I am hestitant to remove the x-planes requirement.

billy54:
My main concern was that the Soviets historically did not do xplanes, but then again they didn't do spaceplanes either, so I guess we should have it as a requirement.

panic-7700:
The Soviets definitely did do x-planes, just not quite in the same sense as the US did. Plenty of weird one-or-two-off designs out there from various design bureaus. The difference is that most of the Soviet stuff was at least on the surface directly intended to lead into military applications, while the US x-plane programs were more about figuring out solutions to very specific problems and rarely (if ever?) grew into production aircraft.

Not to get too nerdy and off topic, but a couple examples are [Yak-1000](https://en.wikipedia.org/wiki/Yakovlev_Yak-1000) (turned into a cruise missile), [MiG I-7](https://en.wikipedia.org/wiki/Mikoyan-Gurevich_I-7) (turned into, eventually, the MiG-21), [Yak-36](https://en.wikipedia.org/wiki/Yakovlev_Yak-36) (evolved into the production Yak-38), [MiG-105](https://en.wikipedia.org/wiki/Mikoyan-Gurevich_MiG-105) (subscale Spiral testing), the list does actually go on and on!

WithCrow:
With [the Advanced Spaceplanes](https://github.com/KSP-RO/RP-1/pull/2786) program growing more mature, I grow more confident in this program. Even if a player does not complete this program with something like DynaSoar but, instead, STS or Buran, this also works fine as a 'test flight' (initial STS flights or the first Buran flight).

In my own playtesting, I've seen quite a bit of value in combining this with Early Stations - DynaSoar is a very suitable crew & resupply shuttle.

WithCrow:
Massive thanks to @hayghostin for the program icons!

Glloyd:
I've been poking at this a little in my current career, and I do think that if Advanced Crew Orbit is complete, this program should be bumped down to 2 slots (at least), with a slight corresponding reduction in funds, and maybe vice versa as well. The goals and contracts of the two programs are fairly similar, the building upgrade requirements are the same, and you could also absolutely complete ACO with a spaceplane if you so desired. Taking this program on in the later game just doesn't feel worth it for the low funds relative to the program slots. The Advanced Spaceplanes program being locked behind this one does give some additional incentive, but it feels difficult to justify the 3 slots at most points after completing ACO, if you do that first (and probably likewise in the other direction if you do this program first).

WithCrow:
A good point! The reason the objectives are similar is that it is supposed to be a sidegrade, but I think you are right on the money. I was never a particularly huge fan of requiring players starting in 1978 to have to do this first.

I'll take a look!

Glloyd:
Yeah, that wasn't a criticism of the objectives of the program, I think it makes sense as a sidegrade, and as a leadup to the Advanced Spaceplanes Program. It's moreso that if you do one of the two, it doesn't feel worth the 3 slots to do the other one. I think lowering the slot count of one on the completion of the other might be better than the current setup, but I don't know if it's even the best way to resolve the issue. Doing this one first, for example, would probably make ACO feel nearly entirely irrelevant, because it doesn't even open up any other programs like this one does. I kind of think that a good way forward might actually be to very slightly raise the funds per slot per year of this program above ACO, but block ACO entirely if you do this one first, like with Heavy vs Light Sats. However, if you complete ACO first, then you can still complete this program as a lead-in to Advanced Spaceplanes, but with less slots and reduced funds. That allows you to still take the historical US path of ACO > Early Spaceplanes >Advanced Spaceplanes, or go with a spaceplanes-focused career and not be left with a now-irrelevant ACO program.

WithCrow:
Oh, don't worry - I got your meaning! Very good proposal all around

WithCrow:
Okay, so after some initial feedback, the following changes have been made:
- Funding raised to 600k/year for total of 2.4 million
- Descent angle increased to 20°
- Blocks ACO & vice-versa
- If ACO is done first, like on a historical US run, reduce the years to 3, slots to 2 & funding to flat 400k/year

siimav:
> Proponents also cite the possibility of using aerodynamic lift to perform plane changes with less propellant. 

Uhhh. What?

siimav:
A bunch of texts still say "Land with a descent angle lower than 10 degrees".
Btw, the parameter only measures a handful of seconds before landing so as long as you flare, it isn't even particularly important.

WithCrow:
> Uhhh. What?

That's the original program description, which I didn't mess with

WithCrow:
> A bunch of texts still say "Land with a descent angle lower than 10 degrees".

Bizarre, I could have sworn I caught all of those. 

> Btw, the parameter only measures a handful of seconds before landing so as long as you flare, it isn't even particularly important.

I don't see this as a massive argument against it - vertical landings are still impossible with this, or am I missing something?

WithCrow:
Fixed the parameters - I also removed the AC 4 requirement, since this never requires an EVA & hence does not need the same level. I also adjusted the conf cost & rep gain/loss to match the sheet. 

I also added ACO as a possible option instead of X-Planes. A player doing ACO first gets less funding for ESP, so this is a different route for players that do not want to do X-Planes first. Aligns with the soviet Spiral/Buran program, for example.

connorjak:
> > Proponents also cite the possibility of using aerodynamic lift to perform plane changes with less propellant. 
> 
> 
> 
> Uhhh. What?

This is a well researched maneuver. Dip periapsis into upper atmosphere, fly with angle of attack normal to orbital plane, exit with plane changed, having only lost some energy to drag.

PR #2750, opened 2026-04-19
Add Solar Exploration program

WithCrow:
This PR aims to add contracts for the Ulysses high inclination solar orbiter and the Parker Solar Probe, which flew extremely close to the Sun. 

Currently, it is set up as a 2 slot program, requiring the player to have completed Jupiter Observation in order to be accepted. My reasoning for this is that Ulysses used a Jupiter gravity assist for a massive inclination change and this means that a player needs to be *at least* capable of flying out to Jupiter.


Clayell:
Should resolve https://github.com/KSP-RO/RP-1/issues/2451 and https://github.com/KSP-RO/RP-1/issues/1477

WithCrow:
siimav's comment in 1477 does ring true - I am not quite happy with the program as is currently. Considering Asteroid exploration (at least for now) also only requires 2 launches, it is acceptable - but once 2710 is merged, this will stick out once more. 

I was thinking of something like ESA's Solar Orbiter, which would be pretty similar to what you mentioned in 2451.

periodically-makes-puns:
> Currently, it is set up as a 2 slot program, requiring the player to have completed Jupiter Observation in order to be accepted. My reasoning for this is that Ulysses used a Jupiter gravity assist for a massive inclination change and this means that a player needs to be at least capable of flying out to Jupiter.

Would a (Complete contract Jupiter Orbit OR Complete contract Jupiter Flyby) condition also make sense, then? That would allow this program to be combined with e.g. a Jupiter atmospheric probe or another Jupiter orbiter, and also allow the scenario where a player has launched a Voyager-esque mission for Outer Planet Flyby (and therefore done a precise Jupiter flyby) but hasn't completed Jupiter Observation.

pap1723:
These are both 21st century missions, correct? In that case, my argument would be post Galileo, which means Jupiter orbit.

WithCrow:
Ulysses was supposed to be launched in 1986 but actually launched in 1990, so it is not post Galileo. I would thus change it to either Orbit or Flyby - I really like the possibility of combining it with an atmospheric probe!

Clayell:
> I really like the possibility of combining it with an atmospheric probe!

Basically impossible in RSS (or even stock). You blow up from overheating long before you get that low

WithCrow:
> Basically impossible in RSS (or even stock). You blow up from overheating long before you get that low

I was referring to combining it with a Jupiter atmospheric probe, not a solar atmospheric probe

wyllowthewisp:
> siimav's comment in 1477 does ring true - I am not quite happy with the program as is currently. Considering Asteroid exploration (at least for now) also only requires 2 launches, it is acceptable - but once 2710 is merged, this will stick out once more.
> 
> I was thinking of something like ESA's Solar Orbiter, which would be pretty similar to what you mentioned in 2451.

A solar wind sample return mission like the Genesis spacecraft could fit well too. That would require a new experiment, but I worked on a potential solar science program a while back and should still have the finished configs for that contract/experiment someplace

wyllowthewisp:
STEREO and SOHO might be good missions to fill things out further, although the contract work for those might be more tricky due to their orbits (esp. SOHO, at L1). 

Alternatively, Close Solar Flyby could be reworked to allow a few contract repeats with lower periapsis targets, to mirror the Parker Solar Probe's gradually deeper dives

WithCrow:
I've decided to add a STEREO inspired contract, as I could not think of a way to codify the L1 position for a stock gravity system. Genesis would be cool to have! If you could find those configs, that'd be great!

wyllowthewisp:
Configs found. I'm going to test and make sure it all still works, but I should be able to send those over sometime in the next few days. Should I just submit them as a PR to your solar_program branch?

WithCrow:
That would be great! Time for some cool solar science!

WithCrow:

siimav:
The STEREO-A / -B contract probably needs some CC extension so that you cannot just plop 2 sats into a rather random solar orbit. Some combination of mean anomaly + distance from Earth?

PR #2753, opened 2026-04-21
Lower the cost of DynaSoar

WithCrow:
Motivated by [2749](https://github.com/KSP-RO/RP-1/pull/2749), this PR aims to lower the cost of DynaSoar, which presently has a higher EC (129k) than Apollo (~90 to 100k). When compared to the Gemini with its 30k EC.

This initial deduction of price was across the board and thus not tailored to the specific balancing applied to RP-1, but it halves the cost and brings it down to ~60k EC. Being twice as expensive as Gemini still does not make it more optimal, but should make it at least a bit more viable. 

This increased cost does, at least in my eyes, still include the added overhead of managing a spaceplane as opposed to an expendable capsule. Even if this current reduction is too much, I do not think it should cost more than Apollo - a far more complicated spacecraft in the grand scheme of things.


WithCrow:
After a discussion on the Discord, I have reduced the entry cost of DynaSoar drastically, as it was far more than Gemini & even costlier than Apollo. So, currently, it should cost around the same to unlock as Gemini, but still be more expensive per launch. There is an argument to be made to reduce the cost even further, but I am not quite convinced of that just yet.

WithCrow:
I've decided to broaden the scope of this PR to include all other spaceplanes, as their costs were not uniform. The stock spaceplane cockpits, for example, had not been given the 'Cockpit' and 'Re-entry' tags, even though they should have received them.

WithCrow:
A Titan GLV with Gemini costs 43k EC - launching DynaSoar (after reuse) is ~75k EC, with ~100k EC for the first launch. I do not know how to feel about this - Dyna on its own is 67k EC! So DynaSoar, on its own, is more expensive than launching an entire Gemini + Titan II, which is still far too much in my eyes. It should, even after reuse, be more expensive than Gemini - but that means it should be closer to 50k EC.

image
image

siimav:
> I've decided to broaden the scope of this PR

An unwise decision

WithCrow:
> An unwise decision

I disagree - all it does apart from the initial scope is add the correct tags to the other spaceplane parts, that are applied non-uniformly. 

But, if you prefer I revert to just lowering Dyna's (unlock) cost and make a separate one for the tags, I can do that

siimav:
I do prefer that.

WithCrow:
Aye aye captain - I'll get on that then. It seems I also missed a few parts anyway.

WithCrow:
There you go - this now only reduces the effective and unlock cost of DynaSoar. Will open a separate one for the other spaceplane parts

GameData/RP-1/Tree/EntryCostModifiers.cfg
```diff
@@ -551,7 +551,7 @@ ENTRYCOSTMODS
 		hypersonicCockpit = 25000
 		protoSpaceCockpit = 120000, hypersonicCockpit, capsulesAirlock
 		spaceCockpit = 500000, protoSpaceCockpit
-		X20 = 300000, spaceCockpit	//800000 total
+		X20 = 150000, protoSpaceCockpit	//295000 total
```
siimav: I'm getting 345000 total. This makes it significantly lower cost than Gemini at 477000.

WithCrow: I never claimed to be able to add numbers - what would be a suitable price? 1.5 * Gemini?

WithCrow: I've increased it to 545,000 - this makes it quite a bit more expensive than Gemini, especially for a normal player, as Gemini will have been prefaced by something like Mercury, which reduces that cost, whereas this has a very small entry cost decrease if someone flew the X-15

siimav: I do wonder though whether it would be better to increase protoSpaceCockpit cost instead. I find it a bit odd that you get generic orbital cockpits way cheaper than the X20.

WithCrow: I would prefer that as well, but I did not want to step outside of the scope of this PR again

Source/Tech Tree/Parts Browser/data/ROCapsules.json
```diff
@@ -2630,7 +2630,7 @@
         "title": "Dynasoar Wing",
         "description": "",
         "mod": "ROCapsules",
-        "cost": 2000,
+        "cost": 25,
```
siimav: 25, huh? That the entire wing? Even the elevons are 50.

Source/Tech Tree/Tree Years.xlsx
siimav, line 1: What changed here?

GameData/RP-1/Tree/EntryCostModifiers.cfg
siimav: Fair enough. I guess you could make a separate PR after this gets merged to rejigger the ECMs.

Source/Tech Tree/Tree Years.xlsx
WithCrow: That specific file haunts me across my entire Github desktop set-up. Nothing changed, I simply opened it & couldn't discard the change (because nothing changed, so there is nothing to discard!). 

I didn't realise it made it through here. I'll do my best to undo it

Source/Tech Tree/Parts Browser/data/ROCapsules.json
WithCrow: It's been a bit since I changed those costs, but that doesn't sound right to me either. I must have meant to type 250

WithCrow: I've increased the cost of the wings to 500, lowered the cost of the vertical stabilizer to 250 and increased the cost of the elevon to 100. That is, in my opinion, a pretty solid distribution in terms of price, as the wings will require the most effort for TPS, etc. with the stabilizers seeing the second most effort.

Source/Tech Tree/Tree Years.xlsx
WithCrow: There we go. Weird binary change is no more!

PR #2754, opened 2026-04-23
Training reminder to X-planes and Crewed Orbit Contract Descriptions

hayghostin:
Adds a little training reminder/mini guide to the x-planes and early crewed program contracts to engage new players to learn the training system. New players do not have a way to finding out out training until they go to launch a freshly built and rolled out vessel to find out zero kerbals are populating because they aren't trained. Added to both these programs descriptions text only, and reason for both programs is due to how most new players might not even touch the planes and find out in the crewed era as well as new players who do start with planes they wont have the reminder until the crewed era.


pap1723:
These look good to me and is a nice feature for new players.

Clayell:
I think a popup for the first time that a user selects a capsule/cockpit to research might be good as well

hayghostin:
> I think a popup for the first time that a user selects a capsule/cockpit to research might be good as well

imo there's already what feels like a million other popups already it would just add the existing clutter of windows that players already deal with. and the only current pop up is at launch time where it give you the warning when you've just rolled it out the pad/runway

siimav:
>  Go to the RP-1 window then to Training and begin their training.

This sounds a bit awkward. I'm sure someone can suggest a better wording.

JasperStedman:
> > Go to the RP-1 window then to Training and begin their training.
> 
> This sounds a bit awkward. I'm sure someone can suggest a better wording.

Perhaps:
Astronauts need to be trained for their missions. There are two types, both found under Training in the RP-1 window. They first need to do Proficiency Training for each part that they will fly in - this can be done once and lasts forever. They also require Mission Training to be completed before flying - this must be repeated before every flight.

panic-7700:
an attempt to simplify and emphasize:
>  Astronauts must be trained before launch. Training can be found in the RP-1 window and comes in two forms. First, complete Proficiency Training -- this is done **once per module** they will fly, and lasts forever. Once proficient, Mission Training is required before each flight, and must be repeated before **every flight**.

njits23:
That suggestion contains a minor inaccuracy. I suggest the following:

> Astronauts need to be trained for their missions. There are two types of training, Proficiency Training and Mission Training. Astronauts need Proficiency Training for each crewable part on the vehicle they launch on and for each part that they will enter during their mission. Proficiency Training has to be done only once and lasts forever. Astronauts also require Mission Training for the part you intend to launch them in. Mission training must be performed before every flight and will expire on use or if it goes unused too long. To begin training your astronauts, go to Training tab in the RP-1 window, accessible via the RP-1 icon in the toolbar.

hayghostin:
update 
"Astronauts must be trained before launch. There are two types of training, Proficiency Training and Mission Training. Astronauts need Proficiency Training for each crewed part on the vehicle they launch on and for each part that they will enter during their mission. Proficiency Training has to be done only ONCE and lasts FOREVER. Next is Mission Training for the part you intend to launch them in. Mission training must be performed before EVERY flight and will EXPIRE on use or if it goes unused too long. To begin training your astronauts, go to Training tab in the RP-1 window, accessible via the RP-1 icon in the toolbar."

a little mix of both of yalls

PR #2758, opened 2026-04-26
Add Tundra Exploration Configs

WithCrow:
This PR places Tundra Exploration's Cargo & Crew Dragon in 2012 and 2019 respectively (Modern & Commercial Capsules, respectively). The prices, etc. are a first guess and definitely need more work.

Needs:
https://github.com/KSP-RO/RealismOverhaul/pull/3137
https://github.com/KSP-RO/ROKerbalism/pull/192


WithCrow:
I have now also applied the pre-existing ECM modifier for CrewDragon, which makes my life much easier

connorjak:
I tried to add this on v4.3.0.0 and the parts don't show up in the tech tree in a career save, or the VAB in a non-career save. Is that to be expected, or is there some issue? I also manually added ELH with no issues (on the PR's latest commit).

Using Tundra Exploration v7.2.0

WithCrow:
> I tried to add this on v4.3.0.0 and the parts don't show up in the tech tree in a career save, or the VAB in a non-career save. Is that to be expected, or is there some issue? I also manually added ELH with no issues (on the PR's latest commit).
> 
> Using Tundra Exploration v7.2.0

Did you also add the [RO PR](https://github.com/KSP-RO/RealismOverhaul/pull/3137)? If yes, then I would be quite puzzled

connorjak:
Oh that was probably it; I didn't know RO would need a config as well.

WithCrow:
> Oh that was probably it; I didn't know RO would need a config as well.

I also forgot to include it in the description! I did for the Knes PR, so my fault!

PR #2763, opened 2026-04-28
Make LC min mass fraction MM-patchable

chambm:
## Summary

- Move the LC min mass fraction from hardcoded `0.75f` in `LCData.CalcMassMin` to a new `lcMassMinFraction` field on `SpaceCenterSettings`, default 0.75 (no behavior change).
- MM patches can now override it, e.g. `@SPACECENTERSETTINGS:FINAL { @lcMassMinFraction = 0.5 }`, for players who want a wider LC mass range without a recompile.

### DIFF
--- a/GameData/RP-1/SpaceCenterSettings.cfg
+++ b/GameData/RP-1/SpaceCenterSettings.cfg
@@ -272,6 +272,3 @@ SPACECENTERSETTINGS
 		XenonGas = 1.0
 	}
 }
-
-// Min vessel mass an LC accepts, as a fraction of its max tonnage. Override example:
-//@SPACECENTERSETTINGS:FINAL { @lcMassMinFraction = 0.5 }
--- a/GameData/RP-1/SpaceCenterSettings.cfg
+++ b/GameData/RP-1/SpaceCenterSettings.cfg
@@ -6,6 +6,7 @@ SPACECENTERSETTINGS
 	hangarCostForMaintenanceOffset = 240000 // these are wrt its cost
 	hangarCostForMaintenanceMin = 20000
 	lcCostMultiplier = 2
+	lcMassMinFraction = 0.75
 	salaryEngineers = 500
 	salaryResearchers = 750
 	researchersToUnlockCreditSalaryMultipliers
@@ -271,3 +272,6 @@ SPACECENTERSETTINGS
 		XenonGas = 1.0
 	}
 }
+
+// Min vessel mass an LC accepts, as a fraction of its max tonnage. Override example:
+//@SPACECENTERSETTINGS:FINAL { @lcMassMinFraction = 0.5 }
--- a/Source/RP0/Settings/SpaceCenterSettings.cs
+++ b/Source/RP0/Settings/SpaceCenterSettings.cs
@@ -31,6 +31,9 @@ namespace RP0
         public double lcCostMultiplier = 2d;
 
         [Persistent]
+        public double lcMassMinFraction = 0.75d;
+
+        [Persistent]
         public PersistentListValueType<double> nautYearlyUpkeepPerFacLevel = new PersistentListValueType<double>();
 
         [Persistent]
--- a/Source/RP0/Settings/SpaceCenterSettings.cs
+++ b/Source/RP0/Settings/SpaceCenterSettings.cs
@@ -1,6 +1,7 @@
 ﻿using ROUtils;
 using ROUtils.DataTypes;
 using System.Collections.Generic;
+using UnityEngine;
 
 namespace RP0
 {
@@ -31,7 +32,7 @@ namespace RP0
         public double lcCostMultiplier = 2d;
 
         [Persistent]
-        public double lcMassMinFraction = 0.75d;
+        public float lcMassMinFraction = 0.75f;
 
         [Persistent]
         public PersistentListValueType<double> nautYearlyUpkeepPerFacLevel = new PersistentListValueType<double>();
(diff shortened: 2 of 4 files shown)


njits23:
I am not against this being a configurable option, but in my opinion the default should remain the same for all difficulty settings.

ilveth:
For lower tonnage rockets it does seem imo a bit silly sometimes but at high tonnage the range really gets massive already. Having a different curve at lower tonnage would be nice. Something like: `minTonnage = maxTonnage * min(1/20*maxTonnage*0.75, 0.75)`. 50% at normal seems like a huge range.

chambm:
> For lower tonnage rockets it does seem imo a bit silly sometimes but at high tonnage the range really gets massive already. Having a different curve at lower tonnage would be nice. Something like: `minTonnage = maxTonnage * min(1/20*maxTonnage*0.75, 0.75)`. 50% at normal seems like a huge range.

I agree but for gameplay sake (lamenting the cruelty of the rocket equation!) I think it's reasonable to use a simple linear factor. I found the whole minimum tonnage thing very tedious and unfun so I have it set it at 10%.

connorjak:
I agree with having the tonnage window size be tweakable as a difficulty setting. 

On the motivation for the rule, I think this is supposed to represent how ground service equipment would need to be re-engineered per rocket design. So, the player is motivated to stick to a limited set of rocket designs, and have an LC dedicated to each.

One aspect that the tonnage rule does not cover well is optional rocket components, i.e. Saturn 1C vs Saturn V, Falcon 9 vs Falcon Heavy, any rocket with small optional SRBs, etc. The player ends up needing to have two launch complexes for these modular rockets, even though often IRL there is not a big difference in ground service equipment.

There are some possible alternatives to tonnage window:
- Rollout cost or time penalty to do minor reconfiguration of a large LC to support a smaller rocket (smaller in mass)
- Focus more on these possible criteria to determine rocket match to LC:
  - Outer dimension limits as they are currently
  - Have a more forgiving lower limit on mass as discussed above
  - Similarity measure of tank radii of largest tanks to LC design rocket
  - Rocket to launch has <= (with margin) the first stage surface thrust of the LC design rocket

DRVeyl:
After a quick discussion with other devs, there is no appetite in difficulty-based LC tonnage range.  The mass basis is a proxy for stage detection.  That range, broader or narrower, is not how to tweak that.

The team has no interest in returning towards a forever-improving VAB that builds all rockets.  That is opposite the direction of P&LC.

Likely to be accepted: changing the 75% threshold to be an MM-patchable value instead of hardcoded.

chambm:
I'm not asking the team to return toward a forever improving VAB (thus I had no issue returning the defaults for all difficulties to 75%). Just to give players the option to do so, preferably without resorting to something as hidden and clumsy as a MM patch. Those seem more suitable for complex alterations that don't work well as difficulty settings, or are related to the interaction of multiple mods together.

DRVeyl:
Noted.  This is simply not a "difficulty" function and RP-1 does not want to expose it that way.

chambm:
This is a horrible UX, especially considering you have to restart the game and recompile all the MM patches to try different values:
image
Why so much resistance to this as a simple slider? I don't care if it's called a difficulty option or accessibility or anti-tedium option. It definitely does make the game easier as you don't need to manage as many LCs, which is a handicap for your budget. I just don't understand wanting to force players to micromanage LCs if they don't want to.

chambm:
I figured out that I could just make another mod to surface the MM-patchable value as a slider in the options, so vanilla RP-1 options stay the same, and I get my slider that can modify it without restarting. Win-win!

chambm:
I'm not sure about the sonarcloud analysis. The other properties in that file are already public, so the new one should be public too? Or not?

siimav:
Nah, don't take it as full truth. It knows nothing about KSP dev and unfortunately the rules aren't configurable in free tier.

GameData/RP-1/SpaceCenterSettings.cfg
```diff
@@ -271,3 +272,6 @@ SPACECENTERSETTINGS
 		XenonGas = 1.0
 	}
 }
+
+// Min vessel mass an LC accepts, as a fraction of its max tonnage. Override example:
+//@SPACECENTERSETTINGS:FINAL { @lcMassMinFraction = 0.5 }
```
siimav: In general we don't keep such examples in config files so should be removed from here as well.

Source/RP0/Settings/SpaceCenterSettings.cs
```diff
@@ -30,6 +30,9 @@ public class SpaceCenterSettings : ConfigNodePersistenceBase
         [Persistent]
         public double lcCostMultiplier = 2d;
 
+        [Persistent]
+        public double lcMassMinFraction = 0.75d;
```
siimav: This should be float too if the calculations it goes into deal with floats

siimav (changes requested):

Source/RP0/Settings/SpaceCenterSettings.cs
DRVeyl: Also clamp this value to 0 <= lcMassMinFraction <= 1 in code, probably in Load() below.

chambm: OK to use UnityEngine for Clamp01?

Glloyd:
Terrible decision here to keep this obfuscated. RP-1 should aim to be more configurable in my opinion. @chambm if you do make that mod, set up a repo for it please, because I'm interested in that functionality as well.

As for the issue that led to this PR, at the very least, some scaling based on maximum mass should be considered, because the minimum LC mass limits at smaller overall mass limits are restrictive to the point of being nonsensical.

norcalplanner:
I know this is already closed, but I wanted to chime in and say how happy I am that this modification is going in.  Having the potential to edit this variable so that a player can use the same LC to launch a Delta IV and a Delta IV Heavy, or a Falcon 9 and a Falcon Heavy, or a full R-7 and a half R-7, or any of the ETS designs that feature variable numbers of SRBs or LFBs, makes me smile.  It makes me smile a lot.

PR #2764, opened 2026-04-28
Show engine config stats on RFUpgrade tooltips

chambm:
Every RFUpgrade_* in R&D currently shows the same "Engine Performance Upgrade" text, so picking a research path is confusing if you're new to the tree.

This appends stats from the matching RealFuels CONFIG (Isp vac/SL, thrust, throttle, ignitions, ullage, propellants) to the tooltip, with deltas vs the engine's default config.

## Screenshot
image

### DIFF
--- a/Source/RP0/Harmony/PartListTooltipUpgradeStats.cs
+++ b/Source/RP0/Harmony/PartListTooltipUpgradeStats.cs
@@ -1,0 +1,181 @@
+using HarmonyLib;
+using KSP.UI.Screens.Editor;
+using RealFuels;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+using UnityEngine;
+
+namespace RP0.Harmony
+{
+    [HarmonyPatch(typeof(PartListTooltip))]
+    internal class PatchPartListTooltipUpgradeStats
+    {
+        private const string UpgradePrefix = "RFUpgrade_";
+
+        private struct ConfigPair { public ConfigNode Config; public ConfigNode Baseline; }
+        private static Dictionary<string, ConfigPair> _cache;
+
+        [HarmonyPostfix]
+        [HarmonyPatch("Setup")]
+        [HarmonyPatch(new Type[] { typeof(AvailablePart), typeof(PartUpgradeHandler.Upgrade), typeof(Callback<PartListTooltip>), typeof(RenderTexture) })]
+        internal static void Postfix_Setup(PartListTooltip __instance, PartUpgradeHandler.Upgrade up)
+        {
+            if (up == null || string.IsNullOrEmpty(up.name) || !up.name.StartsWith(UpgradePrefix, StringComparison.Ordinal))
+                return;
+            if (__instance.textInfoBasic == null)
+                return;
+
+            string configName = up.name.Substring(UpgradePrefix.Length);
+            ConfigPair pair;
+            try
+            {
+                if (_cache == null) BuildCache();
+                if (!_cache.TryGetValue(configName, out pair) || pair.Config == null) return;
+            }
+            catch (Exception e)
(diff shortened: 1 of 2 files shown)


pap1723:
Awesome! I had messed around with doing something like this recently.

siimav:
Is the baseline always the first/default config? I'm wondering if it should be the previously available one instead.

chambm:
> Is the baseline always the first/default config? I'm wondering if it should be the previously available one instead.

Yes it's always the base engine that's being upgraded. It's pretty common that there is no single "previously available one" because some techs unlock multiple upgrades for the same engine. I suppose in those cases it could compare against each available previous one but that seems pretty likely to be confusing.

Source/RP0/Harmony/PartListTooltipUpgradeStats.cs
```diff
@@ -0,0 +1,205 @@
...
+        {
+            if (double.IsNaN(v)) return null;
+            return label + ": " + F(v, "0.#") + " s" + DeltaSuffix(v, baseline, "0.#", " s");
+        }
+
+        private static string BuildThrustLine(double maxT, double minT, double bMaxT)
+        {
+            if (double.IsNaN(maxT)) return null;
+            string throttle = (!double.IsNaN(minT) && minT > 0 && minT < maxT)
+                ? "  (throttles to " + F(minT / maxT * 100.0, "0") + "%)"
```
siimav: Consider "0.#" for one optional decimal, consistent with the Isp format

Source/RP0/Harmony/PartListTooltipUpgradeStats.cs
```diff
@@ -0,0 +1,205 @@
...
+            ConfigNode[] nodes = cfg.GetNodes("PROPELLANT");
+            if (nodes == null || nodes.Length == 0) return null;
+            var fuels = new List<string>();
+            foreach (ConfigNode p in nodes)
+            {
+                if (ParseBool(p.GetValue("ignoreForIsp"))) continue;
+                string n = p.GetValue("name");
+                if (!string.IsNullOrEmpty(n)) fuels.Add(n);
+            }
+            return string.Join(" / ", fuels.ToArray());
```
siimav: Redundant ToArray()

siimav (changes requested):
Something to consider:
 _cache is a static field with no reset path. It will continue to use up memory in all scenes afterwards. Additionally, the part database can be reloaded through the cheat menu (although you can probably count the mods that implement that correctly on one hand).
Might we worth to reset the cache in a GameDatabase.Instance.OnDatabaseLoaded
   callback or in a [KSPAddon].

Source/RP0/Harmony/PartListTooltipUpgradeStats.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        private static int ParseInt(string s, int fb) =>
+            !string.IsNullOrEmpty(s) && int.TryParse(s, NumberStyles.Integer, CultureInfo.InvariantCulture, out int v) ? v : fb;
+
+        private static bool ParseBool(string s) =>
+            !string.IsNullOrEmpty(s) && bool.TryParse(s, out bool b) && b;
+    }
+
+    [KSPAddon(KSPAddon.Startup.Instantly, true)]
+    internal class RFUpgradeStatsCacheReset : MonoBehaviour
```
siimav: I believe this will make KSPCF complain about destroyed instance of event handler. Better to use the `HostedSingleton` pattern like
https://github.com/KSP-RO/RP-1/blob/6ab88793fb5f79c86c5a9ee103810cbe0b1d2bf2/Source/RP0/SpaceCenter/SCMEvents.cs#L14

PR #2766, opened 2026-04-29
Add new contract parameters, refactor EEOS contracts

periodically-makes-puns:
Currently, the EEOS checks for carrying certain experiments do not actually check anything. It is possible to fulfill the conditions for the IR Follow-On contracts and Film contracts with a Sputnik core.

This PR adds two custom parameters. 
`RP1HasExperimentVesselParam` checks if a vessel has (or includes via some higher-tier experiment) the indicated experiment.
`RP1HasSamplesVesselParam` checks if a vessel has samples for the given experiment, no higher-tier replacements allowed.

All the EEOS contracts depending on this behaviour have been refactored to use it. Additionally, two localisation keys have been added to support these parameters.

### DIFF
--- a/GameData/RP-1/Contracts/Earth Observation 1/EOSCorona.cfg
+++ b/GameData/RP-1/Contracts/Earth Observation 1/EOSCorona.cfg
@@ -92,7 +92,7 @@ CONTRACT_TYPE
 			PARAMETER
 			{
 				name = HasFilm
-				type = RP1HasSamplesVesselParam
+				type = RP1HasSamples
 				experiment = RP0photos1a
 			}
 
--- a/GameData/RP-1/Contracts/Earth Observation 1/EOSCorona.cfg
+++ b/GameData/RP-1/Contracts/Earth Observation 1/EOSCorona.cfg
@@ -91,19 +91,10 @@ CONTRACT_TYPE
 	
 			PARAMETER
 			{
-				name = PartValidation
-				type = PartValidation
-				
-				VALIDATE
-				{
-					MODULE
-					{
-						name = Experiment
-						experimentID = RP0photos1a
-					}
-				}
-				title = Carry film from Basic Film Camera
-			}	
+				name = HasFilm
+				type = RP1HasSamplesVesselParam
+				experiment = RP0photos1a
+			}
 
 			PARAMETER
 			{
--- a/Source/CC_RP0/CC_RP0.csproj
+++ b/Source/CC_RP0/CC_RP0.csproj
@@ -50,6 +50,10 @@
       <HintPath>$(ReferencePath)/ContractConfigurator.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="$(KSPRoot)/GameData/Kerbalism/Kerbalism112.kbin">
+      <HintPath>$(ReferencePath)/Kerbalism112.kbin</HintPath>
+      <Private>False</Private>
+    </Reference>
     <Reference Include="$(KSPRoot)/KSP_x64_Data/Managed/System.dll">
       <HintPath>$(ReferencePath)/System.dll</HintPath>
       <Private>False</Private>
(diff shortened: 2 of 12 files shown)


GameData/RP-1/Contracts/Earth Observation 1/EOSCorona.cfg
```diff
@@ -91,19 +91,10 @@ CONTRACT_TYPE
...
-					MODULE
-					{
-						name = Experiment
-						experimentID = RP0photos1a
-					}
-				}
-				title = Carry film from Basic Film Camera
-			}	
+				name = HasFilm
+				type = RP1HasSamplesVesselParam
```
siimav: Check whether RP1HasSamples works. If not, make it work.

Source/CC_RP0/Parameter/RP1HasExperimentVesselParam.cs
```diff
@@ -0,0 +1,60 @@
...
+                return title;
+            return title = Localizer.Format("#rp0_CC_HasExperimentParam", ScienceDB.GetExperimentInfo(experiment).Title);
+        }
+
+        protected override bool VesselMeetsCondition(Vessel vessel)
+        {
+            if (vessel == null) return false;
+            foreach (Part p in vessel.Parts)
+            {
+                foreach (Experiment exp in p.FindModulesImplementing<Experiment>())
```
siimav: FindModulesImplementingReadOnly from KSPCF

Source/CC_RP0/Parameter/RP1HasExperimentVesselParam.cs
```diff
@@ -0,0 +1,60 @@
+﻿using ContractConfigurator.Parameters;
+using KERBALISM;
+using KSP.Localization;
+using System.Linq;
```
siimav: Switch to UniLinq or use a non-linq equivalent on List<T>

Source/CC_RP0/ParameterFactory/RP1HasExperimentVesselParamFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public class RP1HasExperimentVesselParamFactory : ParameterFactory
+    {
+        protected string experiment;
+
+        public override bool Load(ConfigNode configNode)
+        {
+            bool valid = base.Load(configNode);
+
+            valid &= ConfigNodeUtil.ParseValue<string>(configNode, "experiment", x => experiment = x, this, "RP0telemetry1");
```
siimav: Odd to have a default value here

siimav (changes requested):

periodically-makes-puns:
https://github.com/user-attachments/assets/ded16cd3-0611-453d-94ff-2522995439ff

image

The parameter reverts to incomplete if I detach the sample return portion without transferring samples, as expected.

periodically-makes-puns:
I forgot Kerbalism's IncludedExperiments only includes direct children. So now we check that in a recursive call.

periodically-makes-puns:
Also RP1HasExperiment still has the issue where it doesn't work if placed in a sequence?

periodically-makes-puns:
It does work if all of the HasExperiment parameters are wrapped in an All parameter, though.

Source/CC_RP0/CC_RP0.csproj
```diff
@@ -54,6 +54,10 @@
       <HintPath>$(ReferencePath)/Kerbalism112.kbin</HintPath>
       <Private>False</Private>
     </Reference>
+		<Reference Include="$(KSPRoot)/GameData/KSPCommunityFixes/Plugins/KSPCommunityFixes.dll">
+			<HintPath>$(ReferencePath)/KSPCommunityFixes.dll</HintPath>
+			<Private>False</Private>
+		</Reference>
```
siimav, line 60: Spaces vs tabs

siimav:
> Also RP1HasExperiment still has the issue where it doesn't work if placed in a sequence?

What's the issue exactly? Parameter doesn't get evaluated at all?

periodically-makes-puns:
Yeah, if there are multiple PartValidation (or RP1HasExperiment) checks adjacent to each other in a Sequence only the first properly evaluates.

siimav:
Mayhaps it needs to be wrapped in a VesselParameterGroup

periodically-makes-puns:
Oh, that's what VPGs are for. Yep, that works if all of the RP1HasExperiment VesselParameters are wrapped in the VPG.

Source/CC_RP0/CC_RP0.csproj
periodically-makes-puns: D'oh, not sure how that slipped through.

siimav:
Does this still need some changes?

periodically-makes-puns:
I don't think so. So long as Crow properly wraps their multi-experiment checks in a VesselParameterGroup it should work just fine.

PR #2768, opened 2026-05-02
Editor: Modify LC button when editing in-progress vessel

chambm:
## Summary

- New Modify LC button next to Simulate when editing an in-progress vessel
- Saves a round trip: currently you have to cancel the edit, modify the LC at the space center, and restart the edit if the vessel needs a resource the LC doesn't support

### DIFF
--- a/Source/RP0/UI/KCT/GUI_Editor.cs
+++ b/Source/RP0/UI/KCT/GUI_Editor.cs
@@ -397,6 +397,24 @@ namespace RP0
                 SpaceCenterManagement.Instance.LaunchedVessel = new VesselProject(EditorLogic.fetch.ship, EditorLogic.fetch.launchSiteName, EditorLogic.FlagURL, true);
                 SpaceCenterManagement.Instance.LaunchedVessel.LCID = editedVessel.LC.ID; // should already be correct, but just in case.
             }
+
+            LaunchComplex editLC = editedVessel.LC;
+            bool canModify = editLC.CanModifyReal;
+            const string editModifyTooltip = "Modify the launch complex to support this vessel's resources, keeping existing support where possible.";
+            const string editModifyFailTooltip = "\n\nA reconditioning, rollout, or repair is currently in progress and must finish first.";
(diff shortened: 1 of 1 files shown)


Source/RP0/UI/KCT/GUI_Editor.cs
```diff
@@ -397,6 +397,24 @@ private static void RenderEditMode()
                 SpaceCenterManagement.Instance.LaunchedVessel = new VesselProject(EditorLogic.fetch.ship, EditorLogic.fetch.launchSiteName, EditorLogic.FlagURL, true);
                 SpaceCenterManagement.Instance.LaunchedVessel.LCID = editedVessel.LC.ID; // should already be correct, but just in case.
             }
+
+            LaunchComplex editLC = editedVessel.LC;
+            bool canModify = editLC.CanModifyReal;
+            const string editModifyTooltip = "Modify the launch complex to support this vessel's resources, keeping existing support where possible.";
```
Clayell, line 403: It may be wise to change this to specifically mention that the 'vessel' being referred to is the one currently being edited.

siimav:
This looks risky at a glance. It shouldn't be possible to queue up a vessel to a LC that is being modified. So at best committing the LC modifications should also basically end the edit session.

chambm:
I'm not quite sure what you mean. It's possible to modify an LC while it's building or holding a vessel. I assume that's intended. Otherwise you'd be having to scrap vehicles you wanted to edit and build them anew after modifying. Do you mean queueing a NEW vessel while editing an existing/building one?

chambm:
I tested what happens if you click Launch after modifying the LC from the edit mode. It pops up a dialog that says "This vessel must be built in a launch complex." Saving the edits after modifying the LC closes the session as expected, but integration of the edited vehicle is paused with "(site reconstructing)".

siimav:
> It's possible to modify an LC while it's building or holding a vessel

That's... odd. Is the vessel put into queue and no progress will get incremented until the LC reconstruction finishes? Does it properly validate that both the already built vessels and the edited vessel satisfy the parameters of the new LC?

PR #2770, opened 2026-05-06
Allow picking start KSC with scenarios

siimav:
image


chambm:
Better update that green intro text. ;)

PR #2772, opened 2026-05-09
Sort skopos maintenance contracts last

periodically-makes-puns:
image
image

The Skopos maintenance contracts are now sorted after the Records contracts, which should make it easier to see non-Skopos contract parameters when there are many maintenance contracts.

### DIFF
--- a/Source/RP0/Harmony/ContractsApp.cs
+++ b/Source/RP0/Harmony/ContractsApp.cs
@@ -13,64 +13,43 @@ namespace RP0.Harmony
         [HarmonyPatch("CreateItem")]
         internal static bool Prefix_CreateItem(ContractsApp __instance, Contract contract, ref UICascadingList.CascadingListItem __result)
         {
-            if (contract is ContractConfigurator.ConfiguredContract configuredContract)
+            if (!(contract is ContractConfigurator.ConfiguredContract configuredContract))
             {
-                // Skopos goes at the end
-                if (ContractUtils.ContractIsSkoposMaintenance(configuredContract))
-                    return true;
-
-                UnityEngine.UI.Button button;
-                int index = -1;
-
-                if (ContractUtils.ContractIsRecord(configuredContract))
-                {
-                    // Find the index of the first Skopos maintenance contract.
-                    foreach (var kvp in __instance.contractList)
-                    {
-                        if (ContractSystem.Instance.GetContractByGuid(kvp.Key) is ContractConfigurator.ConfiguredContract cc)
-                        {
-                            if (ContractUtils.ContractIsSkoposMaintenance(cc))
-                            {
-                                int idx = __instance.cascadingList.ruiList.cascadingList.GetIndex(kvp.Value.header);
-                                if (idx < 0)
-                                {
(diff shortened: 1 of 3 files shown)


Source/RP0/Utilities/ContractUtils.cs
```diff
@@ -6,5 +6,9 @@ public static bool ContractIsRecord(ContractConfigurator.ConfiguredContract cc)
         {
             return cc.contractType.group.name == "Records" || cc.contractType.group.name == "HumanRecords";
         }
+        public static bool ContractIsSkoposMaintenance(ContractConfigurator.ConfiguredContract cc)
```
siimav, line 10: New line here

Source/RP0/Harmony/ContractsApp.cs
```diff
@@ -15,32 +15,56 @@ internal static bool Prefix_CreateItem(ContractsApp __instance, Contract contrac
         {
             if (contract is ContractConfigurator.ConfiguredContract configuredContract)
```
siimav: Invert it and return early to avoid extra level of indendation.

Source/RP0/Harmony/ContractsApp.cs
```diff
@@ -15,32 +15,56 @@ internal static bool Prefix_CreateItem(ContractsApp __instance, Contract contrac
...
+                                if (idx < 0)
+                                {
+                                    RP0Debug.LogError($"ContractsApp patcher: {cc.contractType.name} can't be found in UI list despite its guid existing in dictionary!");
+                                    continue;
+                                }
+                                if (index < 0 || idx < index)
+                                    index = idx;
+                            }
+                        }
+                    }
```
siimav: Most of this is duplicate of above. Figure out how to refactor it.

siimav (changes requested):

periodically-makes-puns:
done :+1:

Source/RP0/Harmony/ContractsApp.cs
```diff
@@ -13,40 +13,43 @@ internal class PatchContractsApp
...
+            // search for the contract we need to go before:
+            // if record, find Skopos
+            // otherwise, find record
+            bool isRecord = ContractUtils.ContractIsRecord(configuredContract);
+            foreach (var kvp in __instance.contractList)
+            {
+                if (ContractSystem.Instance.GetContractByGuid(kvp.Key) is ContractConfigurator.ConfiguredContract cc)
                 {
-                    if (ContractSystem.Instance.GetContractByGuid(kvp.Key) is ContractConfigurator.ConfiguredContract cc)
+                    if ((isRecord && ContractUtils.ContractIsSkoposMaintenance(cc)) || (!isRecord && ContractUtils.ContractIsRecord(cc)))
```
siimav: One last thing: combine it with the `if` above

periodically-makes-puns:
done, though I do wonder if there's a way to write this that isn't quadratic in the number of contracts

siimav:
Probably, if you want to go down that path. Perhaps keep track of the index of first record and maintenance contract in static fields? Could reset the indexes in CreateContractsList prefix.

periodically-makes-puns:
I think I won't bother for now, since the number of contracts is low enough that it frankly doesn't matter

PR #2775, opened 2026-05-09
Add highlights to cost breakdown window

periodically-makes-puns:
#2729 and #2739 for the Cost Breakdown window. Uses similar code to those two PRs. Hovering a row in the Cost Breakdown window will highlight all parts associated with that row, hovering a part will highlight and jump to the corresponding row.

https://github.com/user-attachments/assets/08111628-7dcf-4f27-a058-3eae7d7c1ec3

### DIFF
--- a/Source/RP0/UI/CostBreakdownGUI.cs
+++ b/Source/RP0/UI/CostBreakdownGUI.cs
@@ -58,6 +58,8 @@ namespace RP0.UI
         protected override void OnDestroy()
         {
             GameEvents.onEditorShipModified.Remove(ShipModifiedEvent);
+            if (_highlightRowStyle != null)
+                UnityEngine.Object.Destroy(_highlightRowStyle);
         }
 
         public UITab RenderCostBreakdownTab()
@@ -88,7 +90,7 @@ namespace RP0.UI
                 if (_hoveredPart != null && _partMap.TryGetValue(_hoveredPart, out CostEntry entry) && _costRects.TryGetValue(entry, out Rect r))
                 {
                     float rowMid = r.y + r.height / 2f;
-                    _partCostsScroll.y = Mathf.Max(0f, rowMid - 102f);
+                    _partCostsScroll.y = Mathf.Max(0f, rowMid - 150f);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -91,6 +151,30 @@ public UITab RenderCostBreakdownTab()
...
+                {
+                    _cachedHighlightedParts.AddRange(_partLookup[_hoveredEntry]);
+                    foreach (Part p in _cachedHighlightedParts)
+                    {
+                        p.SetHighlightColor(Color.yellow);
+                        p.SetHighlightType(Part.HighlightType.AlwaysOn);
+                    }
+                }
+                _hoveredEntry = newHoveredEntry;
+            }
```
siimav, line 184: At a glance it looks like this could be extracted to a method

Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -20,12 +21,35 @@ private struct CostEntry
...
         private readonly List<Part> _singlePart = new List<Part>(); // dummy list for calling GetEffectiveCost, to avoid extra allocations
         private Vector2 _partCostsScroll = new Vector2();
         private bool _isToolingTempDisabled = false;
         private IEnumerator _activeUpdate = null; 
 
+        private static CostEntry DummyEntry = new CostEntry { name = "(n/a)", effectiveCost = -1, effectiveCostModifier = 0 };
+        private CostEntry _hoveredEntry;
+        private readonly List<Part> _cachedHighlightedParts = new List<Part>();
+        private Part _hoveredPart;
+        private Texture2D _highlightRowStyle = null;
```
siimav, line 37: Destroy the texture similar tooling highlights

Source/RP0/UI/CostBreakdownGUI.cs
```diff
@@ -47,6 +71,27 @@ public UITab RenderCostBreakdownTab()
...
+            Part newHoveredPart = Mouse.HoveredPart;
+            if (newHoveredPart != null && EventSystem.current.IsPointerOverGameObject())
+                newHoveredPart = null;
+            if (newHoveredPart != _hoveredPart)
+            {
+                _hoveredPart = newHoveredPart;
+                if (_hoveredPart != null && _partMap.TryGetValue(_hoveredPart, out CostEntry entry) && _costRects.TryGetValue(entry, out Rect r))
+                {
+                    float rowMid = r.y + r.height / 2f;
+                    _partCostsScroll.y = Mathf.Max(0f, rowMid - 102f);
```
siimav: Sure this shouldn't be 150 since scroll height is 300?

siimav (changes requested):

Source/RP0/UI/CostBreakdownGUI.cs
periodically-makes-puns: Ooooh I didn't notice that was half the height. Fixed.

PR #2786, opened 2026-05-13
Add Advanced Spaceplanes Program

WithCrow:
Whereas the [Early Spaceplanes](https://github.com/KSP-RO/RP-1/pull/2749) program focusses on being a sidegrade to Gemini, with the contracts inspired by X-20 & Spiral, this program aims to capture the operations of the Space Shuttle. It is currently set up such that one is not necessarily required to use STS/Buran, but also [Callisto ](https://github.com/KSP-RO/RP-1/pull/2713). Something like [VentureStar](https://github.com/KSP-RO/RP-1/pull/2782) would also be a suitable option to complete this with. 

As Callisto/Hermes are quite smaller in terms of payload bay, this somewhat restricts the contracts that can be offered here, but I do think that most of the interesting aspects of the Shuttle can be captured nevertheless. Specifically:
- Satellite Maintenance ([STS-41-C](https://en.wikipedia.org/wiki/STS-41-C))
- Satellite Recovery  ([STS-51-A](https://en.wikipedia.org/wiki/STS-51-A))
- Satellite Deployment (Also STS-51-A)

The PR, currently, features the above three contracts in three types: 30°, 60° and 90° inclination, to allow for players from all inclinations to do this program. There are then also repeatable versions of these contracts. 

This is almost best used in combination with other PRs, such as the [Space Telescopes](https://github.com/KSP-RO/RP-1/pull/2699) or a [Modular Stations](https://github.com/KSP-RO/RP-1/pull/2605) PRs. 

Future things I would like to add:
- Spacelab esque contracts
- Possibly Magellan/Galileo inspired contracts


billy54:
What should we have for the capstone on this? There wasn't some ultimate final mission goal for Space Shuttle, it was mostly just messing around with LEO doing miscellaneous science stuff for a few decades. Possibly we can have the player do at least one of Magellan/Hubble/station module launch

WithCrow:
> What should we have for the capstone on this? There wasn't some ultimate final mission goal for Space Shuttle, it was mostly just messing around with LEO doing miscellaneous science stuff for a few decades. Possibly we can have the player do at least one of Magellan/Hubble/station module launch

I do not think it needs a specific capstone contract - my current approach is for this to mimick the long-lasting nature of the Shuttle program & simply add quite a few contracts for the player to complete, instead of one in particular.

Inspired by discussions on the discord, I've added a LDEF 'esque' contract, as well as a contract for the deployment of a satellite in just one orbit. 

Future things I want to do: 
- Spacelab/hab esque contract (series?) for repeated launches to take care of the vast amount of experiments the Shuttle completed over its life
- Launch a station module to an already existing station
- Possibly Magellan/Galileo like launches, like you already mentioned

WithCrow:
I wonder about the scope this program should take. On the one hand, it would be easy to set it to 40 years with a ton of contracts - or it could be split up into phases (i.e. Shuttle before ISS & shuttle during ISS, for example, or even before Mir, Mir, before ISS and ISS) to give the player more flexibility

WithCrow:
I've added the experiments flown on the first 10 Shuttle Missions and sketched out the next 20 - if there are any Shuttle experts that have a more detailed knowledge of the power draw, etc. than what I was able to scrounge together with the lackluster train WiFi, I would be thankful for input!

isturdy:
I'd lean toward emphasizing testing and demonstration of capabilities in a relatively short program. If players want to continue to use their spaceplane as a workhorse LV they're welcome to do it under non-spaceplane-specific contracts; the value of the dedicated program is providing an incentive and funding to cover the initial research/tooling.

WithCrow:
> I'd lean toward emphasizing testing and demonstration of capabilities in a relatively short program. If players want to continue to use their spaceplane as a workhorse LV they're welcome to do it under non-spaceplane-specific contracts; the value of the dedicated program is providing an incentive and funding to cover the initial research/tooling.

The Early Spaceplanes program already covers that - this is the "operational" part. Hence the long duration & large amount of required contracts

isturdy:
Mm, never mind--should have read the description more thoroughly.

connorjak:
> I've added the experiments flown on the first 10 Shuttle Missions and sketched out the next 20 - if there are any Shuttle experts that have a more detailed knowledge of the power draw, etc. than what I was able to scrounge together with the lackluster train WiFi, I would be thankful for input!

That's great that you're going through Shuttle missions for experiments! There will be very many. If I'm looking at this correctly they are limited to SpaceCockpit parts? Many of these experiments were flown on Shuttle only because they needed crew monitoring and Shuttle was the only crewed vehicle NASA was operating until 1998. I'm not sure they should be locked behind the SpaceCockpit science type when they could have also been done on a station part or an uncrewed satellite in a different mission profile.

A key distinguishing factor for experiments flown on Shuttle are that they could be flown, retrieved, and possibly re-flown, so it was viable to send some very expensive low-duration experiments. In RP-1 gameplay terms, the cost would be high, but be recouped upon landing. 

Also, with Shuttle it was newly viable to execute experiments that:
- Require returning the entire, large, experiment apparatus for ground analysis
- Require immediate processing once landed, so need to be soft-landed very close to the processing lab
- Were shock-sensitive, so need to be soft-landed on a runway
- Explored various hypersonic aerodynamic phenomena

So, maybe we should only have SpaceCockpit covering experiments that align with the above. Otherwise, these would either be probe science or station science (but still doable from a spaceplane).

The existing numbered science instruments (such as infrared radiometers) should also not be duplicated for their shuttle-specific versions.

WithCrow:
Thank you for the feedback!
> A key distinguishing factor for experiments flown on Shuttle are that they could be flown, retrieved, and possibly re-flown, so it was viable to send some very expensive low-duration experiments. In RP-1 gameplay terms, the cost would be high, but be recouped upon landing. 
> 
> Also, with Shuttle it was newly viable to execute experiments that:
> - Require returning the entire, large, experiment apparatus for ground analysis
> - Require immediate processing once landed, so need to be soft-landed very close to the processing lab
> - Were shock-sensitive, so need to be soft-landed on a runway
> - Explored various hypersonic aerodynamic phenomena
> 
> So, maybe we should only have SpaceCockpit covering experiments that align with the above. Otherwise, these would either be probe science or station science (but still doable from a spaceplane).

Very good points! I initially wanted to go through the first 30 missions and see which ones would be of interest - since all of the experiments lack descriptions at the moment, it was quite easy to implement the 24 (or so) experiments from the first 10 missions. 

> The existing numbered science instruments (such as infrared radiometers) should also not be duplicated for their shuttle-specific versions.

Quite correct! I have already skipped some of the experiments already in RP-1 & will cut more the more I read about these.

These experiments are, in all honesty, at the moment exclusively fluff - the player will already be buried in a mountain of science by the time they do these missions (unless they start in one of the later scenarios, like the [1978](https://github.com/KSP-RO/RP-1/pull/2743) starts), so my idea was to require a certain amount of these to be performed, serving as an incentive to fly the Shuttle frequently, also allowing the possibility of using those flights for other things. 

As the program's scope is presently decently ambitious, I am all ears for feedback!

WithCrow:
After reading up on SMIRR - I will axe it! It was, after all, just a modified version of what flew on the Mariner probes.

WithCrow:
I've added a few Spacelab experiments and also a contract for conducting Shuttle science and one for conducting Spacelab science. These cannot be completed in just one flight, as the slots on Shuttle & Spacelab parts (i.e. just Station parts) are limited. 

All of the science values are just placeholders, I don't really know what values would be appropriate for these. On the one hand, it doesn't _really_ matter that much by the time a normal player gets around to this, as it requires spaceShuttles and modularSpaceStations for Shuttle & Spacelab experiments respectively, which are both deep into the tech tree. On the other hand, a player starting in [1978](https://github.com/KSP-RO/RP-1/pull/2743) would benefit from this science, as they wouldn't have interplanetary probes yet.

WithCrow:
I've added a Station Visit (Shuttle-Mir), Station Resupply (also Shuttle-Mir) and Station Module delivery (ISS) contracts. This brings the total up to 9 required contracts:
- Satellite maintenance (1 flight)
- Satellite delivery (1 flight)
- Satellite recovery (1 flight)
- Station visit (1 flight)
- Station module delivery (1 flight)
- Spaceplane Science (~4 flights)
- Spacelab Science (~3 flights)

and the following optional repeatables:
- Satellite Maintenance
- Satellite Delivery
- Satellite Recovery
- Station Resupply
- Station Module Delivery

WithCrow:
I've also added contracts for interplanetary probes (Venus, Mars & Jupiter). At least one of these is required - I am not quite certain as to whether or not it is possible to design a probe to launch there with Hermes - and I won't be at my set-up to test for another few hours. 

If it is not feasible, then I am not sure whether to make them optional or not. Sizing for the smallest shuttle restricts missions quite a bit (can't launch Hubble/Photo 5 with Callisto), but also means that Euro players can complete the program and not just USA/USSR.

WithCrow:
I've also added an optional Keyhole-11 / Hubble contract, as that is simply too iconic to pass up on. 

@wyllowthewisp - any thoughts on integrations with [Space Telescopes](https://github.com/KSP-RO/RP-1/pull/2699)?

WithCrow:
After discussions on the discord, I've cut the duration & funding down (from 100% of the Shuttle to 25%, i.e. 6.75 million & 10 years) and moved the station related contracts to optionals, which activate when either Early or Modular stations is active.

WithCrow:
After some initial playtesting with Callisto (i.e. French Hermes), I am really quite motivated to stop sizing for it. It is too small to reliably do the satellite recovery, for example - and I have not even looked at the interplanetary probes with it. Removing it would also allow me to make the Hubble / Photo 5 contract mandatory.

WithCrow:
Descriptions for STS 1 to 10 experiments have been added. Spacelab will have to wait a bit. 

While I could add more for the later flights, considering it already has twice the amount of experiments as the early stations program, I think it is fine to leave as is.

WithCrow:
I've also balanced the science points of all experiments against the other crewed experiments, and added the descriptions to all of the Spacelab experiments. I am glad to be done - my wrist hurts!

WithCrow:
I forgot to mention it at the time - I've decided to make the Fast Deploy (i.e. deploy a satellite & land within 1 orbit) contract optional, as it is not possible for all combinations of spaceplanes & launchsites. 

Since there is no DoD that demands it, there's no reason to mandate a player *must* be able to do it

GameData/RP-1/Contracts/Advanced Spaceplanes/AdvPlane_FastDeployment.cfg
Clayell, line 26: Prestige

WithCrow: You know, when I wrote that previous comment, I thought "Hey, did I fix the prestige? Surely, right"

WithCrow:
I'm debating on renaming this to 'Operational Spaceplanes' to stick to the "Early", "Operational" & "Advanced" naming scheme I had for the earth observation successor programs.

Advanced Spaceplanes would be a successor to this & be more something like the VentureStar program, for example.

PR #2788, opened 2026-05-13
Improve crew state checks on launch

### LINKED ISSUE #2785: Possible to launch crew while training
Image

Image

If the user has a craft on the pad, with crew selected for it, but then opens SCM and trains them, it is possible to launch these 'ghosts' that never activate and don't count as crew.

siimav:
Fix #2785


WithCrow:
Can confirm it works! The bug in 2785 no longer appears

PR #2789, opened 2026-05-13
Conformal Decals cost reduction

epicduck43:
Reduces the cost of Conformal Decals parts from 1 to 0.01 to reduce the effect of decals on vessel cost and integration time.


siimav:
Changes need to be done to the part browser json files as well.

PR #2791, opened 2026-05-15
Add more Lunar Orbiter experiments

WithCrow:
The Apollo capsules performed a few experiments specifically designed for HEO & LLO, this PR aims to add those as additional fluff for players flying around the Moon.

Presently, it adds the Apollo Light Flash Emulsion Moving Emulsion Detector, which was designed after early Apollo astronauts reported flashing lights, as scientists suspected this to be due to the impact of Cosmic Rays hitting the optical nerves (later confirmed by ALFMED) and the Bistatic Radar experiment, in which the CSM's antennas were pointed at the lunar surface, with the reflections picked up by the DSN and later evaluated. 

ALFMED is dependent on Lunar Science, Bistatic Radar on improvedComms - if I could, I would make ALFMED dependent on having sent crew around the Moon before. If anyone has an idea as to how to achieve this - I am all ears!

### DIFF
--- a/GameData/RP-1/Science/Experiments/CrewScience/LunarOrbiterCapsulesExperiments.cfg
+++ b/GameData/RP-1/Science/Experiments/CrewScience/LunarOrbiterCapsulesExperiments.cfg
@@ -109,7 +109,7 @@ EXPERIMENT_DEFINITION
 		name = Experiment
 		experiment_id = RP0ALFMED
 		ec_rate = 
-		data_rate = 21.6 // 1 kb/s 
+		data_rate = 1 // 1 kb/s 
 		@data_rate /= 345600 // 4 days
 		requires = CrewMin:2
 		resources = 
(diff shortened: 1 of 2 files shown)


WithCrow:
image
Works like a charm! The durations are both wrong, as I had set the data_rate incorrectly. Fixed for both.

hayghostin:
if the science is supposed to run on the moon it should be set there in bodies allowed otherwise you have it on home body in the screenshot

WithCrow:
Wow, I really forgot to switch that over. Will fix!

WithCrow:
Has been corrected!

GameData/RP-1/Science/Experiments/CrewScience/LunarOrbiterCapsulesExperiments.cfg
```diff
@@ -37,7 +37,7 @@ EXPERIMENT_DEFINITION
 		// sample mass in tons. if undefined or 0, the experiment produce a file
 		SampleMass = 0
 		// Body restrictions, multiple lines allowed (just don't use conflicting combinations).
-		BodyAllowed = HomeBodyAndMoons
+		BodyAllowed = Moon
```
siimav: But the description says *while en route to the Moon*

hayghostin: It's not possible to detect that but it's designed to be Lunar Orbit Based.

WithCrow: Oh, that was a change I very much did not intend to do! Reverting that ASAP

WithCrow: That change has been reverted.

WithCrow:
If anyone has ideas for more, I'd be happy to add more - but I otherwise think this PR is done

PR #2792, opened 2026-05-16
Simulator: hyperbolic approach + autostage

chambm:
**AKA This is going to save you a TON OF TIME and tedium simulating interplanetary insertions.**

The greatest thing about this feature is that you don't have to trim down your vessel in the editor to easily iterate on it at the target planet. Without this, you have to either manually stage every time you simulate, or isolate the part you're iterating on and reintegrate it into the LV after you're done. Either way is a pain.

https://github.com/user-attachments/assets/b35f1fad-8944-4d2b-a632-f4bc482c8d17
(Video predates a few changes — staging is now a single step, and the field is "Insertion ΔV" rather than "Exit velocity".)

image

 ### Hyperbolic "Start in Orbit" mode

  Places the vessel on a physically correct arrival hyperbola at the target so you can test interplanetary insertions/aerobraking. Inputs: target body, origin body, departure date (the Time field), transfer time, insertion ΔV, and periapsis altitude.

  The approach plane (inclination / LAN / arg. of periapsis) is derived from the real transfer: it Lambert‑solves the origin→target transfer for the arrival v∞ and builds the minimum‑inclination plane that contains it (validated in‑game against TWP and a from‑the‑pad flight to Mars within ~0.1° inclination). Inclination/LAN can also be set manually (blank = derive).

  - Flyby plans (insertion ΔV = 0, as TWP reports) take the approach speed straight from the solver; a blank periapsis defaults to safe altitude.
  - The atmosphere clamp on periapsis is dropped here so aerobraking trajectories can be simulated.
  - Warns on screen when the geometry is degenerate (origin/target share no parent, or the transfer angle is within 2° of 180°).

 ### Load from a KAC alarm

  A "Load from KAC alarm" button fills origin, departure date, transfer time, insertion ΔV and periapsis from a matching Kerbal Alarm Clock transfer‑window alarm (matched on origin + target, departure offset shown on the button). The alarm notes are parsed language‑agnostically — values are read positionally by unit (UT:, m/s, @…km), not by label words.

###  Autostage + editor preview

  "Autostage to stage" sheds (decouples) every stage above the chosen one in a single step at sim start and deletes the resulting debris, so launch stages don't clutter the target scene. The editor previews this live: parts are highlighted red (will be shed) / blue (will remain), and occluding fairings and cargo bays are faded so the payload inside is visible.

 ### Misc

  - Lambert solver ported from the alexmoon / Transfer Window Planner method (Sun 1979), validated against Curtis Example 5.2.
  - The sim configuration window is now draggable (it's hard to take a good screenshot with the sim window insisting on being in the same place as the vessel)
  - Mode picker is a SelectionGrid


siimav:
Screenshot?

chambm:
> Screenshot?

I made a nice video but it ended up being 260MB and I didn't realize GH had a 10MB limitation. 😆 So had to trim it down.

siimav:
Perhaps find the right decoupler and trigger only that instead of going through the entire stack. Imagine if there's some solid stages on the vessel and these start burning! Could also automatically delete the vessels created from staged off parts.
Origin body is kinda unnecessary though? Sounds like it's only for selecting in which orbit to shed stages in. Use just homebody? And if the stages get autodeleted then you might as well always go straight to target orbit.

chambm:
Fair point about deleting the autostaged debris though, I just didn't know how to do that or if it was really risky (probably less risky than letting solids fly around, hah). I like the idea of just specifying a stage with a decoupler and just skip straight to that stage instead of iterating through all of them. What if there are some action groups on stage though? Maybe we need to support both approaches, or perhaps just accept we don't support vessels with that kind of complex setup. If the latter then I agree we don't need two stage teleport.

Isn't origin body necessary for calculating the hyperbolic trajectory as it would be if you really simmed the TWP plan? Setting origin body seems an easy way to get an approximately accurate hyperbolic trajectory without manually setting the insertion LAN (which TWP doesn't give you; if it did, that would be better than origin body). I think it's currently only approximate though, for a fully accurate LAN we'd need the travel time from TWP and run a Lambert solver IIUC. The insertion trajectory is important though if you want to simulate a mission that interacts with the target planet AND its moons, because you want to insert as close to the moons' plane as possible (maybe I need to add specifying insertion inclination from TWP as well?).

siimav:
So assume you have something like a LES that gets discarded during second stage burn or some RCS that gets activated. The first case I can see being tricky to detect and automatically resolve. For RCS I assume it will still stick around as a separate stage and it won't be the end of the world if the player has to tap the space key once. Although I guess this would also make the LES get yeeted.
I did take a quick glance at the code and didn't notice anything useful being done with the origin body. The only relevant value appears to be insertion dv and only way to affect that is changing the Pe of the hyperbolic orbit.

chambm:
> So assume you have something like a LES that gets discarded during second stage burn or some RCS that gets activated. The first case I can see being tricky to detect and automatically resolve. For RCS I assume it will still stick around as a separate stage and it won't be the end of the world if the player has to tap the space key once. Although I guess this would also make the LES get yeeted. I did take a quick glance at the code and didn't notice anything useful being done with the origin body. The only relevant value appears to be insertion dv and only way to affect that is changing the Pe of the hyperbolic orbit.

`ComputeApproachLAN` uses origin body (if set) to calculate LAN. As I mentioned AFAIK that's important to line up with moons the way a manual simulation would.

Orbital mechanics are quite far outside my wheelhouse, but I think that Pe is not the only thing that affects insertion dV (in theory). Maybe you mean in practice because so many insertions will be same from Earth->X because optimal transfer windows are cyclical? But there's no reason players can't use non-optimal transfer windows in case they want to get to X much faster. So if they use more dV to get there faster, their insertion dV will be higher as well, right? So you can have the same Pe but different hyperbolic trajectory because you came in at a higher speed. To be fair it's probably not a very different trajectory within the SoI of X, but if you did a flyby and kept going, it'd be very different.

Regarding LES it seems like you're arguing for not trying to skip to the last stage/decoupler?

One thing that would be good to simulate is setting the fuel amount as if the real exit burn was done at the origin body. In my designs it's pretty typical I need some amount of dV from the insertion stage to actually finish the exit burn and make the "fine tune approach" burn. Any idea how to do that?

chambm:
I just realized that insertion dV is 0 in TWP when you plan a flyby with no capture. Might have to pull in that LambertSolver after all, or just tell users they have to ask TWP for a capture. But that sometimes significantly changes the approach.

siimav:
> tell users they have to ask TWP for a capture

That certainly is what they should be doing.

For origin body and LAN - I have serious doubts it is gonna be anywhere near close to what you'd get when actually flying the mission afterwards. It is gonna depend heavily on the speed and timing of the transfer afterall. And I still think it doesn't have an effect on the final insertion dv. Perhaps discuss with @Nazfib if it's possible to expose the arrival orbital parameters in TWP instead.

Simulating fuel amounts is pretty pointless IMO since these are mostly done with a single-ignition or cryogenic stage that will need to get discarded anyway.

Staging - I'd prefer not having to wait until the game cycles through 10 stages one at a time and rather skip to the end. Even if it means I'd have to manually tap it once myself.

chambm:
Let me assuage your doubts! I changed it to use alexmoon's Lambert solver and it is able to achieve the exact same insertion inclination without the user entering it. The user provides origin body, travel time, departure time (the existing time field) and Pe, and the solver solves the rest. It matched very closely to my from-the-launchpad simulation. I also added a "Copy from KAC Alarm" button (only for hyperbolic mode, it doesn't make sense for the others, right?) and highlighting of the vessel parts that will be removed during autostaging, which is now a single step rather than manual staging.

WithCrow:
I still do not understand the advantage of having an auto-decouple feature at all, considering that players start by building their probe and only add the LV at the end

chambm:
> I still do not understand the advantage of having an auto-decouple feature at all, considering that players start by building their probe and only add the LV at the end

Ah... you've never gone to a previously used probe and wanted to apply it to a new mission without rebuilding the whole LV? Perhaps make minor but not trivial changes to the payload that definitely require simulating but don't affect the mass enough to need different dV? Also, it's optional! :)

WithCrow:
> Ah... you've never gone to a previously used probe and wanted to apply it to a new mission without rebuilding the whole LV? Perhaps make minor but not trivial changes to the payload that definitely require simulating but don't affect the mass enough to need different dV? 

Not joking, I can't think of a single time. 

> Also, it's optional! :)

That is good, yeah.

> Orbital mechanics are quite far outside my wheelhouse, but I think that Pe is not the only thing that affects insertion dV (in theory).

Insertion delta-v is dependent on the hyperbolic excess velocity (v_inf) and the position at which you perform the burn (i.e. the periapsis). 

> So if they use more dV to get there faster, their insertion dV will be higher as well, right? 

The aforementioned v_inf would be larger, and thus the dv is larger at the same position

> So you can have the same Pe but different hyperbolic trajectory because you came in at a higher speed. To be fair it's probably not a very different trajectory within the SoI of X, but if you did a flyby and kept going, it'd be very different.

The trajectory can be very different, it depends on the standard grav. param. of the body. 

> (only for hyperbolic mode, it doesn't make sense for the others, right?)
I think it would be cool to be able to immediately jump to the insertion orbit the probe will have, if you're already doing this much work for it. Screw parabolic. No one ever needs parabolic - ah, wait, you have circular and elliptical, I assume you mean those two?

IDK - even if it is optional, I think your screenshot does not provide a lot of clarity. Typically parts are illuminated red if they are deselected, if I did not know the caption, I'd assume I had incorrectly removed something.

And lastly: couldn't you just plug into MJUtils anyway? That means you could use the exact same Lambert's solver as TWP2

chambm:
Yeah it's s bad screenshot 😕. I will post a better one tonight now that I have fairings set to 15% opacity and the kept parts highlighted blue. 

Circular and elliptical are the two existing options, it's just they are currently a checkbox instead of explicit. Is hyperbolic the right name?

Is it ok to make rp1 dependent on mjutils (if it's not already, I haven't checked)?

Source/RP0/SpaceCenter/LambertSolver.cs
```diff
@@ -0,0 +1,221 @@
+using System;
+using UnityEngine;
+
+namespace RP0
+{
+    // Port of the alexmoon / Transfer Window Planner Lambert solver (Sun, F.T., "On the Minimum Time
+    // Trajectory and Multiple Solutions of Lambert's Problem", AAS 79-164, 1979). Used to orient the
+    // hyperbolic-approach sim so its arrival v∞ direction matches TWP's: stock TransferMath's solvers
+    // gave a v∞ direction ~1.5° off the value TWP and a real flown trajectory produce.
```
siimav: What's that trying to say? That this solver is more accurate than KSP's own variant?

Source/RP0/Utilities/LambertSolver.cs
```diff
@@ -0,0 +1,221 @@
...
+
+namespace RP0
+{
+    // Port of the alexmoon / Transfer Window Planner Lambert solver (Sun, F.T., "On the Minimum Time
+    // Trajectory and Multiple Solutions of Lambert's Problem", AAS 79-164, 1979). Used to orient the
+    // hyperbolic-approach sim so its arrival v∞ direction matches TWP's: stock TransferMath's solvers
+    // gave a v∞ direction ~1.5° off the value TWP and a real flown trajectory produce.
+    //
+    // Validated against Curtis "Orbital Mechanics for Engineering Students" Example 5.2.
+    public static class LambertSolver
```
siimav, line 13: Move to Utilities

Source/RP0/UI/KCT/GUI_Simulation.cs
```diff
@@ -411,5 +668,75 @@ private static double GetDefaultAltitudeForBody(CelestialBody body)
...
+                p.highlightType = Part.HighlightType.AlwaysOn;
+                p.SetHighlightColor(p.inverseStage > target ? XKCDColors.Red : XKCDColors.Blue);
+                p.SetHighlight(true, false);
+                if (p.highlighter != null) p.highlighter.SeeThroughOn();
+                _autostageHighlightedParts.Add(p);
+
+                // Fade occluding shells (fairings, cargo bays) so the highlighted payload inside is visible.
+                if (IsOccludingPart(p))
+                {
+                    p.SetOpacity(0.15f);
```
siimav, line 739: I assume this does nothing with Deferred rendering? If so then there's IMO no point even bothering with the fade.

Source/RP0/SpaceCenter/SpaceCenterManagement.cs
```diff
@@ -1288,26 +1288,279 @@ private static IEnumerator SetSimOrbit(SimulationParams simParams)
...
-                FlightGlobals.fetch.SetShipOrbit(body.flightGlobalsIndex, ecc, sma, simParams.SimInclination, simParams.SimLAN, simParams.SimMNA, simParams.SimArgPe, 0.0); // selBodyIndex, ecc, sma, inc, LAN, mna, argPe, ObT
-                FloatingOrigin.ResetTerrainShaderOffset();
+                RP0Debug.Log($"Moving vessel to orbit. {body.bodyName}:{simParams.SimOrbitPe}/{simParams.SimOrbitAp}:{inc}");
+                FlightGlobals.fetch.SetShipOrbit(body.flightGlobalsIndex, ecc, sma, inc, lan, simParams.SimMNA, simParams.SimArgPe, 0.0);
+            }
+            FloatingOrigin.ResetTerrainShaderOffset();
+
+            if (simParams.SimAutostageTarget >= 0)
+            {
+                yield return new WaitForSeconds(1f);  // let the teleported vessel settle before shedding stages
```
siimav: The better check here is vessel being off rails

Source/RP0/SpaceCenter/SpaceCenterManagement.cs
```diff
@@ -1288,26 +1288,279 @@ private static IEnumerator SetSimOrbit(SimulationParams simParams)
...
+            Vessel active = FlightGlobals.ActiveVessel;
+            var preStaging = new HashSet<Guid>(FlightGlobals.Vessels.Select(v => v.id));
+            int triggered = 0;
+            foreach (Part p in active.parts.ToList())
+            {
+                if (p.inverseStage <= autostageTarget) continue;
+                foreach (PartModule m in p.Modules)
+                {
+                    if (m is ModuleDecouple md && !md.isDecoupled) { md.Decouple(); triggered++; }
+                    else if (m is ModuleAnchoredDecoupler mad && !mad.isDecoupled) { mad.Decouple(); triggered++; }
```
siimav: Both inherit from ModuleDecouplerBase

siimav (changes requested):
Also has conflicts now.

Source/RP0/SpaceCenter/LambertSolver.cs
chambm: I tried the built-in ksp solver first and the inclinations I was getting were off by a few percent from what I got with the twp solver (which more closely matches my manual simulation from the launch pad).

Source/RP0/UI/KCT/GUI_Simulation.cs
chambm: I don't understand. The fade works to let you see inside fairings. I wish I could just highlight parts through solid fairings but I couldn't get that to work.

Clayell: See [here](https://github.com/KSP-RO/ProceduralFairings/issues/59) for context

chambm: Thanks for the link. I don't think that bug affects this fade. I don't think I have HDR turned on (currently on a trip so I can't check) and the fade is working. But it's not a dithered fade, the whole part gets reduced opacity.

siimav: Hmm, I guess it's fine if TUFX HDR can make the fade work. We just have to include that mod in Express and ship some neutral profiles with HDR enabled.

Clayell: There's also that IVA bug with deferred + HDR on.

siimav: Tough luck for those 3 people who use IVAs.

PR #2794, opened 2026-05-17
Activate Jupiter Moon Landing Program

WithCrow:
This PR activates the existing Jupiter Moon Landing program and freshens it up for the modern-day state of RP-1. I have added descriptions and made the landing on Europa mandatory. 

As these contracts are already in the game & the program was simply disabled, I think this is a relatively smooth process. The contracts themselves are simple in their definition but incredibly interesting in how varied the possible execution paths are for a player.


WithCrow:

WithCrow:
I've added several optionals - to give this program some more meat! This now also has repeat landings, as well as sample returns. I've given them 2.5x the value of Mars Sample Return in terms of reputation/confidence gained.

PR #2795, opened 2026-05-17
Activate Saturn Moon Landing Program

WithCrow:
This PR activates the Saturn Moon Landing program and freshens it up for the modern-day state of RP-1. I have added descriptions to all of the contracts & the program itself, but other than that, no changes were made. 

Like with 2794, I am fully confident in this program: all of the files are in the game already and the program itself is simply deactivated. Like the Pluto Landings program, this is very much an 'end game' program.


WithCrow:

Glloyd:
Working through this one right now. I imagine this applies to the other programs in this series, but I do wonder what utility there is in requiring a new vessel for these landings. Given that the orbit contracts are in an entirely separate program, keeping that requirement doesn't stop you from completing multiple contracts at the same time, and only restricts the path you might take in completing these contracts. imo, for a lategame program like this that is only focused on landings (especially landings within the same planetary system), you shouldn't require new vessels, akin to some of the rover contracts.

WithCrow:
I thought so too when I did Pluto Landings, but I have come around on it. 

From a budget perspective: Why would you get paid for this if all of the probes have already been sent there? 

From a scientific perspective: You would design these landers only after the orbiters have performed their mission, so that you can calibrate your suite of experiments to the local environment.

Glloyd:
I agree that makes sense for Pluto, but I disagree in the case of Saturn/Jupiter (and maybe the other gas giants, but primarily the ones with orbiter programs) from a gameplay perspective. Mainly, the scenario I have here is a Saturn orbiter, the third that I sent during my Saturn observation program. I made use of multiple gravity assists from the other moons to get a relatively cheap orbit of Iapetus, leaving me with just enough D/V to possibly land there. My hope in taking Saturn Moon Landings was that I could then use that orbiter to attempt to complete the landing contract, which would be a longshot, but a very fun thing to attempt to try if there was a reward for doing so. Obviously, it isn't a new vessel, which from a gameplay perspective, is just a very frustrating limitation because which late game contracts require a new vessel and which don't feels very arbitrary.

I think this applies even moreso to something like Uranus, where it is very very easy to land on the moons, and could realistically be done with an orbiter sent as part of the Outer Gas Giant Survey program. We do even have an example of this in real life - NEAR Shoemaker made a landing on Eros that was not part of the original mission plan, and they even reprogrammed some of the onboard scientific instruments to make doing so more scientifically useful. From a gameplay perspective, they picked up the contract to orbit Eros, and then, after completing that contract, picked up the contract to land on there, when they realized it was possible to do so with the fuel in the maneuvering thrusters. Such a scenario is impossible here because of this somewhat arbitrary restriction (and imo this restriction should be removed from Phobos/Deimos landings for the same reason). I also think the fact that these programs don't explicitly require any specific scientific instrumentation somewhat undermines your argument from a scientific perspective. I could do some of the Uranus landings with an airbag and a bare sci core, and it would count for the contract. If these contracts required specific science to be done, then I think I would agree more with the restriction, but as is, it feels bad to play, speaking as someone playing through them right now.

Edit: To be clear, I do think that the new vessel restriction _should_ be in place for any optional landing contracts. I just think it's more interesting from a gameplay perspective to allow some extra freedom with the required landing contracts.

WithCrow:
> I also think the fact that these programs don't explicitly require any specific scientific instrumentation somewhat undermines your argument from a scientific perspective.

That is actually what I see as an arguement for my position: the ability to design a lander after you have received data from an orbiter is pretty valuable. But, if a player designs a super large mother bus of a saturn orbiter ages in advances that can land on three moons - maybe they should indeed be rewarded.

I'll remove the new vessel requirement. Consider me convinced!

WithCrow:
Just like in the Jupiter program, I've added several new optionals. In addition to the repeat landings & sample returns, this also features a dedicated Titan Lake mission. 

Due to the complexity of a Titan sample return, I gave it an immense reputation & confidence reward. Might need to be nerfed, but I honestly don't think anyone will ever do this, so might as well tempt them...

GameData/RP-1/Contracts/Saturnine Moon Landings/TitanLakes.cfg
```diff
@@ -0,0 +1,91 @@
...
+	group = SaturnMoonLandings
+
+
+	description = <b>Program: Saturnine Moon Landings<br>Type: <color=green>Required</color></b><br><br>Now that we have landed something on Titan, our scientists are desperate for a dedicated mission to one of its lakes or oceans. It is the only body where these can be readily accessed without needing to drill through an ice sheet first, so this would be a massive boon for our scientists!
+	genericDescription = Achieve a soft landing on Titan and transmit science.
+
+	synopsis = Achieve a soft landing on Titan and transmit science
+
+	completedMessage = Mission Success! You have safely landed on Titan and returned valuable science data from the surface.  
+
```
gabo463: for generic description, synopis and completed message  i find that it would be a good idea to specify  the splashdown instead of the landing

gabo463:
i would like to suggest for titans lake contract  that  the genericDescription  synopsis and complete message  of the contract be changed to put the incline on the splashdown part of the contract

WithCrow:
Strange, could have sworn I had changed it to "on Titan's lakes"

PR #2797, opened 2026-05-17
Activate Neptune Moon Landing Program

WithCrow:
This PR activates the Neptune Moon Landing program already in the game. I've added descriptions to the program and its contract.


Clayell:
Same comment as https://github.com/KSP-RO/RP-1/pull/2796#issuecomment-4472183257.

periodically-makes-puns:
I feel like this should have a longer duration. 12 years requires a rather fast transfer to Neptune, and that's assuming you launch immediately upon accepting the program. Less slots would probably also make sense?

WithCrow:
I like the thought of reducing the slot count.
@Clayell: do you think a similar thing for Uranus would be appropriate? I.e. long duration, reduce cost a bit (let's say 10 million?) 

My dev install lacks TWP and I don't have the time to spin up my main one - what's the ToF of a probe to Neptune? 40 years? Uranus was ~20 years for my RiS atmospheric probe. It is also important to note that a player won't even start this program before ~2004 (assuming standard transfers), [this website](https://transfercalculator.com/calculator/lowRes.html?pid=234) says ~40 years for a transfer from 2000!

Clayell:
Seems fine for Neptune. Uranus seems difficult, since it's multiple probes. I suppose you would need to send like 5 all in a group at once?

WithCrow:
That is what I did for Pluto Landings - so I think that makes sense. I am actually quite content with making them proper long term programs. But I wonder how many slots it should have? 2 or 3? Neptune is fine being just two (or even one?), but I am unsure about Neptune. 

I like the thought that a player could take Jupiter, Saturn, Uranus, Neptune and Pluto Landings at the same time now.

periodically-makes-puns:
Reasonable direct transfers are 15-33 years of flight time, with windows every year. Unsure about gravity-assist transfers, those might be faster but be less common.

WithCrow:
I've added a required contract to map Triton (VI5 & IR4) & also made the Neptune Atmospheric Probe from OGGS a required contract.

WithCrow:
Thans to @hayghostin for the program picture!

WithCrow:
Will anyone ever do a Triton sample return? Who knows! But they certainly can now with these optionals.

PR #2605, opened 2025-08-06
Modular Space Station Program

fugufish123:
A new program to add funding for making stations like Mir and ISS in earth orbit

PR #2798, opened 2026-05-19
Operational Earth Space Stations Programs, an alternative to Modular Space Stations program

connorjak:
A set of programs to represent all Low-Earth Orbit crewed science post-Early Earth Space Stations. The 4 programs very vaguely line up with:
- US: Shuttle/SpaceLab, Shuttle/SpaceLab, Shuttle-Mir/ISS, ISS
- USSR: Salyut 6+, Mir, Mir/ISS, ISS

The key differences between this and #2605, as well as other similar WIP programs:
- Focus on completing science objectives rather than generic Multi-Year Habitation
- No stipulation for the exact mission profile to accomplish the program.
  - Flexibility to work on either free-flyer vehicles like Shuttle/SpaceLab or modular stations.
  - Compatible with possible future changes to have old stations degrade.
  - If there is a program for Shuttle, that program could handle the capability demonstration and test flights, then these programs can handle operational usage of that infrastructure to accomplish the research.
- Many programs with flat funding, about 8 years each, which would allow for shifts in program priorities over many years.
  - Can, for example, have a break to free up slots for lunar habitation or crewed mars.

This PR doesn't have any details on added experiments. I figure we can reuse the experiments from:
- #2605 
  - And more recent fork https://github.com/panic-7700/RP-1/tree/modular-station
- #2786 

When experiment replacement from visiting vehicles becomes more relevant around 3 or 4, may need to revert to using ResearchPayload for a few generic experiment categories like "Life Sciences Experiments", "Crystal Growth Experiments", "Space Hardware Testing", etc. Could also introduce a game mechanic where parts can exchange experiments.

### DIFF
--- a/GameData/RP-1/Programs/Programs.cfg
+++ b/GameData/RP-1/Programs/Programs.cfg
@@ -926,6 +926,154 @@ RP0_PROGRAM_MODIFIER
 	}
 }
 
+{
+	name = OperationalEarthSpaceStation1 // Typically solved with something like later Salyut or SpaceLab
+	isHSF = true
+	title = Operational Earth Space Stations 1
+	description = The first space station missions have been accomplished. The possibilities for living and working in space seem endless. Now, it is time to make a sustainable architecture for long term low Earth orbit human spaceflight science. During this program, launch and operate crewed vehicles or stations that are capable of a variety of long-term science experiments.
+	requirementsPrettyText = Have completed the Early Earth Space Stations program.
+	objectivesPrettyText = Complete many station science experiments.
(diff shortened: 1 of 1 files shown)


connorjak:
image
image

Reference planning gantt chart showing a post-Crewed-Lunar possible fast attempt at completing these 4 programs, compared to real-life station programs and Shuttle. I figure the typical player may choose to space out the programs more to complete other programs in between.

connorjak:
If 3/4 should indeed allow the player to construct the ISS, funding would probably have to expand drastically.

PR #2800, opened 2026-05-21
Mission Modules as 'free flyers' for very limited station science

panic-7700:
This change was discussed at length on Discord yesterday, with some interest and consensus, so here's my first pass at it:

- **Adds** a single Prototype Station science experiment slot to Apollo D-2 Block II Mission Module and TKS Service Module (Mature Capsules; same era as Station Prototypes) 
- **Adds** a single Early Station science experiment slot to Apollo Block IV Mission Module and the service modules for TKS and Big Gemini (Improved Capsules; same era as Early Stations)
- **Removes** the Lab capability from Apollo Block IV Mission Module, although as it's located in Advanced Capsules perhaps there is an argument for it

There is interest (on the discord) in the creation of future station programs that would be more permissive in *how* they are accomplished. A great deal of research starting at least in the 60s went into so-called "free flyers", small modules that could be operated for shorter periods of scientific study than a full size space station. Their size would be both an advantage, allowing the use of smaller and cheaper launchers, and a disadvantage, limiting their capabilities significantly. The speculative Apollo Mission Modules we have in RP-1 now, based on the ETS alt history, are directly derived from Apollo Applications Program proposals that preceded the cut-back Skylab program.

### DIFF
--- a/GameData/RP-1/Science/Configure.cfg
+++ b/GameData/RP-1/Science/Configure.cfg
@@ -1167,6 +1167,15 @@
 	}
 }
 
+//Prototype station science - limited Mission Module support
+@PART[ROC-D2MissionModule2|rn_tks]:AFTER[zzzKerbalism]
+{
+	@MODULE[Configure]:HAS[#title[Crew?Science]]
+	{
+		@slots = 1
+	}
+}
+
 //Early Space Stations
 @PART[*]:HAS[#capsuleTier[EarlyStation]]:AFTER[zzzKerbalism]
 {
@@ -1381,6 +1390,15 @@
 			  id_value = RP0LunarOrbitHabit
 			}
 		}
+	}
+}
+
+//Early station science - limited Mission Module support
+@PART[ROC-ApolloMissionModule|ROC-ApolloMissionModule4|ROC-BigGeminiSM]:AFTER[zzzKerbalism]
+{
+	@MODULE[Configure]:HAS[#title[Crew?Science]]
+	{
+		@slots = 1
 	}
 }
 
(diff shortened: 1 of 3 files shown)


WithCrow:
I would like to see Early station science also added to TKS if the player has researched it, that's my only present thought on the matter - I like it!

panic-7700:
What might it take to get this reviewed, and either merged or declined?

WithCrow:
I'm surprised to see siimav didn't auto-merge this, it makes their favourite capsule even better!

Jokes aside: I'm no maintainer, but it looks good to me. I see no conflict with how the system at large works and this aligns with ELH's philosophy of simply stating objectives & letting the player figure out how to deal with it.

siimav:
But this adds "Multi-Month Space Habitation" to the capsules as well? I'm fine with the shorter experiments getting added to orbital modules but certainly not this one.

njits23:
Not only that, it reduces the slot count on the 2nd tier D2 Mission Module to 1, rather than the four it has now:
https://github.com/KSP-RO/RP-1/blob/a426a25cb2c0030291b493b4d5799e1ae452ccc0/GameData/RP-1/Science/Experiments/CrewScience/MatureCapsuleExperiments.cfg#L4-L6
https://github.com/KSP-RO/RP-1/blob/a426a25cb2c0030291b493b4d5799e1ae452ccc0/GameData/RP-1/Science/Configure.cfg#L647-L666

GameData/RP-1/Science/Configure.cfg
```diff
@@ -1167,6 +1167,15 @@
 	}
 }
 
+//Prototype station science - limited Mission Module support
```
njits23, line 1170: In https://github.com/KSP-RO/RP-1/pull/2800#issuecomment-5386301917 I'm specifically talking about this patch. This is overriding the four slots it normally has.

njits23:
If you really want free-flyer science to be a thing, you should add an extra kind of capsule tier and assign only a subset of experiments to it. Not any of the multi-week or multi-month science, but the historical experiments.

PR #2803, opened 2026-05-22
Make "Atmospheric Analysis Satellite" Contract Non-Random

Hyperion-21:
For some reason, the contract currently randomizes its Pe and Ap requirements between 200km-400km and 600km-700km respectively. This commit makes it a fixed 200-600km requirement. This is not communicated to the player in any way, is easy to cheese, and seems to serve no practical purpose, so removing the randomness seems like the best call.


jimmymcgoochie:
I would go with the middle of the requirements, so 300x650km, just to make it a little bit more of a challenge.

Clayell:
I think a big intention of this contract is that you should be able to do a perigee raise, so maybe even 400km perigee might be necessary.

siimav:
200km Pe would just decay too fast

PR #2806, opened 2026-05-23
Make stratospheric research an optional contract

panic-7700:
I understand that this might be an unpopular suggestion, but here goes anyway.

This change makes Stratospheric Research optional, for a couple of reasons:
1. players aren't able to apply what they've learned in previous contracts to low-speed, high-altitude designs
  - these temu U-2s are basically powered gliders, and they have entirely different design considerations 
2. players *also* aren't able to apply what they learn in *this* contract to any future aspects of RP-1
  - the handling of a subsonic stratospheric jet is *very* different from the handling of a hypersonic spaceplane
  - there are no subsequent contracts (or even situations) in which a high-aspect-ratio, high-altitude jet is useful or required

The amount of clock time spent designing and testing, and then flying the contract itself, can be significant, for no future payoff, and while it seems appropriate as a stretch goal for an X-planes program, it does not in my opinion fit into the rest of the current program. 

I could envision a future series of x-planes contracts *in a new program* that might take more advantage of stratospheric data gathering, but the time investment into something like that would need to be justified by significant science or reputation/confidence gains.

### DIFF
--- a/GameData/RP-1/Contracts/X-Plane Research/XPlanesStratosphericResearch.cfg
+++ b/GameData/RP-1/Contracts/X-Plane Research/XPlanesStratosphericResearch.cfg
@@ -5,7 +5,7 @@ CONTRACT_TYPE
 
 	title = X-Planes (Stratospheric Research)
 
-	description = <b>Program: X-Plane Research<br>Type: <color=green>Required</color></b><br><br>Future airliners may fly at unprecedented altitudes, and it is important to understand atmospheric behavior such as turbulence at high altitudes. In order to do this, design, build, and fly a crewed jet aircraft to maintain @VesselGroup/HoldSituation/minAltitude meters in level subsonic flight, then return home safely. Optionally, landing back at the runway will award extra reputation. If playing from the Cape, the Space Shuttle runway provided by the RSS-CanaveralHD mod is also a valid landing target.
+	description = <b>Program: X-Plane Research<br>Type: <color=blue>Optional</color></b><br><br>Future airliners may fly at unprecedented altitudes, and it is important to understand atmospheric behavior such as turbulence at high altitudes. In order to do this, design, build, and fly a crewed jet aircraft to maintain @VesselGroup/HoldSituation/minAltitude meters in level subsonic flight, then return home safely. Optionally, landing back at the runway will award extra reputation. If playing from the Cape, the Space Shuttle runway provided by the RSS-CanaveralHD mod is also a valid landing target.
 
 	genericDescription = Design, build, and fly a crewed jet aircraft to maintain a specific altitude, then return home safely.
 
--- a/GameData/RP-1/Programs/Programs.cfg
+++ b/GameData/RP-1/Programs/Programs.cfg
@@ -21,7 +21,6 @@ RP0_PROGRAM
 		}
 		complete_contract = BreakSoundBarrier
 		complete_contract = XPlanesSupersonicMach2
-		complete_contract = XPlanesStratosphericResearch
 		complete_contract = XPlanesHighAltitude
 		complete_contract = XPlanesHypersonic
 		complete_contract = XPlanesKarman
@@ -31,6 +30,7 @@ RP0_PROGRAM
 	{
 		XPlanesSupersonicOptionalLow = true
 		XPlanesSupersonicOptionalHigh = true
+		XPlanesStratosphericResearch = true
 		XPlanesStratosphericResearchOptional = true
 		RocketPlaneDevelopmentOptional = true
 		XPlanesHighAltitudeOptional = true


WithCrow:
I do not agree, necessarily. It is pretty easy to complete, whereas my main grievance is the 2 minute hypersonic contract, which is uniquely different in all of its design considerations and far more difficult relative to what you learn from it. It is a similar dead-end design line and would be far better suited for an advanced program.

I have had the idea of reviving Luci's Advanced X-Planes program bouncing around the back of my mind

panic-7700:
I do agree with the very long duration high speed requirements as well.

The main driver for this is that x-planes is often presented as The Way You Start RP-1, but the requirements for stratospheric and the hypersonic contract you mention are a great way to lose new players, or get them stuck in a time sink that potentially causes them to miss the program deadline.

Making early x-planes behave fully as an extended tutorial for spaceplane design (eventually) seems like a desirable structure for it. Unless the default start date scoots back to 1948, the earliest of early x-plane contracts we have are already far behind the historical timeline anyway. The program is currently representing an abstraction and snapshots of the real-world programs, and as long as the outliers are optional rather than required that seems okay.

billy54:
As someone who recently tried doing planes after having never touched them since legacy, I found that the hypersonic contract (at least the first one) can be done by adding drop tanks to my Karman line plane, while the stratospheric contract was a one off where I did not bother trying to optimise past the requirements of the first contract, due to the intense tedium and nonexistent science gain. I do agree with splitting X-planes in half, as this would allow players to have a small taste of RP-1 planes without being forced into the complicated stuff mentioned above. 

My suggestion: stratoplane could be kept as a requirement since it is not too hard, but it feels weird that we are essentially encouraging new players to make a one off design that will literally only be used for the first contract (or 3 if you are a masochist). Hypersonic should either be made easier (drop speed requirement to 1500 in line with the experiment/reduce or remove the 2 minute hold requirement) or moved to a new Advanced X-Planes program. Then there is more room to add fun stuff in that, like some cooler spy plane contracts (Reach this/these waypoint(s) and then fly back to KSC) but idk if that would fit the non-military principle of RP-1.

jimmymcgoochie:
The hypersonic contract is hardly difficult, the same plane that does the Karman line contract can do the hypersonic one too and you can even do it with the X-2 cockpit (I did that yesterday with only moderate remodelling of an existing rocketplane design). There isn't much difference between "go fast and very high" and "go fast and less high" and the only real challenge is overheating at lower altitudes, which the X-2 cockpit can still handle and the X-15 definitely can.
Stratospheric jet is a totally different challenge that takes a very long time, gives no science and contributes nothing to the overall progression of the program as a whole, while requiring a design that actively rejects all the lessons learned from fast jets and rocketplanes in favour of MASSIVE WING. I don't understand why this contract was ever added to the X-planes program in the first place- maybe someone really liked the U-2 and decided to make everyone else suffer for it?- but it adds nothing to the program except being a time-consuming bottleneck.

tenehea:
X-planes is already not really required for players that are not very interested in aircraft. I think both stratospheric 22km and hypersonic are uniquely interesting and these players that specifically opt in to designing aircraft should be expected to complete them

chambm:
I agree it should be optional. Rewards should also be buffed before first satellites are sent to orbit (valuable photography). After that, lower rewards. Alternately or perhaps additionally we could allow collecting photo data on Earth for "weather" purposes in exchange for money from mysterious sponsor organizations.

Jovus:
As a thought experiment, imagine attempting to complete this contract without an autopilot, which is going to be the standard bias of people coming over from 'stock'.

It's hellish.

I agree it should be optional for all the reasons given as well.

Clayell:
I don't think that's a great point. The expectation for RP-1 is that almost all missions are driven with some sort of auto-pilot or auto-guidance. Not using these tools essentially purpose-built for RP-1 is an intentional handicap.

panic-7700:
My main position with this change is not necessarily whether or not the stratospheric contract is interesting or not, it's whether or not it provides any utility to the player in either knowledge or rewards. 

As it stands right now, it serves no purpose for progressing player skills in design or flight. It also doesn't come with any science rewards, confidence, or meaningful reputation. The latter could be fixed, but the former seems unlikely to change. It has about as much relevance to the rest of the RP-1 contracts/programs (and gameplay) as a land-speed record would. **A player skipping subsonic stratospheric contracts will not find themselves missing critical skills or concepts later in the game.**

Making it optional, possibly with better rewards (any ideas?) is a best of both worlds solution in my opinion -- players who find it rewarding or who already have designs for it can just continue on as always. Players who find the _rest_ of x-planes rewarding can opt out. This seems like it hurts nobody.

WithCrow:
> A player skipping subsonic stratospheric contracts will not find themselves missing critical skills or concepts later in the game.

The same can be said for the rest of the x-planes program (currently). If 2749 and 2786 (early & advanced spaceplanes) make it into the game, then the only thing the player can transfer is, at most, how to land on the runway - something x-planes does not require. 

I love the U-2, so I am biased: the solution to this, in my opinion, is not to make it optional from this program, but to add a later advanced x-planes program, where the player can apply these skills.

panic-7700:
I disagree - learning about how to construct a stable aircraft, especially in the glide phase, is very relevant to spaceplanes, as is the heating aspect of both hypersonic and the karman line reentry. 

Although I absolutely agree that landing under control, ideally on a runway, should be required to successfully complete *all* x-planes contracts 😅

WithCrow:
> I disagree - learning about how to construct a stable aircraft, especially in the glide phase, is very relevant to spaceplanes, as is the heating aspect of both hypersonic and the karman line reentry.
 
I misspoke before with 'landing on the runway', I did mean the landing approach (including the terminal glide phase). But orbital re-entry is quite different than that of a hypersonic / karman line craft. 
 
 > Although I absolutely agree that landing under control, ideally on a runway, should be required to successfully complete all x-planes contracts 😅
 
 The X-1 touched down on a lakebed, so I disagree with making it a requirement. For the jet contracts, that's a different matter, but as njd pointed out a few days ago, not all players can do so. I do not think they need to be forced to landing on it.

panic-7700:
Yeah, lakebeds are why I said "under control" as opposed to "under a parachute" 😅 -- so basically just the "under 10° impact angle" or whatever is appropriate for the given type of aircraft.

WithCrow:
Pardon me - I must not have had enough coffees yet - I missed the 'ideally' there. I agree with that

panic-7700:
Regarding the advanced x-planes aspect, I am of the opinion that if there's a desire for required stratospheric content, it should go into its own program, maybe as a restructuring of the entire x-planes program, with programs covering a smaller spread of tech nodes. 

- Make most of the current x-planes into something like "early high speed flight"
- Add a "high altitude flight research" program 
- move X-15 contracts (hypersonic, Kerman) into their own program, possibly 
- develop an "advanced x-planes" program that does further hypersonic research and other things relevant to reentry and or very high speed flight 

Unfortunately (in my opinion) there really probably isn't enough content to spread around. The most interesting historical x-planes (imo) are the ones that pushed theoretical limits, like lifting bodies and variable-geometry wings, and those are if nothing else probably going to be *really difficult* to write contracts for. How do you detect a lifting body? Can we pull FAR data for CC checks?

Anyway. I still don't know that keeping u-2 content around as required for x-planes as opposed to optional has any benefits (aside from keeping the folks in `#planes-help` busy? 🫠)

billy54:
>  The most interesting historical x-planes (imo) are the ones that pushed theoretical limits, like lifting bodies and variable-geometry wings, and those are if nothing else probably going to be really difficult to write contracts for.

Not to mention these things are hard if not impossible to make in RP-1. I suppose if we added some hinge parts we could force the player to have a robotic hinge on board the craft for a variable geometry wing, but lifting body is literally just stacking and offsetting proc wings all over the place until it looks right.

PR #2808, opened 2026-05-25
Add Interstellar Probe Program

WithCrow:
This adds a super late-game program, which takes up all admin slots and requires all other uncrewed programs to have been completed. As there are no other solar systems, it requires a speed at 40 AU from the Sun of 60 km/s. This is larger than the 2021 JHUAPL proposal, which comes out to around 35 km/s at 40 AU, but less than the proposals which perform a burn at perihelion and achieve an escape velocity of >90 km/s at 40 AU. 

Currently, the program requires one probe to achieve this, but the program also pays less per slot per year than Crewed Mars.
image


hayghostin:
Should require it as NeptuneMoonExploration not as Landings via https://github.com/KSP-RO/RP-1/pull/2797

WithCrow:
I don't like making PRs dependent on each other, so I used the current program name for it. If either one is merged, the other would adapt

chambm:
This is a good program idea, but I don't see why this should take up all admin slots. Aren't the Voyagers interstellar probes? Isn't it more deltav to go close to the sun that it is to escape it? (counterpoint late game goal: sending trash into the sun efficiently 🤣 )

WithCrow:
> This is a good program idea, but I don't see why this should take up all admin slots. 

The program requires all others to be done, so this is a "Hey, you did everything - go wild!" program to fund whatever super-late game shenanigans the player wishes to.

> Aren't the Voyagers interstellar probes? Isn't it more deltav to go close to the sun that it is to escape it? 

Voyager 1 & 2 had a speed of 15-17 km/s at Pluto, this program requires a speed four times that. The speed could be cranked up even further: 95 km/s is possible using a Jupiter assist to lower perihelion with a 15 km/s burn at perihelion.

> (counterpoint late game goal: sending trash into the sun efficiently 🤣 )

Deal with Trash Program🤣

billy54:
So is the intent of this mission to explore interstellar space? Like even if we had other solar systems, that is only a tiny fraction of c and it would take tens of thousands of years for the probe to get anywhere close to amother star system. There is no actual reason for this to he gated behind Pluto Landing or whatever if it is just Voyager on steroids.

WithCrow:
> So is the intent of this mission to explore interstellar space? Like even if we had other solar systems, that is only a tiny fraction of c and it would take millions of years for the probe to get anywhere close to amother star system. 

I had indeed misunderstood the Hopkins proposal - what would be a reasonable fraction of c, then? 

> There is no actual reason for this to he gated behind Pluto Landing or whatever if it is just Voyager on steroids.

It shouldn't be just Voyager on steroids but even if it were, the reason for requiring all programs & all slots is that it is simply the end. The player has done all programs that can be done and if they choose to continue playing, they can do so with some amount of funding.

Considering there is exactly one logged completion of Pluto Landings on the career log, I do not believe this program will be reached by many. It is, however, a reason to do all programs and to then get a constant funding for decades.

WithCrow:
It would appear the answer is somewhere a bit above 4% of c (13,000 km/s) at 40 AU. While this is far higher, the long program duration & large funding means a player can build up a vessel capable of achieving such speeds over time (& with super late game tech). I think this is fine.

WithCrow:
So, I've changed the structure a little: the player must first send a probe into the interstellar medium (i.e. the old 60 km/s contract) and then a proper interstellar one (14,000 km/s)

GameData/RP-1/Contracts/InterstellarProbes/InterstellarProbe.cfg
```diff
@@ -0,0 +1,112 @@
...
+{
+    name = InterstellarProbe
+    group = InterstellarProbes
+
+    title = Interstellar Probe
+    genericTitle = Interstellar Probe
+    description = <b>Program: Interstellar PRobes<br>Type: <color=red>Capstone</color></b><br><br> Our probes have explored all there is to explore in the Solar System. We have reached the apex of uncrewed exploration: sending a probe out into interstellar space with sufficiently high velocity to reach another solar system in a reasonable time frame.
+    genericDescription = Reach 40 AU from the Sun at a minimum velocity of 14,000 km/s.
+    completedMessage = Telemetry confirmed. The probe has crossed the 40 AU mark at the target velocity. Godspeed, our brave robotic explorer!
+    synopsis = Launch a probe that reaches an altitude of 40 AU from the Sun while maintaining a heliocentric speed of at least 60 km/s.
```
Clayell: This still says 60km/s

WithCrow: Thanks!

billy54:
Pretty sure accelerating something to 0.04c is impossible unless you orbitally assemble a TD-level spacecraft the likes of which have never been seen since Project Andoria, using 12 launches of pure Orion drive propellant, but it's not like anyone is ever going to reach this point in career for the next five (IRL) decades.
Well, I think it is much easier with solar sails, but those are not configured for RO/RP-1 in any way. Also FFT has antimatter engines but that is again not RP-1 configured

WithCrow:
I disagree that no one would reach it - see my RiS 25 log, I'd "just" have to do the gas giant landings, which are deactived (& hence I did not do them for that run). But I do agree that it is a very steep requirement.

That is also why it spans a century - the player can build up the tech, etc. to send a probe out.

PR #2814, opened 2026-05-27
Allow engineers to be transferred from one KSC to another KSC

### LINKED ISSUE #2584: Add ability to transfer staff between launch sites
As [Kaga](<https://discord.com/channels/319857228905447436/1115321536807718992/1386438314042593473>) brought up, there can sometimes be scenarios where one would like to transfer staff between launch locations. I think this is a totally reasonable and realistic request, provided that there is some sort of 'transfer cost' per staff. I think a reasonable cost would be 5 funds per staff (not covered under unlock credit). This translates to about $5000 USD per staff, which seems reasonable for the cost for a long-distance plane trip + housing etc. 10 funds per staff might also be reasonable. [Overloader](https://discord.com/channels/319857228905447436/1115321536807718992/1473037045361283183) also suggested that this should take some time, depending on the distance between the launch sites and perhaps on the amount of staff being transferred as well.

In Kaga's case, 5 funds per staff would come out to 3800 funds to transfer their staff of 760.

Clayell:
Resolves https://github.com/KSP-RO/RP-1/issues/2584

The functionality is similar to the auto-hire project, in that the transfer can be set up to take money away from you over a period of time instead of taking the money away immediately. The cost to do an engineer transfer is based on the great circle distance with a maximum of 30 funds per engineer and a minimum of 5 funds per engineer. It cannot be run at the same time as a fund target or auto hire target, and vice versa.

The button to switch to the "All Engineers" tab will not show up if KSCSwitcher is not installed.

image
image

Other changes:
* Fix bug where only having engineers in a KSC would not trigger it to be saved
* Cache KSCSwitcher sites, store their latitude and longitude as well, add some methods to calculate great circle distance and get display name
* Remove some unused variables in GUI_Personnel.cs


Source/RP0/ConfigurableStart/PresetPickerGUI.cs
```diff
@@ -125,7 +126,7 @@ private void RenderSelectionWindow(int windowID)
 
         private void InitKSCs()
         {
-            List<(string id, string displayName)> sites = KSCSwitcherInterop.GetAvailableSites();
+            List<(string id, string displayName)> sites = KSCSwitcherInterop.AllSites.Select(s => (s.id, s.displayName)).ToList();
```
Clayell, line 129: Confirming that scenario selector still works
Image

Clayell:
image

Made some changes to make this feel better, notably the improvement that you can now start transfers from any SC, not just the active SC.

This involves some changes to the saving of LCSpaceCenter, so I also ensured that these were working correctly.
image

Clayell:
[RP0.zip](https://github.com/user-attachments/files/31750536/RP0.zip)

Here's a compiled dll of this PR. It's based on RP-1 v4.6.0.0, and should be fully compatible with existing saves. Looking for people to test out the balance of this PR in their own saves to see if the price should be increased or decreased.

Source/RP0/ConfigurableStart/PresetPickerGUI.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
```
siimav: High

1. Transfer charges for a single engineer regardless of how many move — TransferEngineerProject.cs:138
transferCost is documented as per-engineer and GetTimeLeft budgets NumLeftToTransfer * modifiedTransferCost, but IncrementProgress spends a flat 1 ×. Compare KCTUtilities.HireStaff, which spends workersToHire * HireCost. Moving 200 engineers Cape→Woomera costs ~5–30 funds total — effectively free. Same line also deducts funds when numCanTransfer clamps to 0 (all source engineers assigned to LCs), charging for zero engineers moved. Wants numCanTransfer * modifiedTransferCost inside an if (numCanTransfer > 0).

Medium

2. Transfer silently abandoned when the source momentarily has no unassigned engineers — TransferEngineerProject.cs:149
IsComplete() => NumLeftToTransfer <= 0 || OriginalSC.UnassignedEngineers <= 0. Assign the remaining unassigned engineers to an LC mid-transfer and the next ProgressBuildTime tick calls Clear(), dropping the rest with no notification. Meanwhile RenderCombinedList skips the row (if (t.IsComplete()) continue;) so there's no "X" to cancel, yet RenderHireNButton and the MaintenanceGUI warp-to-funds both stay blocked on transferTarget.IsValid.

3. Unclamped Math.Acos → NaN → corrupted funds — KSCSwitcherInterop.cs:173
Dot of two unit vectors can exceed ±1 by ~1e-16, and GetSiteLatLong silently returns (0,0) for any unknown site id (e.g. a legacy Stock KSC). Both endpoints at (0,0) → dot ≈ 1.0000000000000002 → Acos = NaN → Math.Max(NaN, 5d) returns NaN in .NET → SpendFunds(NaN) sets Funding.funds to NaN and the save is corrupted. Clamp the dot; make GetSiteLatLong signal "not found".

4. TransferringEngineers = 1L << 32 degrades to TransactionReasons.None — CurrencyModifierQueryRP0.cs:497
CurrencyUtils.Stock() maps anything > uint.MaxValue to None, and 1L << 32 is exactly one past it. So the real SpendFunds reaches AddFunds with reason None, CareerLog can't attribute it to a personnel bucket, and leader currency modifiers on Personnel never fire on the actual transaction (only on the estimate). Bit 31 is unused and round-trips correctly; bits 33+ are reserved for non-stock reasons. The // unused bits: 0,1,7,8,12,19 comment on line 501 is now stale.

5. Transfer dialog doesn't guard against a transfer already in progress — GUI_Personnel.cs:498
RenderTransferButton checks fundTarget.IsValid || staffTarget.IsValid but not transferTarget.IsValid, though the reciprocal guards exist in both other directions (GUI_Personnel.cs:455, MaintenanceGUI.cs:255). TryAddTransferEngineers then does transferTarget?.Clear() and overwrites — a partially-done transfer is cancelled with no confirmation, leaving engineers stranded at the first destination.

6. Static originalSC never cleared on window close / scene change / save load — GUI_Personnel.cs:19
KCT_GUI statics outlive SpaceCenterManagement. Click "Relocate", close the window, load another save: originalSC still points at the discarded game's LCSpaceCenter, so the originalSC == KSC branch never matches — no Cancel button renders anywhere, and every row offers "Transfer here", building a project whose _originalSC mutates an orphaned space center.

Low

7. AllSites returns null when rawList == null, but GetSiteDisplayName/GetSiteLatLong guard only on IsKSCSwitcherInstalled → NRE (KSCSwitcherInterop.cs:146). PresetPickerGUI.cs:129 has the same NRE, and its following if (sites != null …) is now dead code since .ToList() never returns null.

8. _allSites cached permanently (KSCSwitcherInterop.cs:66) — if first access precedes KSCSwitcher populating its list, an empty list sticks for the session: display names fall back to raw site ids and get persisted into LCSpaceCenter.DisplayName, and every distance is 0 so cost pins to the 5-fund floor.

9. OriginalSC/TargetSC setters NRE on null assignment (TransferEngineerProject.cs:47/:65), unlike HireStaffProject.LC which handles it.

10. GetTimeLeft waits for reserveFunds + NumLeftToTransfer * cost while IncrementProgress fires at reserveFunds + cost, and ignores the UnassignedEngineers cap — "Warp to next" overshoots (TransferEngineerProject.cs:119).

11. Destinations only list already-instantiated KSCs, but SpaceCenterManagement.cs:561 prunes empty non-active KSCs on load — after a save/reload there's no way to seed a first transfer to a fresh site (GUI_Personnel.cs:240).

Findings 1, 4, and 3 are the ones I'd block on: the first makes the cost mechanic non-functional, the second breaks career-log attribution and leader modifiers, the third can write NaN into the player's funds.

PR #2815, opened 2026-05-27
Overhaul Edit Logic

WithCrow:
This PR attempts to 'fix' the KCT/SCM Edit logic. The current implementation is very random, where an edit can reset your build progress to 0% if, for example, a capsule is added onto a launcher. This is a very drastic example, but a more common one is recovering a spaceplane and editing a cheap booster onto it, or editing a payload onto a launcher. The one most people are thinking of, though, is reusing the first stage - this is something that is not viable in the current state.

I will state upfront that my experience with C# is limited to, well, this PR. I know C & C++, but this is truly out of my scope. So I most kindly ask for reviews, etc.

Presently, the new progress after an edit is equal to the current BP - 1.1 * diffBP, where diffBP is the difference between the new total amount of BP and the original total. This changes that to instead use the difference in effective cost and require that (converted to BP) be built. This means that the player only pays for the added parts instead of, for example, paying for the *entire* thing to be integrated again. 
 
 Something siimav flagged is that using this alone, it is possible for the player to circumvent things by adding a part of similar EC as one they remove, jumping the integration. In order to catch this, I've also added a mass based flag. I tried to use a part comparison function to do this, but that was supremely laggy. 
 
On the topic of reuse: there are a few more things that should be tackled (most likely not in this PR), relating to the cost of recovery and refurbishment. In testing, I recovered a F9 first stage and it was ready to be edited in a few hours, with the edit of a 2nd stage onto it taking eight days. This needs to be far more punishing - but that is conceptual, whereas this is technical.


siimav:
OK, cavemen, shoot it full of holes!

WithCrow:
## Current State - Falcon 9 - 29k EC (18k after leaders)
### From Scratch
image
image
As one can see, starting from scratch, the 'bare' F9 costs 16 days to integrate and with Gemini added, 31 days. 
### Editing
image
image
When now editing a fully built F9, the progress is reset to 0% and it takes exactly as long as it would have from scratch.

## Proposed Change - Falcon 9 - 29k EC (18k after leaders)
image
Adding a Gemini now resets the progress to 58%, with 12 days remaining.
### Reuse
image
image
Adding a F9 second stage onto a first stage resets the progress to 49% and costs 8 days to integrate.

WithCrow:
> OK, cavemen, shoot it full of holes!

At the end, this will have five bazillion else if statements to ward against exploits...

WithCrow:
Comparing the editing progress I got previously with reusing DynaSoar:
image
image
This takes 3 days to recover (from the runway). The above is the old/current version, the below is the new one.
image

.gitignore
```diff
@@ -182,3 +182,4 @@ Source/Unity/NewspaperUI/Library/
 Source/Unity/NewspaperUI/PartTools.cfg
 GameData/RP-1/Plugins/PluginData/RP1UpgradeWarningPLC
 GameData/RP-1/Plugins/PluginData/RP1MentionLoadingImages
+*.csproj
```
Clayell: csproj changes should definitely not be in .gitignore. Also, the RP0.dll should not be in the github gamedata.

WithCrow: I will undo the gitignore - but I'm curious as to the latter. With no fancy auto nightly build, I wanted to provide the compiled dll to interested players - what would be a better way of doing that?

Clayell: I usually just post a release with the dll in the fork and link to it. If the dll is in the PR when this gets merged, and then someone downloads the source code because they think all of the dlls are included, they will run into errors. I suppose if it were removed before merge it would be fine, but that requires extra attention.

WithCrow: I realised I could manually post the release as well, so I'll do that. Resolving this conversation now, thanks for the input!

WithCrow:
I hate AI! It said it wouldn't close the PR if I renamed the branch!

PR #2817, opened 2026-05-28
Add Early XPlanes Program Modifier

CXG-2827:
Right now Early X-planes is a major financial advantage for the default 1951 start, but it is not tempting to take once you are in the orbital era.

This modifier makes it more flexible to accept in later stages of a career, by reducing the program slot to 1 and adjusting the funding. From a cost per slot per year, this pays similar to the early satellite programs.

### DIFF
--- a/GameData/RP-1/Programs/Programs.cfg
+++ b/GameData/RP-1/Programs/Programs.cfg
@@ -50,6 +50,42 @@ RP0_PROGRAM
 	}
 }
 
+RP0_PROGRAM_MODIFIER
+{
+	srcProgram = EarlySatellites
+	tgtProgram = EarlyXPlanes
+	nominalDurationYears = 7
+	baseFunding = 400000
+	fundingCurve = Flat
+	repDeltaOnCompletePerYearEarly = 105
+	repPenaltyPerYearLate = 105
+	slots = 1
+
+	CONFIDENCECOSTS
+	{
+		Normal = 350
+		Fast = 700
+	}
+}
+
+RP0_PROGRAM_MODIFIER
+{
+	srcProgram = EarlySatellites-Heavy
+	tgtProgram = EarlyXPlanes
+	nominalDurationYears = 7
+	baseFunding = 400000
+	fundingCurve = Flat
+	repDeltaOnCompletePerYearEarly = 105
+	repPenaltyPerYearLate = 105
+	slots = 1
+
+	CONFIDENCECOSTS
+	{
+		Normal = 350
+		Fast = 700
+	}
+}
+
 RP0_PROGRAM
 {
 	name = SuborbRocketDev
--- a/GameData/RP-1/Programs/Programs.cfg
+++ b/GameData/RP-1/Programs/Programs.cfg
@@ -63,8 +63,8 @@ RP0_PROGRAM_MODIFIER
 
 	CONFIDENCECOSTS
 	{
-		Normal = 250
-		Fast = 500
+		Normal = 300
+		Fast = 600
 	}
 }
 
--- a/GameData/RP-1/Programs/Programs.cfg
+++ b/GameData/RP-1/Programs/Programs.cfg
@@ -54,17 +54,17 @@ RP0_PROGRAM_MODIFIER
 {
 	srcProgram = EarlySatellites
 	tgtProgram = EarlyXPlanes
-	nominalDurationYears = 7
-	baseFunding = 400000
+	nominalDurationYears = 5
+	baseFunding = 285000
 	fundingCurve = Flat
-	repDeltaOnCompletePerYearEarly = 105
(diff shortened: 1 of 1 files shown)


njits23:
This is an interesting idea, worth expanding. For example, the funding of downrange and suborb research programs could be increased to match the current funding of taking X-planes alongside one of those programs. Then a modifier could be added that would reduce the funding of the programs to current levels if you take X-planes.

Blothorn:
X-planes paying more is deliberate; astronaut salaries and the need to research things you could otherwise skip make it a decent bit more expensive to run than the sounding rocket programs. That said, I think the difference is somewhat overtuned right now. My intention was to provide enough funding in the first year to hire an astronaut (7500 including salary) and do one more point of research (20 researchers, 21000 for hiring and salary). That gives 28.5k, somewhat less than the 33k extra F/F X-planes gives compared to F/B sounding rockets. (Maybe I foolishly assumed two astronauts?) Moreover, you can get at least 40 reputation from X-Planes in the first year, which is good for another ~4k in annual income. I'm confident reducing X-Planes funding to 650k would still leave it at least on rough parity with a noplanes start.

Back on topic, I quite like this PR, especially with the in-development spaceplanes programs giving people a reason to catch up.

siimav:
Sounds like a good idea. I do wonder though whether going from 9 years to 7 is still too slow. Researching all the required nodes should be trivial after reaching the orbital rocketry era.

CXG-2827:
I took a blind sample of ~10 careers on spyware to get an idea of when players were completing the program objectives, and that made the 7 year choice. But to your point, the accelerated research should help.

Thinking more about it now, there should also be a larger workforce of engineers at this point in a career, with LC/Hangar engineering efficiencies to moving things along faster.

I'll tweak it down to 5 years.

isturdy:
Aye. It takes a long time if you take it from the start because it has a hard requirement of Hypersonic Flight for the Karman Line contract, and that's a 1959 tech. Otherwise it's only 31 points above the mat sci and rocketry techs that people should already have.

If you take it late enough you actually only need to build three craft--a supersonic jet, high-altitude jet, and rocketplane. The real bottleneck is crew downtime and training, but someone taking it later probably has 3+ astronauts to share the load.

CXG-2827:
Values have been adjusted.

siimav:
Hmm, what were your considerations when picking the funding? I think it might be a bit too low now. Unless... the flat curve helps a ton here?
image

siimav:
OTOH you can pick it alongside either of the orbit programs. Still not sure if it's worth it.

CXG-2827:
I based the funding on (Funds per Year)/Slot. Heavysat is $58.33/yr/slot and that was the original baseline. 

with a 1951 start, X-planes pays ~$60k/year before it starts to ramp up, so you are right, this is a bit light. The modifier doesn't enable until you complete either Heavy or Light Sats, so this would run alongside Skopos0, Early Lunar Probes and the likes.

What about total program budget of 350k, so 70k/yr?

siimav:
> The modifier doesn't enable until you complete either Heavy or Light Sats

I am pretty sure the modifier applies on program acceptance. Because that is the only way it could work with stations.

CXG-2827:
Yes you are correct, I misremembered.

Something I just noticed is the Program button slot count doesn't update to (Slots: 1), but it shows correct in the description window.

image

siimav:
https://github.com/KSP-RO/RP-1/pull/2851 should fix the issue with slot counts

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -50,6 +50,42 @@ RP0_PROGRAM
...
+	baseFunding = 350000
+	fundingCurve = Flat
+	repDeltaOnCompletePerYearEarly = 80
+	repPenaltyPerYearLate = 80
+	slots = 1
+
+	CONFIDENCECOSTS
+	{
+		Normal = 300
+		Fast = 600
```
siimav: I assume you didn't want different values here

CXG-2827: This change was intentional, mostly due to the cheaper admin investment of 1 slot.

siimav: But look up, the other instance has 250 / 500

CXG-2827: I see now, I will update.

siimav: Spreadsheet says 250/500 is correct though. Otherwise this would become the first to cost more than suggested.

CXG-2827: Ok, I reverted both. I was unaware of a balance spreadsheet, is there a link or directory?

siimav:
Let's see if the cave-dwellers find this of use.

PR #2818, opened 2026-05-28
Overhaul Edit

WithCrow:
Continuation of [2815](https://github.com/KSP-RO/RP-1/pull/2815) because I really should stop trusting what Gemini tells me about Github. Attached is 2815's description, see the linked PR for initial comparison pictures.

This PR attempts to 'fix' the KCT/SCM Edit logic. The current implementation is very random, where an edit can reset your build progress to 0% if, for example, a capsule is added onto a launcher. This is a very drastic example, but a more common one is recovering a spaceplane and editing a cheap booster onto it, or editing a payload onto a launcher. The one most people are thinking of, though, is reusing the first stage - this is something that is not viable in the current state.

I will state upfront that my experience with C# is limited to, well, this PR. I know C & C++, but this is truly out of my scope. So I most kindly ask for reviews, etc.

Presently, the new progress after an edit is equal to the current BP - 1.1 * diffBP, where diffBP is the difference between the new total amount of BP and the original total. This changes that to instead use the difference in effective cost and require that (converted to BP) be built. This means that the player only pays for the added parts instead of, for example, paying for the entire thing to be integrated again.

Something siimav flagged is that using this alone, it is possible for the player to circumvent things by adding a part of similar EC as one they remove, jumping the integration. In order to catch this, I've also added a mass based flag. I tried to use a part comparison function to do this, but that was supremely laggy.

On the topic of reuse: there are a few more things that should be tackled (most likely not in this PR), relating to the cost of recovery and refurbishment. In testing, I recovered a F9 first stage and it was ready to be edited in a few hours, with the edit of a 2nd stage onto it taking eight days. This needs to be far more punishing - but that is conceptual, whereas this is technical.


chambm:
Can you come up with a table of cases and show what the BP results are before and after this change?

WithCrow:
I've provided some initial ones in the linked PR, once I have time to properly elaborate on this, I'll provide more. The change is, in the end, a simple change of the formula.

Clayell:
> I tried to use a part comparison function to do this, but that was supremely laggy.

Why was it laggy?

WithCrow:
> Why was it laggy?

Most likely a skill issue on my end. I think my code was, instead of comparing the parts only upon a change, comparing it every single frame. On my first attempt, all of the parts of the edited vessel were duplicated, so I know I did something wrong for sure. 

I'll revisit it in the future, I just lack the C# knowledge to do it properly

billy54:
image

Well, that can't be right. I removed some experiments from my FSO so I could start building it before the tech unlocked, but after editing the experiments onto the craft, it actually got more complete, not less!

I apparently cannot upload .craft files in this comment, would you like me to send them to you for debugging?

(in the meantime, I will definitely be abusing this bug hehe)

WithCrow:
Three letters: W T F

periodically-makes-puns:
okay cool it's in semi-working condition.

We define a part as "identical" to another part if the following properties hold:
- Their `name`s are identical.
- Their dry mass and dry cost are identical.
- Their resource capacities are identical.

To compute the new edit progress (in BP), we do the following:
- Attempt to match up each part on the old craft with all parts on the new craft with the same `name`. If they are "identical", remove them both from consideration and add that part to the list of "identical" parts.
- We now define two vessels and one "pseudo-vessel". The two vessels are the Original and Edited vessels. The pseudo-vessel (which I shall call Similar) contains the final list of all "identical" parts.
- We use the BP values of the Original, Similar and Edited "vessels" as well as the initial progress in the Original vessel to compute a new progress BP value. This formula requires more work, but currently:
  - The progress "carried over" is the initial progress or final total BP, whichever is lower.
  - We penalise the progress by the difference in BP between the Original and Similar vessels, multiplied by a weight (PartRemovalTimePenalty). This represents removing the non-"identical" parts on the Original vessel to reach the Similar vessel.
  - We penalise the progress by the difference in BP between the Similar and Edited vessels, multiplied by a weight (PartAdditionTimePenalty). This represents adding the non-"identical" parts on the Edited vessel to the Similar vessel.
- This formula is extremely subject to change. Also it definitely needs UI work to display which parts are considered "identical".

periodically-makes-puns:
https://github.com/KSP-RO/RP-1/pull/2825#issuecomment-4883991746

Crow found an issue where Kerbals are counted towards part mass only in the VesselProject version (ShipConstruction calculates dryMass through the part prefab + module mass), so we now subtract out anything relating to kerbals or kerbal inventory mass. This appears to have fixed this particular issue on my end.

periodically-makes-puns:
> Also it definitely needs UI work to display which parts are considered "identical".

A toggle now appears if there are any non-matching parts on the new craft, which colour-codes parts according to the first check that failed (cost, mass, resources, name). Still not ideal, but at least there's some visual feedback instead of having to dig through debug logs.

After some discussion with Crow the part removal penalty is gone, for now. The cheese was from counting removed parts towards the progress, which we no longer do (only parts that are "identical" are carried over). This does mean that destructive editing is free, though...

Also the addition penalty has been weakened. This should probably be tech-based? Or maybe it shouldn't exist at all, who knows.

This edit algorithm is still very sensitive to small edits, which I am not a huge fan of. As a theoretical example, editing the crew experiments in a recovered capsule would mark the entire capsule as "non-matching" due to either mass or cost differences.

chambm:
Good example. I haven't tested the latest code, but to deal with that example, you probably need special upfront handling for editing crew experiments and then count those as empty when comparing the capsules later. Alternately, has it been considered to allow parts to be equal just if their part names were equal, and then compare their costs and mass differences to decide how much that edited part should affect the overall similarity, instead of just a boolean for each part?

WithCrow:
I can't think of a way where we can cover crew capsules but still capture engine upgraded

chambm:
> I can't think of a way where we can cover crew capsules but still capture engine upgraded

Each part category could have a multiplier for how much mass/cost/configuration changes would cost in edit%. Capsules could have a very lower multiplier, engines a very high multiplier.

periodically-makes-puns:
> Alternately, has it been considered to allow parts to be equal just if their part names were equal, and then compare their costs and mass differences to decide how much that edited part should affect the overall similarity, instead of just a boolean for each part?

Yes. I considered this, but one of the core assumptions I made with the part matching algorithm is that greedily matching up parts from the old and new crafts is always optimal. If there are multiple parts with the same name (e.g. multiple Procedural Isogrid tanks), the algorithm must distinguish which parts from the old craft are optimal to pair with the parts on the new craft (lest it pair a first-stage fuel tank with an RCS tank on the top stage!)

This can be expressed as a bipartite matching problem (and therefore as a network flow problem). This is possible to solve quickly, but I do not want to maintain a network flow algorithm in the edit function unless I absolutely have to.

chambm:
Do you have a set of test cases that you're testing each iteration of the algorithm on? Seeing those might help the devs and users review the changes of the PR more easily and identify deviations from the intended behavior.

periodically-makes-puns:
Not as of yet, because I don't have any good examples of reusable rocket designs on hand to test with. I've just been editing saved plane designs from my previous runs.

periodically-makes-puns:
Some changes:
- Toggle Non-Matching Highlights button now is green when enabled.
- Parts with the NoResourceCostMult tag get a free pass as to their mass/cost/resources. In practice, this should mean that capsules (and some service modules) can have their contents changed freely.
- Fix merging ignoring the EditedVessel. Also, we now penalise merging vessels (which was not correctly penalised before).
  - Specifically, we credit the sum of the BP of the merged vessels as the "old progress", but the total old BP is computed from the sum of the effective costs of the parts. For two vessels with equal cost being merged together, this results in a ~6.67% penalty. 
  - In my opinion, this is the bare minimum acceptable penalty. It means that if you build two vessels separately then merge them together, it is equivalent to just building the merged vessel to begin with, since the total BP from the three projects sums to the total BP of the merged vessel.

periodically-makes-puns:
Fixed: Cover edge case where the above formula returns >100% initial progress by capping to 100%.

PR #2820, opened 2026-05-29
Add more Airlaunch Upgrades

WithCrow:
This PR adds three more air-launch levels, increasing in speed, altitude and payload mass. 

I suppose more modern airlaunches such as Aurora (or [Vega](https://www.polaris-raumflugzeuge.de/products/hypersonic-platform-vega)  (POLARIS Spaceplanes) or... [Aurora](https://www.dawnaerospace.com/spaceplane-program) (Dawn Spaceplanes) could also be possible.

The original intent was to have sidegrades (go very fast but with a light craft, or have a heavy craft but go slow, etc.), but I couuld not get this to work. If anyone has an idea on how to achieve this, I would be all ears. Another solution would be to change the XB-70 esque launcher to be the last one, but I deem this to be somewhat silly considering it would have (theoretically) been possible.

### DIFF
--- a/GameData/RP-1/SCMData/AirlaunchLevels.cfg
+++ b/GameData/RP-1/SCMData/AirlaunchLevels.cfg
@@ -53,4 +53,66 @@ KCTAIRLAUNCHTECHS
 		MaxMass = 25000
 		MaxSize = 15, 8, 21
 	}
+
+	// C-141 / IL-76
+	AIRLAUNCHTECHLEVEL
+	{
+		TechRequired = advancedJetEngines
+		MinAltitude = 5000
+		MaxAltitude = 13750
+		MaxKscDistance = 1000000
+		MinVelocity = 125
+		MaxVelocity = 272
+		MaxMass = 45000
+		MaxSize = 20, 12, 28
+	}
+	// 747 SCA
+	AIRLAUNCHTECHLEVEL
+	{
+		TechRequired = matureTurbofans
+		MinAltitude = 5000
+		MaxAltitude = 13750
+		MaxKscDistance = 1000000
+		MinVelocity = 125
+		MaxVelocity = 275
+		MaxMass = 78000
+		MaxSize = 28, 20, 41
+	}
+	// An-225
+	AIRLAUNCHTECHLEVEL
+	{
+		TechRequired = refinedTurbofans
+		MinAltitude = 5000
+		MaxAltitude = 13750
+		MaxKscDistance = 1000000
+		MinVelocity = 125
+		MaxVelocity = 275
+		MaxMass = 280000
+		MaxSize = 38, 20, 46
+	}
+	// Heavy Mach 2 Transport 
+	AIRLAUNCHTECHLEVEL
+	{
+		TechRequired = scramjetEngines
+		MinAltitude = 5000
+		MaxAltitude = 18000
+		MaxKscDistance = 1000000
+		MinVelocity = 125
+		MaxVelocity = 600
+		MaxMass = 280000
+		MaxSize = 38, 20, 46
+	}
+	// Heavy Mach 3 Transport
+	AIRLAUNCHTECHLEVEL
+	{
+		TechRequired = advancedTurbofans
+		MinAltitude = 5000
+		MaxAltitude = 21500
+		MaxKscDistance = 1000000
+		MinVelocity = 125
+		MaxVelocity = 850
+		MaxMass = 280000
+		MaxSize = 38, 20, 46
+	}
+
 }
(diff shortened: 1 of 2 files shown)


WithCrow:
Instead of the XB-70 launch, I've changed it to:
- C-141 / IL-76 launch, 45 tons (Advanced Jet Engines)
- 747, 78 tons (Mature Turbofans)
- An-225, 280 tons (Refined Turbofans)
- Mach 2, 280 tons (Scramjet Engines)
- Mach 3, 280 tons (Advanced Turbofans)

chambm:
Is there a vehicle that can carry 280 tons at 21km at 850m/s? :)

WithCrow:
As indicated by the fact I did not cite one, I think you can draw that conclusion yourself :p 

It is most certainly possible to design an aircraft capable of such a feat. HORUS/Saenger II are an historical example of the first step in designing something like that. A modern example is POLARIS. The only reason why something along those lines was never developed is because doing so represents an effort no one is willing to make - but RP-1 should reflect a more optimistic world. 

The XB-70 scratched a bit above what was possible in that time (i.e. 1964), but a modern day bomber can most certainly be designed for Mach 3 performance with a 280 ton load. It's just not worth it to do so.

chambm:
Those spaceplanes seem good for only small payload. Polaris is good for 1 ton to orbit or 10 tons to hypersonic. This is pretty far off from 280! The B1B and Tu-160 have 30ish ton payloads at mach 1.2 and 3 respectively. It really seems like there has to be two class of airlaunch: low & slow, but heavy, and high & fast but light.

WithCrow:
> Those spaceplanes seem good for only small payload. Polaris is good for 1 ton to orbit or 10 tons to hypersonic. This is pretty far off from 280! The B1B and Tu-160 have 30ish ton payloads at mach 1.2 and 3 respectively. 

I think we are talking past one another. I fully agree that no such aircraft exists in real life. However, this is not due to feasibility, but due to practicality. XB-70 perished due to ballistic missile performance & with it, heavy supersonic bombers turned into something like the B-1. 

But, in the world of RP-1, where the player has bases on the Moon and sends crew to Mars, I do not see a problem in providing a realistic extrapolation of present day designs for these upgrade paths.

If the decision is made to exclusively use real life aircraft data - which I disagree with, but what is merged is not up to me - then I will change it to that. 

> It really seems like there has to be two class of airlaunch: low & slow, but heavy, and high & fast but light.

See my comments in #moddev-help, that was my thought as well. With two on-going 'retooling' projects of the KCT/SCM code, I am hesitant to open a third to rework this. Perhaps I will

WithCrow:
On the note of Saenger: Mach 6.8 with a 90 ton payload is far above the current Mach 3 & 280 tons (at the end of the tech tree) ICAS-88-1.5.1.pdf https://share.google/BGftCrVZWsPGZcnFB

chambm:
Really neat document! I don't think we're talking past each other, we just have different ideas of feasibility. I'm not an aerospace engineer so my ideas are very much biased by what has been done before. I'm having trouble figuring out the real payload capacity of the Sanger: the 90t seems unlikely:
image
image

But on the other hand, I think if you can design and fly it in RO, then it's fair game for putting in the tech tree somewhere as a convenient air launch option. Can you design a hypersonic super cargo carrier like that? Now I'm really tempted to try. :P

WithCrow:
I did ask ezsnack for input on this, who said it would be quite easy to do - but I too am tempted to try! After reading through the document with a bit of a closer look, I am of the opinion that they were really optimistic with that design. Then again, they claim Hermes will fly on Ariane 5...

I definitely want to give it a shot (& Saenger in general). I'll report back tomorrow / whenever I finish this new sub-project

GameData/RP-1/SCMData/AirlaunchLevels.cfg
```diff
@@ -53,4 +53,66 @@ KCTAIRLAUNCHTECHS
...
+		TechRequired = advancedJetEngines
+		MinAltitude = 5000
+		MaxAltitude = 13750
+		MaxKscDistance = 1000000
+		MinVelocity = 125
+		MaxVelocity = 272
+		MaxMass = 45000
+		MaxSize = 20, 12, 28
+	}
+	// 747 SCA
```
siimav, line 69: ## Performance

Cruise speed: 250 kn (290 mph, 460 km/h) / M0.6 with Shuttle Orbiter loaded
Range: 1,150 nmi (1,320 mi, 2,130 km) with Shuttle Orbiter loaded
Service ceiling: 15,000 ft (4,600 m) with Shuttle Orbiter loaded

WithCrow: With how these work currently, I only increased things and kept others constant - there's, at least as far as I understand, no way to make a composite envelope - which is what something like that would require

siimav: Yeah, I was kinda hinting at the fact that we need a better system before we can actually add more configs.

WithCrow: Oof. Seperate PR for that, or just change this to "Airlaunch overhaul"?

PR #2823, opened 2026-05-31
Improve the build list auto alarms

### LINKED ISSUE #2813: When auto alarms is on, the alarm will be stuck on the very last build item
Image

Since there's nothing to trigger a deletion. It's not as simple as placing a delete alarm call [here](https://github.com/KSP-RO/RP-1/blob/master/Source/RP0/UI/KCT/GUI_BuildList.cs#L270), as then it would just delete any alarm that starts with "RP-1 " for eternity until a new build item shows up, including alarms that the user creates. It needs to be triggered when the build item first switches from nonnull to null.

Clayell:
~~This has a soft dependency on https://github.com/linuxgurugamer/KerbalAlarmClock/pull/12.~~

* Resolve https://github.com/KSP-RO/RP-1/issues/2813
* Rework the very weird auto alarm logic to just have a cached ISpaceCenterProject and string for the current build item.
  * Fix excessive delete alarm log spam
* Rename buildItem to nextThing
* Turn the long string of if elses into a switch statement
* When auto-alarms is on, deleting an alarm tied to a build item automatically re-generates it

image

### DIFF
--- a/Source/RP0/ModIntegrations/AlarmHelper.cs
+++ b/Source/RP0/ModIntegrations/AlarmHelper.cs
@@ -22,7 +22,6 @@ namespace RP0.ModIntegrations
         /// </returns>
         public static string CreateAlarm(string title, string description, double UT, KACAPI.AlarmTypeEnum alarmType = KACAPI.AlarmTypeEnum.Raw)
         {
-            string s = "Alarm created with ID: ";
             if (UseKAC)
             {
                 if (alarmType == KACAPI.AlarmTypeEnum.Contract || alarmType == KACAPI.AlarmTypeEnum.ContractAuto)
@@ -38,12 +37,12 @@ namespace RP0.ModIntegrations
                     {
                         alarm.AlarmAction = KACAPI.AlarmActionEnum.KillWarp;
                         alarm.Notes = description;
-                        RP0Debug.Log(s + alarmID);
+                        RP0Debug.Log($"Alarm created with ID: {alarmID}");
                         return alarmID;
                     }
                     else
                     {
-                        RP0Debug.Log($"KAC Alarm with ID: {alarmID} could not be found."); // this is usually fine, because KAC waits until the end of the frame to add the alarm
+                        RP0Debug.LogWarning($"KAC Alarm with ID: {alarmID} could not be found.");
                         return alarmID;
                     }
                 }
@@ -66,7 +65,7 @@ namespace RP0.ModIntegrations
                 if (alarm.Id != 0u)
                 {
                     string alarmID = alarm.Id.ToString();
-                    RP0Debug.Log(s + alarmID);
+                    RP0Debug.Log($"Alarm created with ID: {alarmID}");
                     return alarmID;
                 }
                 else
@@ -85,7 +84,6 @@ namespace RP0.ModIntegrations
         /// </returns>
         public static bool ChangeAlarm(string id, string title, string description, double UT)
         {
-            string s = $"Alarm with ID: {id} changed to have title: {title}, description: {description}, and time: {UT}.";
             if (UseKAC)
             {
                 KACAlarm alarm = KACWrapper.KAC.Alarms.FirstOrDefault(a => a.ID == id);
@@ -98,7 +96,7 @@ namespace RP0.ModIntegrations
                 alarm.Notes = description;
                 alarm.AlarmTime = UT;
                 // KAC does not let you change the alarm type of an alarm after it has been made
-                RP0Debug.Log(s);
+                RP0Debug.Log($"Alarm with ID: {id} changed to have title: {title}, description: {description}, and time: {UT}.");
                 return true;
             }
             else
@@ -116,7 +114,7 @@ namespace RP0.ModIntegrations
                 alarm.title = title;
                 alarm.description = description;
                 alarm.ut = UT;
-                RP0Debug.Log(s);
+                RP0Debug.Log($"Alarm with ID: {id} changed to have title: {title}, description: {description}, and time: {UT}.");
                 return true;
             }
         }
@@ -129,13 +127,11 @@ namespace RP0.ModIntegrations
         /// </returns>
         public static bool DeleteAlarmWithID(string id)
         {
-            string s1 = "Alarm deleted with ID: ";
-            string s2 = "Could not delete alarm with ID: ";
             bool successful;
             if (UseKAC)
             {
                 successful = KACWrapper.KAC.DeleteAlarm(id);
-                RP0Debug.Log((successful ? s1 : s2) + id);
+                RP0Debug.Log((successful ? "Alarm deleted with ID: " : "Could not delete alarm with ID: ") + id);
                 return successful;
             }
             else
@@ -143,7 +139,7 @@ namespace RP0.ModIntegrations
                 if (uint.TryParse(id, out uint alarmID))
                 {
                     successful = AlarmClockScenario.DeleteAlarm(alarmID);
-                    RP0Debug.Log((successful ? s1 : s2) + id);
+                    RP0Debug.Log((successful ? "Alarm deleted with ID: " : "Could not delete alarm with ID: ") + id);
                     return successful;
                 }
                 else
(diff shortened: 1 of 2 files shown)


Source/RP0/ModIntegrations/AlarmHelper.cs
```diff
@@ -43,7 +43,8 @@ public static string CreateAlarm(string title, string description, double UT, KA
                     }
                     else
                     {
-                        RP0Debug.LogError($"KAC Alarm with ID: {alarmID} could not be found.");
+                        //RP0Debug.LogError($"KAC Alarm with ID: {alarmID} could not be found."); // this is usually fine, because KAC waits until the end of the frame to add the alarm
```
siimav: The old commented-out lines should just be deleted.

Source/RP0/UI/KCT/GUI_BuildList.cs
```diff
@@ -127,88 +129,95 @@ public static void DrawBuildListWindow(int windowID)
...
-                    locTxt = txt;
-                    txt = "Training";
+                            locTxt = ar.LC.Name;
+                        }
+                        break;
+                    case ProjectType.VAB:
+                    case ProjectType.SPH:
+                        if (nextThing is VesselProject vp)
+                        {
+                            locTxt = vp == null || vp.LC == null ? "Vessel" : vp.LC.Name;
```
siimav: vp == null is always false here

Source/RP0/UI/KCT/GUI_BuildList.cs
```diff
@@ -127,88 +129,95 @@ public static void DrawBuildListWindow(int windowID)
...
+                                    locTxt = reconRoll.launchPadID;
+                                    break;
+                                }
+                                case ReconRolloutProject.RolloutReconType.Rollback:
+                                {
+                                    VesselProject associated = reconRoll.LC.Warehouse.FirstOrDefault(vp => vp.shipID == reconRoll.AssociatedIdAsGuid);
+                                    txt = $"{associated.shipName} Rollback at {reconRoll.launchPadID}";
+                                    locTxt = reconRoll.launchPadID;
+                                    break;
+                                }
```
siimav, line 165: The header renders both labels side by side: GUILayout.Label(txt); GUILayout.Label(locTxt, ...), so this produces "Saturn V Rollout at Pad A | Pad A". Either remove locTxt = reconRoll.launchPadID for those two cases, or drop the pad from txt and let locTxt carry it.

Source/RP0/UI/KCT/GUI_BuildList.cs
```diff
@@ -217,61 +226,37 @@ public static void DrawBuildListWindow(int windowID)
...
-                        SpaceCenterManagement.Instance.AlarmId = AlarmHelper.CreateAlarm(txt, "", SpaceCenterManagement.Instance.AlarmUT, alarmType);
-                    }
-                }
             }
             else
             {
                 GUILayout.Label("No Active Projects");
             }
             GUILayout.EndHorizontal();
+            if (KCTSettings.Instance.AutoAlarms && buildItem != null && (buildItem != nextThing || txt != itemText || !AlarmHelper.AlarmExistsTitle(txt, out string _)))
```
siimav: AlarmExistsTitle searches for any alarm whose title contains txt. But txt here can be things like "Training", "Reconditioning", or a vessel name like "Saturn V". This can match unrelated alarms the player created manually (e.g. a KAC alarm titled "Training for Mars Mission"). A false positive here suppresses alarm creation when it shouldn't.

Better approach: AlarmHelper.AlarmExistsID(SpaceCenterManagement.Instance.AlarmId) is already available and checks the exact stored ID, which is unambiguous.

siimav (changes requested):

Source/RP0/UI/KCT/GUI_BuildList.cs
Clayell: I'll do the former, as in this case txt is carried into the auto alarm, and I wouldn't want the pad to be left out of there. I'll also explicity set locTxt to be empty, as otherwise it would still be "VAB".
It makes me think that GetItemName should really just give out these strings without us having to format them further, but that's a problem for a future time.

Source/RP0/UI/KCT/GUI_BuildList.cs
Clayell: Now that the weird logic from before is gone, we can probably just entirely rely on the AlarmId. I'll do some tests and confirm that this is fine.

Clayell:
Pictures of changes:

image
image
image


As you adjust the time for a construction project, it'll update the alarm without needing to create or delete it. This reduces a lot of log spam that was previously present.

Works in stock too:
image
image

Although, if you change the name while the alarm window is open, the alarm window won't show the change. This is definitely a stock issue, because once you hit 'apply changes' it'll show up (or delete the alarm and let RP-1 regenerate it), and you can see it showing up correctly in the top left. Maybe I'll throw it in KSPCF, idk.
image

There's also an NRE from stock that shows up sometimes and that goes away when you delete the alarm and let RP-1 regenerate it. Idk, the stock alarm stuff is weird.
image
```
[EXC 15:16:47.577] NullReferenceException: Object reference not set to an instance of an object
	AlarmTypeRaw.OnUIEndInitialization (KSP.UI.AlarmUIDisplayMode displayMode) (at <4b449f2841f84227adfaad3149c8fdba>:0)
	AlarmTypeBase.UIEndInitialization (KSP.UI.AlarmUIDisplayMode displayMode) (at <4b449f2841f84227adfaad3149c8fdba>:0)
	KSP.UI.AlarmClockUIAddPane.SetupEditAlarm (AlarmTypeBase alarm) (at <4b449f2841f84227adfaad3149c8fdba>:0)
	KSP.UI.AlarmClockUIFrame.OnEdit () (at <4b449f2841f84227adfaad3149c8fdba>:0)
	UnityEngine.Events.InvokableCall.Invoke () (at <12e76cd50cc64cf19e759e981cb725af>:0)
	UnityEngine.Events.UnityEvent.Invoke () (at <12e76cd50cc64cf19e759e981cb725af>:0)
	UnityEngine.UI.Button.Press () (at <0c883c16262a4198973f823d01e07427>:0)
	UnityEngine.UI.Button.OnPointerClick (UnityEngine.EventSystems.PointerEventData eventData) (at <0c883c16262a4198973f823d01e07427>:0)
	UnityEngine.EventSystems.ExecuteEvents.Execute (UnityEngine.EventSystems.IPointerClickHandler handler, UnityEngine.EventSystems.BaseEventData eventData) (at <0c883c16262a4198973f823d01e07427>:0)
	UnityEngine.EventSystems.ExecuteEvents.Execute[T] (UnityEngine.GameObject target, UnityEngine.EventSystems.BaseEventData eventData, UnityEngine.EventSystems.ExecuteEvents+EventFunction`1[T1] functor) (at <0c883c16262a4198973f823d01e07427>:0)
	UnityEngine.DebugLogHandler:LogException(Exception, Object)
	ModuleManager.UnityLogHandle.InterceptLogHandler:LogException(Exception, Object)
	UnityEngine.EventSystems.EventSystem:Update()
```

Source/RP0/ModIntegrations/AlarmHelper.cs
siimav: Log string allocation. This builds the full interpolated log string unconditionally at the top, even if it returns false immediately

Source/RP0/UI/KCT/GUI_BuildList.cs
```diff
@@ -32,6 +32,8 @@ public enum RenameType { None, Vessel, Pad, LaunchComplex };
         private const int _width1 = 120;
         private const int _width2 = 100;
         private const int _butW = 20;
+        private static ISpaceCenterProject buildItem;
```
siimav: buildItem now means "the build item from the previous frame". Consider _prevBuildItem for the static
field to signal its temporal nature.

Source/RP0/UI/KCT/GUI_BuildList.cs
```diff
@@ -32,6 +32,8 @@ public enum RenameType { None, Vessel, Pad, LaunchComplex };
...
+        private static string itemText;
```
siimav: _

siimav (changes requested):

Source/RP0/ModIntegrations/AlarmHelper.cs
Clayell: Maybe it would be good to reduce the repetition from these logs across KAC and stock, but eh... would probably make it more complicated to read.

PR #2826, opened 2026-06-01
Add German Localization

WithCrow:
Provides German localization - not sure if 'de-De' is the right one for this.


siimav:
Other mods appear to use de-de. No idea if it actually is case sensitive.

WithCrow:
Better to be safe than sorry - both filename & the localization itself have been changed to lowercase

PR #2828, opened 2026-06-03
Add Advanced X-Planes Program

WithCrow:
Continuation of the (seemingly?) discontinued [2507](https://github.com/KSP-RO/RP-1/pull/2507) PR. If Luci resumes / wants this one closed, I will obviously close it.

I have - for now -  tweaked things superficially (contract titles, descriptions, file structure, etc.) and brought the contracts up to date. Since the original PR also added more airlaunch upgrades (something I wasn't aware of before), but I already have [2820](https://github.com/KSP-RO/RP-1/pull/2820), so I did not include them here.

I also disagree with 'either VTOL or Supersonic Transport' capstone missions. The former is, once certain parts are added, not terribly difficult - and the later should be required. Until I start playtesting, I will not tweak things further, but a few preliminary notes:
- As capstones, it could be any of the presently optional SST Speed, Endurance or Capacity contracts
- Or, a combination - STOL SST, etc. 
- I really do not think that requiring 3 to 6 hour flight times are reasonable. That is simply tedium in my opinion, but I am not sure how to enforce the capability of a sustained supercruise, etc. without it.


WithCrow:
Ah, one actual tweak did happen: Confidence costs have been increaed to align with other crewed programs

chambm:
The supersonic transport development contracts seem interesting but the passenger requirements don't seem feasible with RP-1 parts. Shall we use sounding or research payload as a proxy? But then it'd be almost entirely redundant with supersonic heavy transport contracts. So they should probably just be nixed.

WithCrow:
Isn't there a passenger cabin? What disqualifies that (other than the diameter?

chambm:
The narrow body passenger cabin holds 9 and the Mk3 fuselage cabin holds 16. Both would be pretty annoying to build a 120+ crew plane with. Also the Mk3 parts are in the 1981-1985 node and each one costs $3.7M to unlock, which seems excessive for this contract's purpose. How about I refactor them to be Heavy Transport contracts:
New contract | Ramps | Fixed | Payload mechanic
-- | -- | -- | --
AXP_Supersonic_Heavy_Speed | speed M2.5→3.7 (850→1225) | 25 t payload, 6.12 Mm range | HasResource ResearchPayload = 25 000 u
AXP_Supersonic_Heavy_Capacity | payload 30→60 t (30/40/50/60 k units) | M2.0–2.25, 7.02 Mm range | ramped via payloadUnits list
AXP_Supersonic_Heavy_Endurance | range 7→14 Mm | 25 t payload, M2.0–2.25 | HasResource ResearchPayload = 25 000 u

The "passengers" are the research payload.

WithCrow:
The narrow body only holds 9?! In regards to MK3: that's the STS unlock credit, so yeah, not what we want here

chambm:
Yeah and it has a door on the texture so it looks super goofy when you concatenate 13 of them together to reach 120 capacity. :)

WithCrow:
After looking at this some more in a real playthrough: 3 slots is just too much - even Mars surface exploration pays more per year, and that's a flat curve! So I am debating going down to two slots, or maybe even just one. There is no way you're taking this program with three slots currently

Its-Just-Luci:
> The narrow body only holds 9?! In regards to MK3: that's the STS unlock credit, so yeah, not what we want here

My plan was to make configs for [NeistAir Reupdated](https://forum.kerbalspaceprogram.com/topic/208435-neistair-reupdated-updated-by-kerbmario-112x/) as they have passenger cabins with realistic capacities and layouts, just need to be scaled and massed appropriately. 

Separate parts for Subsonic (I.E. not heat shielded), Supersonic (Mach 3-4), and Hypersonic could be done if mass-accuracy is considered important, though IDK how drastic the shielding mass would be.

Its-Just-Luci:
IIRC they have double-decker cabin segments seating up to 37, and sizes down to single-aisle as well.

WithCrow:
Now that Luci's back in action, I will close this (but not delete) PR for the time being

Its-Just-Luci:
https://github.com/KSP-RO/RP-1/pull/2940

Just in time :D

WithCrow:
LOL, what a coincidence!

PR #2829, opened 2026-06-03
Fix build cost calculation building non-human-rated vessels in HR LCs

Blothorn:
`effectiveEngCount` doesn't actually account for the fact that non-human-rated vessels may not be able to use all the engineers at a human-rated LC. (`bR / rateWithCurEngis` is more or less one--it seems to be just the ratio between the build rate and the build rate after being round-tripped through a string.)

Testing:
- Verified that the net salary estimate of a non-human-rated vessel was the same in a human-rated and non-human-rated LC (aside from a small difference from efficiency)
- Verified that the net salary estimate did not vary with the number of assigned engineers in both human-rated and non-human-rated LCs


siimav:
> (aside from a small difference from efficiency)

From the fact that efficiency gain is slower when the LC is only running with partial amount of engineers?

Blothorn:
Mostly from the fact that my HR LC had 2% lower starting efficiency but yes, even when I corrected for that there was a residual with the sign and rough magnitude I'd expect from that difference in efficiency gain.

PR #2830, opened 2026-06-03
Activate and Revamp Crewed Venus Exploration Program

hayghostin:
This Pr Activates Crewed Venus and freshens up old contracts related to crewed venus travel. There is still WIP to do such as needing feedback on funding and flesh out contract requirements to modern rp-1 contract standards. 

Things I know need to be changed
Crewed Orbit Parameters:
-Looking for altitude ideas to hit? Its just using the old math in the old contract as place holder. 
-Needs Duration of situation (Should it be 5 minutes or a month or what?)

Funding:
image
Credit to @WithCrow for this spreadsheet 
This is the funding compared to what I have tried to at least place around Early Space Stations and ELH. The reasoning I'm starting off with is from the Apollo Applications Program Venus flyby was proposed to be on existing Sat V and Skylab Wet Workshop tech at that time, so in game you should have already unlocked and funding the cost for development after doing moon boots and early space station. This is just something to at least get something i thought is reasonable and would start us somewhere. Everything about this is flexible.

I haven't made them yet but want to add some science that you can take with this and adding in newspaper milestones for the contracts

The structure of the program is set up to either be done by doing a venus flyby which was as far as i understand as a main goal for crewed venus irl. There is the option of going future and orbiting and having that longer duration stay (that needs to be implement as said above just need feedback on how long). And repeated contracts after so the play can go back a few times.

Why add venus when we have mars with more to do? The idea i have is this would be a great step to mars from moonboots and stations as a way to at least get your space agency to then learn about human interplanetary travel before risking lives on the way to mars which takes longer times away from home.


GameData/RP-1/Programs/CrewedVenusExp.truecolor
CXG-2827, line 1: change this from .png to .truecolor

GameData/RP-1/Contracts/Venus Crewed/VenusFlybyCrewREP.cfg
```diff
@@ -0,0 +1,102 @@
...
+			crewOnly = true
+			title = Have at least 2 crewmember on board
+			hideChildren = true
+		}
+		PARAMETER
+		{
+			name = FlybyPlanet
+			type = ReachState
+			targetBody = Venus
+			maxAltitude = 20000000
```
CXG-2827: You might need to increase this to 40Mm. I did a 1978 Venus free-return mission and had a 25Mm periapsis. Another KSPTOT calc gave a solution for a 1979 window with a 33Mm periapsis.

GameData/RP-1/Contracts/Venus Crewed/VenusOrbitCrew.cfg
```diff
@@ -0,0 +1,88 @@
+CONTRACT_TYPE
+{
+	name = first_VenusOrbitCrew
+	title = First Crewed Venus Orbit
+	group = CrewedVenusExp
+	agent = Flag Planting
+
+	description = Design and successfully launch a craft that can carry at least two crew into orbit around Venus and return them safely to Earth.&br;&br;<color=green>NOTE: This contract will become unavailable once the Crewed Venus Flyby contract is accepted or completed.</color>
```
CXG-2827, line 8: "Requirement" parameters will need to be adjusted to hide this orbit contract if the flyby has been accepted.

GameData/RP-1/Contracts/Venus Crewed/VenusFlybyCrew.cfg
```diff
@@ -0,0 +1,87 @@
+CONTRACT_TYPE
+{
+	name = first_VenusFlybyCrewed
+	title = First Crewed Venus Flyby
+	group = CrewedVenusExp
+	agent = Flag Planting
+
+	description = We have sent uncrewed missions to our neighboring planet, Venus, but now we want to send Humans to explore. Design, build and launch a mission that sends at least two humans on a flyby of Venus and return them home safely to Earth.&br;&br;<color=green>NOTE: This contract will become unavailable once the Crewed Venus Orbit contract is accepted or completed.</color>
```
CXG-2827, line 8: "Requirement" parameter needs to be updated to hide this contract if the Orbit contract is accepted.

billy54:
We already have a contract in Crewed Mars where you have to flyby Venus on the way to or back from Mars, how do we integrate that with this program? That would conceivably allow both programs to be completed with one flight, first a Venus flyby to get to Mars, then capture and come back to Earth. I am not sure if it is worth making a program whose objectives are just a basic flyby.

I suggest to create a second Venus program (Advanced Crewed Venus Exploration) containing the orbit contract as well as some more interesting stuff regarding remote-controlled probes dropped from a crewed orbiter. In that case, I would propose to take the Venus-Mars contract out of Crewed Mars, and instead place it into this program, which would be renamed "Early Crewed Interplanetary Exploration" and can be taken alongside Advanced Venus or what is currently Crewed Mars, renamed to Advanced Crewed Mars Exploration (and then the hypothetical mars colony program could be "Long-Term Mars Habitation" or something). In fact there is a case for reducing the station science requirement of ECIE to just prototype station science: the IRL Apollo-Venus proposal would have launched in 1973, which is the year that Skylab launched. They proposed to launch wetlab S-IVBs as habitation tests in the time before that but they would not have lasted the entire 2 years that second gen station science currently requires.

I will also do some testing when I have time to see if it is actually impossible to have a floating balloon base on Venus as so many people seem to think. There is a mod called AirPark Continued that allows you to set vessels as landed in midair, and a number of balloon mods that let you make airships. Although those are not RO/RP-1 configured, other people on the Discord have said they work fine. TD has also done a crewed Venus landing using those mods on an older version of RP-1. If it works I suggest we add it to Advanced Venus as an alternative program completion path, like how Crewed Mars can currently be completed by landing crew on Mars or uncrewed sample return and sending crew to the moons. There is also the possibility of adding an optional uncrewed Venus atmospheric sample return, which has been proposed for some future missions. I think it would make sense if this is required for the path where you dont send the nauts down, but that would mean both completion paths require Airpark as the real life missions would have the sample collector float around for multiple days to gather samples from various interesting regions. Even if we shorten the time to a few hours that is still a lot of physics warping. With Airpark you can use on rails warp and even switch to other vessels.

GameData/RP-1/Programs/CrewedVenusExp.truecolor
hayghostin: oops ive made so many of these and completely forgot to do it correctly on my own program

hayghostin:
> We already have a contract in Crewed Mars where you have to flyby Venus on the way to or back from Mars, how do we integrate that with this program? 

I was thinking having the opposition contract itself be locked behind this program's flyby or having crewed Venus be locked away if you have mars active and vice versa. 

>  I am not sure if it is worth making a program whose objectives are just a basic flyby.

Right now its just a basic flyby to at least match the apollo-venus proposal and give a start of what a program would look like for those missions and allow the option to also orbit as a potential test for a venus station as we dont have anything currently in RO that would allow the HAVOC concept to work in-game without adding Airpark to support the concept of Venus airships.

>  In fact there is a case for reducing the station science requirement of ECIE to just prototype station science

I initially had it as prototype i just didn't know if it would be right to do since 2nd gen is multi month habitation research and prototype is multi week. A mission to Venus is still multi-month but i'm still onboard for prototype science though.

> There is a mod called AirPark Continued that allows you to set vessels as landed in midair, and a number of balloon mods that let you make airships. Although those are not RO/RP-1 configured, other people on the Discord have said they work fine. TD has also done a crewed Venus landing using those mods on an older version of RP-1. 

Im just worried that this is just another mod that would need to be maintained on top of the entire suite of mods that are required. I don't want to introduce a whole new mod and disrupt everything unless its been cleared that it would be fine and everything. Which is why I think it would be better to at least have this program for the crew portion be focused on flyby and orbital venus station science that nicely pairs with the existing Venus surface program. 

>  There is also the possibility of adding an optional uncrewed Venus atmospheric sample return,

I kind of would like this to be in this program as a capstone. Haven't made a contract for this yet because i got to make the science piece to make it work.

GameData/RP-1/Contracts/Venus Crewed/VenusFlybyCrew.cfg
Clayell: I'd also add that `<color=green>Required</color>` should preferably be in the description of these contracts somewhere, to make it consistent with existing contracts.

CXG-2827:
> Right now its just a basic flyby to at least match the apollo-venus proposal and give a start of what a program would look like for those missions and allow the option to also orbit as a potential test for a venus station as we dont have anything currently in RO that would allow the HAVOC concept to work in-game without adding Airpark to support the concept of Venus airships.

What about having the crew go into the atmosphere and maintaining altitude between an upper and lower limit for 5-10 minutes? This would simulate the challenge of descending into the atmosphere and returning back to orbit, but remove the airpark dependency.

PR #2831, opened 2026-06-04
Reduce X-Planes funding

Blothorn:
I believe that X-Planes should have rough financial parity with the sounding rocket starts, accounting for its increased costs (both hiring astronauts earlier than otherwise needed and researching nodes that have little other benefit). It's clear that the current funding is generous for that aim--even for people who do fly in the first year, X-Planes provides enough extra funding to meaningfully accelerate the sounding rocket programs, and pulling early sats funding forward more than compensates for any later trouble.

Looking back at my original program cost estimates I do not seem to have accounted for research efficiency, increasing the number of researchers required and increasing overall personnel costs by about 100k. I thus think that makes a reasonable starting point for funding reduction. If anything I expect it's still possibly generous to X-Planes--it doesn't account for Flight Director bonuses, and since those aren't available to no-planes runs until crewed orbit research there isn't much opportunity cost. It also doesn't account for the fact that X-Planes offers extra research no-planes runs can't do, permanently increasing the research bonus, and that they generally provide more reputation than Suborbital Research.

### DIFF
--- a/GameData/RP-1/Programs/Programs.cfg
+++ b/GameData/RP-1/Programs/Programs.cfg
@@ -6,7 +6,7 @@ RP0_PROGRAM
 	description = This program tasks the space agency with the development of experimental aircraft to test aerodynamic behavior at velocities and altitudes previously unreachable by crewed craft.
 	objectivesPrettyText = Complete X-Planes (Mach 2 Supersonic) and X-Planes (Karman Line) contracts.
 	nominalDurationYears = 9
-	baseFunding = 800000
+	baseFunding = 700000
 	fundingCurve = BimodalBackloaded
 	repDeltaOnCompletePerYearEarly = 130
 	repPenaltyPerYearLate = 130


CXG-2827:
If you compare the funding received the first 3 years in either Suborbital or Rocket dev (73.4k funds) to X-planes (181.6k) , X-planes pays over double, even with your funding reduction applied (158.9k). If you have the time, why not run a brand new career and only accept x-planes on Fast? This way you can isolate the program and see how the funding plays with the progression.

Running that single program test should get a good read I would think.

Blothorn:
I've largely been comparing Fast X-Planes to Breakneck sounding rockets. X-Planes on fast is slightly more confidence, and I think it's actually slightly more demanding in tech--completing X-Planes requires 1959 tech with a deadline at the end of 1959, while both sounding rocket programs can be comfortably completed with 1952 tech and have a deadline halfway through 1955 on Breakneck.

I think a single-program comparison is pointless because just doing X-Planes doesn't unlock any other programs. You need to take one of the sounding rocket programs alongside X-Planes, and then the lack of synergy hurts. The second sounding rocket program costs little other than the engineers' salaries; X-Planes requires hangar GSE, pilot salaries, and extra researchers and unlock costs. I will try an X-Planes run with this funding on my next start, but I'm confident that the balanced funding level is meaningfully higher.

CXG-2827:
> I think a single-program comparison is pointless because just doing X-Planes doesn't unlock any other programs. 

The point of that suggestion was to test X-planes solo, and end the run when program goals are met. Do the run just focusing on staffing levels and tech research just to accomplish the x-planes program. You can then see if you end up with a considerable financial surplus.

siimav:
I'm interested in seeing whether https://github.com/KSP-RO/RP-1/pull/2817 changes the dynamics now.

PR #2832, opened 2026-06-05
Add Crewed Asteroid Exploration Program

WithCrow:
Based on proposals to visit NEO and [others](https://www.universetoday.com/articles/a-novel-concept-for-a-multiplanetary-crewed-mission-to-mars-and-ceres), before the player sends crew out to the Gas Giants, Ceres & Vesta pose an interesting alternative destination that is a bit more feasible. This also builds on CXG's [Asteroid Exploration overhaul](https://github.com/KSP-RO/RP-1/pull/2710), which already calls for the player to land probes on both.

The program currently calls for the player to land on either Ceres or Vesta. The funding per year per slot is between Crewed Lunar (220k, ish) and Crewed Mars (370k) at 230k - depending on how complicated such a mission architecture is, this is subject to change.

Based on preliminary maths, these missions should take ~1800 days, which is - as far as I understand it - a bit outside of the realm of possibility of stress management. With some more additions to ROKerbalism, this should not be a major challenge, though. Due to the fact that I cannot guarantee the feasability of such a mission in the present RO/RP-1 suite, I am leaving this as a draft for now.

### DIFF
--- a/GameData/RP-1/Contracts/Asteroid Crewed/CrewedCeresLanding.cfg
+++ b/GameData/RP-1/Contracts/Asteroid Crewed/CrewedCeresLanding.cfg
@@ -1,0 +1,103 @@
+CONTRACT_TYPE
+{
+	name = CrewedCeresLanding
+	title = First Crewed Ceres Landing
+	group = CrewedAsteroid
+	agent = Flag Planting
+
+	description = <b>Program: Crewed Asteroid Exploration<br>Type: <color=red>Capstone</color></b><br><br>
+
+	synopsis = Send a crew of 2 to land on the surface of Ceres and return home safely
+
+	completedMessage = Congratulations! Human beings have set foot on a dwarf planet in our Solar System. The monumental task that you have accomplished is something to be proud of!
+
+	sortKey = 2000
+
+	cancellable = true
+	declinable = false
+	autoAccept = false
+	minExpiry = 0
+	maxExpiry = 0
+	maxCompletions = 1
+	maxSimultaneous = 1
+	deadline = 0
+
+	targetBody = Ceres
+
+	prestige = Exceptional
+	advanceFunds = 0
+	rewardScience = 0
+	rewardReputation = 4000
+	rewardFunds = 0
+	failureReputation = 0
+	failureFunds = 0
+
+	// ************ REQUIREMENTS ************
+
+	REQUIREMENT
+	{
+		name = ProgramActive
+		type = ProgramActive
+		program = CrewedAsteroid
+	}
+
+	PARAMETER
+	{
+		name = vesselGroup
+		type = VesselParameterGroup
+		title = First Crewed Ceres Landing
+		define = CrewedCeresLanding
+
+		PARAMETER
+		{
+			name = TwoCrew
+			type = HasCrew
+			minCrew = 2
+			maxCrew = 99
+			title = Have at least 2 crewmembers on board
+			hideChildren = true
+		}
+		PARAMETER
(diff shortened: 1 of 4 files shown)


Glloyd:
I'll add this to my save and try it out at some point. That said, given that the complexity is similar to Marsboots, and the timeframe is even longer (and the D/V requirements much higher), I would err on the side of giving more funding than less. Imo it should be closer to the Marsboots end of the scale.

WithCrow:
@CXG-2827 suggested having a funding change similar to Moonboots & Stations where picking one lowers the other - I think this is quite wise. If you do one, you'll have the infrastructure to do the other, but I do agree that the funding can be taken up even further

Glloyd:
That's a good idea, although even then I don't think it should be as much of a drop as in those programs. This is arguably more complex than Marsboots because of the additional shielding and stress management issues that a longer trip entails.

WithCrow:
I've now brought it up to 6 slots, 25 years and 80 million, coming out to a per slot per year of 350k, whereas crewed mars has  370k (but only pays 40 million for 12 years)

WithCrow:
Upon thinking it some more - what about making crewed mars an explicit requirement? I've begun drafting Crewed Jupiter, so my thought process would be Mars -> Belt Asteroids -> Jupiter, with a continuous increase in mission duration.

CXG-2827:
I think both Asteroids and Jupiter could unlock with just Crewed Mars. Allow more freedom to players to skip the asteroids entirely.

25 Years seems a bit much for the program duration. 15 Years might be more appropriate. Here is my opinion related to funding, and this was how I approached the Crewed Mars PR:
- Fast timeline allows for some fluff to build a new LC (assuming no Mars gating), time for integration and launching of the hardware to orbit, and give you 1-2 transfer windows to accomplish the program. Some tolerance for failures.
- Breakneck should make you think hard before committing, as the compressed timeline likely requires the use of existing LCs and limit you to a single launch window to the complete the program before the deadline.
- Normal speed will naturally follow the balance from Fast/Breakneck.

Another funding consideration. If Mars will gate this program, as assumption could be made that improved researcher and engineer efficiencies means that that total funding for operational expenses and integration will be reduced. I'm actually revisiting the funding for Crewed Mars, and I might be submitting a PR to reduce it some. The pricing was modeled from early stations era tech with a hefty 3500 researchers on payroll. From what I can tell, about 2000 researchers is enough to stay slightly ahead of IRL research dates.

Clayell:
Crewed asteroids are certainly easier than crewed mars, no? Asteroids would certainly be easier to launch off of, although I suppose there would be no aerobraking.

CXG-2827:
related asteroid mission difficulty discussion here:
https://github.com/KSP-RO/RP-1/pull/2710

Glloyd:
Definitely not easier than Mars, at least based on my experience back in pre-P&LC RP-1. It's a much longer trip to and from either, needing more shielding for the journey and more attention put into stress management. On top of that, the d/v requirements are slightly higher even for the most energy efficient direct transfer, which takes even longer, compounding the radiation/stress problems. They're easier to land and take off from, but that's about it.

WithCrow:
Here's what I would propose: 30 Million, 15 years, conf. cost increased to 7000 & 14000 - this pays 220k per year per slot, compared to ELH's 200k per year per slot and crewed mars' 370k per year per slot. This fits quite well IMO, as just like ELH, you can continue using the infrastructure of the previous program. 

And I agree on Asteroid & Jupiter being parallel instead of sequential programs.

PR #2833, opened 2026-06-05
Reduce Outer Planet Flyby slot count, add optional Titan Flyby

billy54:
Currently, completion of Outer Planet Flyby requires flyby of at least 4/5 of Jupiter, Saturn, Uranus, Neptune, Pluto, with Charon as an optional. This makes little sense to me, as I have not heard of a combination of trajectories which would flyby only 3 of the gas giants plus Pluto as opposed to just the 4 gas giants as Voyager 2 did. IRL Voyager 1 sacrificed the Pluto flyby opportunity to study Titan instead, and the periapsis (6490km) was well within the 20Mm limit required by the Titan Flyby contract in RP-1.

This PR adds Titan to the list of required completions and require 5 out of 6, which means the player would probably do the 4 gas giants and then flyby either Titan or Pluto. I still think it is weird and I would ideally like to make the gas giants all compulsory and just have an ANY block for either Pluto or Titan, with Charon remaining as optional, but I assume there is a reason it is written like this. Or we could make both Pluto and Titan optional and just require the 4 gas giants for program completion.

This has a slight impact on Saturn Observation since Titan Flyby is required there, but I assume most players will do Titan atmo probe instead of Saturn so the player would probably go to Titan again anyway? I don't think it's a major balance issue.

### DIFF
--- a/GameData/RP-1/Programs/Programs.cfg
+++ b/GameData/RP-1/Programs/Programs.cfg
@@ -1361,13 +1361,12 @@ RP0_PROGRAM
 	{
 		ATLEAST
 		{
-			count = 5
+			count = 4
 			complete_contract = flybyJupiter
 			complete_contract = flybySaturn
 			complete_contract = flybyUranus
 			complete_contract = flybyNeptune
 			complete_contract = flybyPluto
-			complete_contract = flybyTitan
 		}
 	}
 	
@@ -1380,6 +1379,7 @@ RP0_PROGRAM
 	OPTIONALS
 	{
 		flybyCharon = true
+		flybyTitan = true
 	}
 }
 
--- a/GameData/RP-1/Programs/Programs.cfg
+++ b/GameData/RP-1/Programs/Programs.cfg
@@ -1368,7 +1368,6 @@ RP0_PROGRAM
 			complete_contract = flybyNeptune
 			complete_contract = flybyPluto
 			complete_contract = flybyTitan
-
 		}
 	}
 	
--- a/GameData/RP-1/Programs/Programs.cfg
+++ b/GameData/RP-1/Programs/Programs.cfg
@@ -1350,7 +1350,7 @@ RP0_PROGRAM
 	repDeltaOnCompletePerYearEarly = 285
 	repPenaltyPerYearLate = 285
 	repToConfidence = 4
-	slots = 3
+	slots = 2
 
 	REQUIREMENTS
 	{
--- a/GameData/RP-1/Programs/Programs.cfg
+++ b/GameData/RP-1/Programs/Programs.cfg
@@ -1361,12 +1361,14 @@ RP0_PROGRAM
 	{
 		ATLEAST
 		{
-			count = 4
+			count = 5
 			complete_contract = flybyJupiter
 			complete_contract = flybySaturn
 			complete_contract = flybyUranus
 			complete_contract = flybyNeptune
 			complete_contract = flybyPluto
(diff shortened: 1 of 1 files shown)


CXG-2827:
This program does not force you to complete the flybys via a single mission performing gravity assists to hit them in a sequence. You can send individual missions to each of the gas giants separately and complete the program that way.

If you check out the folder for the contracts, you will also see that they are written to not require a "New Vessel" parameter. So if you launch a mission to Jupiter and then realize you can get a gravity assist to reach Saturn, you can just accept the contract and not be penalized with having to launch a new mission.

billy54:
I see. The way it is currently set up, you can complete the program in a single launch on the Voyager 2 trajectory. With my changes you have to launch a Voyager 1 to Titan or Pluto. (unless there exists a trajectory that allows Voyager 2 to hit Titan but I doubt it due to NASA not doing it) I think this is better otherwise the Voyager window is too OP compared to sending separate probes for each destination.

CXG-2827:
This program is one of the worst-paying for the late game (Funds per year, per slot). Yes, you can technically complete this program with a single launch, but now you are stuck with this low payout for 12 years while the probe makes its rounds.

billy54:
Then I think we should reduce the slot count. There are really no further expenses once the probes are launched. Maybe cut it down to 2 or even 1. So you can do other stuff while you wait for the probes to go around.

Glloyd:
I agree. Both funding-wise and how long this thing hangs around, there's no reason it should take up 3 slots.

WithCrow:
image
For reference, outer flyby sucks quite heavily. Reducing it to 2 slots makes it pay the same per slot per year as outer gas giants, which I would advocate for being the only change to it.

CXG-2827:
I would agree on the reduction to 2 slots. But I still think the Titan flyby addition should be reverted. Leave it to the Jupiter Program. This flyby program should only focus on the Outer Planets and not any of the moons.

billy54:
Hmm, but Voyager 1 IRL did perform a flyby of Titan. I suppose the player could do a contractless Titan flyby, but it feels weird to not reward following a path that was taken in real life.

*btw, Titan is a moon of Saturn, not Jupiter

CXG-2827:
> Hmm, but Voyager 1 IRL did perform a flyby of Titan. I suppose the player could do a contractless Titan flyby, but it feels weird to not reward following a path that was taken in real life.
> 
> *btw, Titan is a moon of Saturn, not Jupiter

Thanks for the correction.

Regarding making the programs follow IRL examples, they shouldn't be railroading you to mimic exactly what happened. Look at the small bodies flyby program for instance. Mercury, Vesta and Ceres are lumped together.

billy54:
Well, that is true. I still think there is no reason to not have a contract for something that happened in real life though... I can set Titan Flyby as purely optional, and revert the requirements to the original 4/5. Not sure what this means for the rep/confidence though, considering it is required for Saturn Observation.

What I could also do is make Pluto optional as well, and force a flyby of all four gas giants, but that might be a bit restrictive. I'm not sure why you would want to do 3 gas giants and Pluto, but there's probably some combination where it makes sense, like a Saturn flyby combined with a Jupiter -> Uranus -> Pluto or something.

billy54:
I am still open to any further suggestions regarding this PR, but if not then I don't have any more ideas, so if someone can review this then it can be merged. It's literally a single line addition since I reverted the Titan flyby requirement.

PR #2836, opened 2026-06-06
Add Advanced Mars Surface

WithCrow:
This PR adds a successor program to the Mars Surface Exploration program, which requires the following:

- A landing on Phobos
- A landing on Deimos
- A Mars Sample Return

The landing on phobos & deimos requires a new, more advanced probe to be sent with explicit new scientific experiments. The funding per year per slot is at 185k, as opposed to the base Mars Surface fo 133k. 

 It also keeps a few of the optionals from the predecessor programs, such as additional rovers and phobos + deimos landers.


Glloyd:
This is very appreciated. The absence of the old uncrewed mars sample return contract in P&LC was disappointing.

If you haven't done so already, the previous mars program has a phobos sample return, but not a deimos one. The latter could be a nice optional addition here, as could the return of the former if you didn't complete it the first time around.

WithCrow:
Huh! I never realised it didn't have a Deimos sample return. IDK why I didn't include either as optionals here - I will do so.
Happy to hear it's appreciated 😄

GameData/RP-1/Contracts/Mars/Phobos Sample Return.cfg
CXG-2827, line 98: Take a look at the Asteroid Exploration Revamp PR for the sample return parameter used. I may be wrong, but the one here may look for generic data, so even an untransmitted telemetry reading could be recovered and complete the "sample return". 

it needs 
experiment = surfaceSample

GameData/RP-1/Contracts/Mars/samplesMars.cfg
```diff
@@ -0,0 +1,94 @@
...
+			type = RP1ReturnHome
+			title = Return to Earth safely with the samples
+			hideChildren = true
+			completeInSequence = true
+		}
+	}
+	PARAMETER
+	{
+		name = SurfaceScience
+		type = CollectScience
```
CXG-2827, line 78: Same comment as Phobos

CXG-2827 (changes requested):

CXG-2827:
Would this program be a good opportunity to include contracts to test ISRU capabilities on mars or the moons?

Clayell:
These are uncrewed probes, so probably not?
I'm also not sure if ISRU is even unlocked by the expected time of this contract.

CXG-2827:
Since this program is requiring a mars sample return, I would see this more appropriate for a later tech period. Otherwise this just feels like an immediate follow-up to Mars Surface Exploration, which people might already be running in the early to mid 70s.

It might be better to align the experiments for this program with 80-90s tech.

regarding ISRU, I was thinking more as sending a demonstrator, like MOXIE

billy54:
Should we set this as a requirement for Crewed Mars? Makes sense that you would test the mission profile of a crewed mission with an uncrewed sample return first.

Glloyd:
I have to disagree with CXG here. I don't think it should be tech locked that late, especially if it becomes a Marsboots requirement. I think a lot of players are going to be finishing Mars Surface in the early-mid 1970s, it would make sense to pick this up immediately afterwards. If you're going to tech limit it at all, it shouldn't be later than the mid-late 1970s. That's the time period that the Soviet sample returns were planned for regardless, had the N1 not failed.

Glloyd:
That said, an optional contract for an ISRU demonstrator is actually a really interesting idea, either as part of this or as part of MarsBoots

CXG-2827:
> Should we set this as a requirement for Crewed Mars? Makes sense that you would test the mission profile of a crewed mission with an uncrewed sample return first.

Crewed Mars has the option to perform the program in a manner similar to Red Rocks/Mars Base Camp, which includes a mars sample return instead of sending crew to the surface. If this new Advanced Mars Program will now be a requirement for Crewed Mars, the Mars Basecamp method likely be removed entirely as the sample return will already be accomplished.

> If you're going to tech limit it at all, it shouldn't be later than the mid-late 1970s. That's the time period that the Soviet sample returns were planned for regardless, had the N1 not failed.

The current Mars Surface Exploration program is structured with a Mars landing and a rover. This to me feels like the Viking thru Sojourner era. The Advanced Mars Exploration program should have some more significant mission variety and technical challenge compared to the first program. Otherwise it feels too similar. The sample return is a step in the right direction, but I feel like the Deimos and Phobos contracts could have a bit more depth. 

If you look at the tech tree, Sample Return Science Experiments is 1986-1996 and is what I would think is more appropriate for this program. It could be an opportunity to have contracts to use the Surface Scanning Module and the Narrowband Scanner. Maybe even have a targeted landing contract to land on either of the poles

WithCrow:
I am not sure if requiring this program for Marsboots is the right choice - we landed on the Moon without first returning samples. 

I also like the targeted landing contracts. SSame for MOXIE - I think it could be a neat stepping stone.

GameData/RP-1/Contracts/Mars/Phobos Sample Return.cfg
WithCrow: I can't believe I didn't notice before - those contracts are from the OG program & merely optional here, thus it's somewhat out of scope for me to change that. I did, however, add the sample return parameter from Crewed Mars to the Mars sample return contract.

WithCrow:
Okay! I've finally gotten back to this & consider it ready for review! I've added a contract for landing on the north/south pole. While I did take a look at MOXIE - I think this is something a future PR will have to take a look at. There are other ISRU related experiments that could be very interesting, so at least both due to time & scope, I think it is outside of the realm of this PR.

PR #2837, opened 2026-06-06
Add Commercial Constellation Programs

WithCrow:
Whereas Skopos sees the player deploy a constellation of their own, these three programs focus entirely on deploying a customer's constellation with increasing requirements in terms of size and number. In order to allow players from all launch sites to (at least attempt to) complete this, there are three different types, either a LEO Walker, Molniya or GEO constellation. 

I do very much think that Skopos telecom *and* this can coexist peacefully and one should not attempt to replace the other. The reason for this is simple: not all players want to maintain a telecom constellation. Skopos is far more fun than these three programs, but these allow a player who does not want to maintain a constellation to have some "busy work" to do in the Earth system. 

That hits it on the head in my eyes: more stuff to do around Earth - of which there is far too little past the early game.

If someone knows how - I would prefer to restrict this to the 2000s at the earliest. While one could argue that ArianeGroup already saw commercial launches, they were not of this order of magnitude and they should thus not appear at an ahistorically early point in time.

### DIFF
--- a/GameData/RP-1/Contracts/Commercial Constellations/cc1_geo.cfg
+++ b/GameData/RP-1/Contracts/Commercial Constellations/cc1_geo.cfg
@@ -1,0 +1,177 @@
+CONTRACT_TYPE
+{
+    name = cc1_geo
+    title = Commercial GEO Constellation (Slot @index / 12)
+    group = CommConst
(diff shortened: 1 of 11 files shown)


WithCrow:
There are three main Earth focussed satellites:
- Earth Observation (covered by 2707)
- Telecom (Skopos + this)
- GNSS
A GNSS program that is just 'Deploy NavSat payload into MEO' would be quite identical in terms of gameplay to this one that I'll wait until Skopos GNSS is a thing.

periodically-makes-puns:
Some initial thoughts:
- No optional contracts, so currently these are acting as a confidence sink. The confidence cost is perhaps a bit high for a 1-slot Earth program? 
- Very high funding per slot-year! 375K for CC1 to 625K for CC3. That's honestly way too high imo, given that this program is almost entirely launch costs and requires basically no research.

WithCrow:
After much consideration, I've decided to close this (for now?). The contracts really are not that fun to fly, which is reason enough to just close this alltogether.

PR #2840, opened 2026-06-08
Group tooling types and add Refit button

chambm:
## Summary

Two-commit change to the Toolings tab in the editor:

1. **Group tank tooling types into category buttons** keyed by construction (Conventional / Isogrid / Balloon / Fuselage / Service Module / Shielded / Cryogenic / Other) and pressure class (HP variants get their own button where they exist). Buttons sorted alphabetically by title. Each leaf row tags the underlying material(s) supplied it. Lets a player search for an existing tooled diameter for a tank type without picking material first. `ProceduralAvionicsWindow` now also shows the avionics tooling abbreviation ("N3") next to the config name so it cross-references the grouped `Avionics-N` entries.

Before:
Screenshot 2026-06-03 170800

After:
Screenshot 2026-06-05 161536

Screenshot 2026-06-05 162333

2. **Refit button on each tooled (d, L) row** that writes the dims to whichever part has its PAW open in the editor, switching the part's RF tank type when the row only supplies materials the part isn't currently on. Disabled when the PAW target's tank type belongs to a different bucket than the one on screen -- refit only crosses materials within the same construction family. When the user opens a PAW on a tank, the viewer auto-switches to that tank's bucket (from the main bucket list or from inside another bucket page). Manual bucket picks suppress the auto-switch until the PAW target changes.
Screenshot 2026-06-07 222638

## Test plan

- [x] Tooling tab in editor -- categories render in alphabetical order, HP variants appear as separate buttons where data exists.
- [x] Open a PAW on each tank class (Conventional/Isogrid/Balloon/Fuselage/ServiceModule/Cryogenic) -- viewer jumps to the matching bucket from both the main list and from inside another bucket.
- [x] In a bucket, click Refit on a row -- part resizes to (d, L), and RF tank type switches if the row's materials don't include the current type.
- [x] PAW open on a tank in a non-matching bucket -- Refit greys out with the "switch to its bucket" tooltip.
- [x] No PAW open -- Refit greys out with "Open a part's PAW" tooltip.
- [x] Avionics-N bucket -- ProceduralAvionicsWindow shows the N-level abbreviation next to the config name.

### DIFF
--- a/Source/RP0/UI/ToolingGUI.cs
+++ b/Source/RP0/UI/ToolingGUI.cs
@@ -41,6 +41,48 @@ namespace RP0
         private string _mergedCacheKey;
         private int _mergedCacheGeneration = -1;
         private List<ToolingEntry> _mergedCache;
+
+        // Cached grouped-tooling button list, keyed on ToolingDatabase.Generation so the per-frame
+        // RenderToolingTab / RenderTypeTab passes don't re-walk ToolingDatabase.toolings.Keys and
+        // re-allocate the group dictionary every OnGUI call.
+        private int _groupedCacheGeneration = -1;
+        private List<KeyValuePair<string, List<string>>> _groupedCache;
+
+        // Category keys for tank groupings. Prefix "Tank-" is reused so GetParametersForToolingType
+        // routes them to the diameter/length parameter set the same way bare "Tank-*" DB keys would.
+        // Conventional / Isogrid / Balloon / Fuselage each have a non-HP and an HP variant; ServiceModule
+        // and Shielded have no HP. HP buttons only appear when at least one underlying type matched.
+        private const string TankConventional   = "Tank-Conventional";
+        private const string TankConventionalHP = "Tank-ConventionalHP";
+        private const string TankIsogrid        = "Tank-Isogrid";
+        private const string TankIsogridHP      = "Tank-IsogridHP";
+        private const string TankBalloon        = "Tank-Balloon";
+        private const string TankBalloonHP      = "Tank-BalloonHP";
+        private const string TankFuselage       = "Tank-Fuselage";
+        private const string TankFuselageHP     = "Tank-FuselageHP";
+        private const string TankServiceModule  = "Tank-ServiceModule";
+        private const string TankShieldedKey    = "Tank-Shielded";
+        private const string TankCryogenic      = "Tank-Cryogenic";
+        private const string TankOther          = "Tank-Other";
+
+        // Titles are prefixed "Tank (...)" so every tank bucket sorts together, and the HP / Non-HP
+        // qualifier sits first inside the parens so all HP variants cluster and all Non-HP variants
+        // cluster within that block (siimav review: HP and base families sorted non-adjacently).
+        private static readonly Dictionary<string, string> _groupTitles = new Dictionary<string, string>
+        {
+            { TankConventional,   "Tank (Non-HP Conventional)" },
+            { TankConventionalHP, "Tank (HP Conventional)"     },
+            { TankIsogrid,        "Tank (Non-HP Isogrid)"      },
+            { TankIsogridHP,      "Tank (HP Isogrid)"          },
+            { TankBalloon,        "Tank (Non-HP Balloon)"      },
+            { TankBalloonHP,      "Tank (HP Balloon)"          },
+            { TankFuselage,       "Tank (Non-HP Fuselage)"     },
+            { TankFuselageHP,     "Tank (HP Fuselage)"         },
+            { TankServiceModule,  "Tank (Service Module)"      },
+            { TankShieldedKey,    "Tank (Shielded)"            },
+            { TankCryogenic,      "Tank (Cryogenic)"           },
+            { TankOther,          "Tank (Other)"               },
+        };
         private bool _isToolingTempDisabled = false;
         private float _nextUpdate = 0f;
         private float _allTooledCost;
@@ -108,12 +150,12 @@ namespace RP0
             GUILayout.FlexibleSpace();
             GUILayout.EndHorizontal();
 
-            var grouped_list = GetGroupedToolingTypes().ToList();
-            TryAutoSwitchBucket(grouped_list);
+            List<KeyValuePair<string, List<string>>> groupedList = GetGroupedToolingTypesCached();
+            TryAutoSwitchBucket(groupedList);
 
             int counter = 0;
             GUILayout.BeginHorizontal();
-            foreach (var grouped in grouped_list)
+            foreach (KeyValuePair<string, List<string>> grouped in groupedList)
             {
                 if (counter % 3 == 0 && counter != 0)
                 {
@@ -207,38 +249,58 @@ namespace RP0
             return _currentToolingType == null ? UITab.Tooling : UITab.ToolingType;
         }
 
-        // Switches the current bucket to match whatever part has its PAW open, if that part's tank
-        // type maps to one of the buckets we know about. Only fires once per PAW target so a
-        // manual bucket pick survives until the user opens a different part's PAW.
-        private void TryAutoSwitchBucket(List<KeyValuePair<string, List<string>>> grouped_list)
+        /// <summary>
+        /// Switches the current bucket to match whatever part has its PAW open, if that part's tank
+        /// type maps to one of the buckets we know about. Only fires once per PAW target so a
+        /// manual bucket pick survives until the user opens a different part's PAW.
+        /// </summary>
+        private void TryAutoSwitchBucket(List<KeyValuePair<string, List<string>>> groupedList)
         {
-            var pawForAutoSwitch = ToolingPartResizer.PawTarget();
+            Part pawForAutoSwitch = ToolingPartResizer.PawTarget();
             if (pawForAutoSwitch == null) { _lastAutoSwitchPart = null; return; }
             if (pawForAutoSwitch == _lastAutoSwitchPart) return;
             _lastAutoSwitchPart = pawForAutoSwitch;
-            var pawType = ToolingPartResizer.CurrentTankType(pawForAutoSwitch);
+            string pawType = ToolingPartResizer.CurrentTankType(pawForAutoSwitch);
             if (string.IsNullOrEmpty(pawType)) return;
-            var bucketKey = GetGroupingKey(pawType);
-            var match = grouped_list.FirstOrDefault(g => g.Key == bucketKey);
+            string bucketKey = GetGroupingKey(pawType);
+            KeyValuePair<string, List<string>> match = groupedList.FirstOrDefault(g => g.Key == bucketKey);
             if (match.Value == null) return;
             _currentToolingType = bucketKey;
             _currentToolingTitle = GetGroupTitle(bucketKey);
             _currentUnderlyingTypes = match.Value;
         }
 
-        // Returns the ordered list of tooling-type buttons to show, paired with the DB keys each one covers.
-        // Avionics types are collapsed by category letter (all tech levels of "Avionics-N*" share one entry).
-        // Tank types are collapsed by construction (Stringer / Isogrid / Balloon / Fuselage / ServiceModule)
-        // so a player can search for an existing tooled diameter without picking material first. Non-tank,
-        // non-avionics types pass through unchanged.
-        private static IEnumerable<KeyValuePair<string, List<string>>> GetGroupedToolingTypes()
+        /// <summary>
+        /// Returns the cached grouped-tooling button list, rebuilding only when ToolingDatabase.Generation
+        /// changes. Both RenderToolingTab and RenderTypeTab call this every frame, so caching avoids
+        /// re-walking ToolingDatabase.toolings.Keys and re-allocating the group dictionary each pass.
+        /// </summary>
+        private List<KeyValuePair<string, List<string>>> GetGroupedToolingTypesCached()
+        {
+            int gen = ToolingDatabase.Generation;
+            if (_groupedCache == null || _groupedCacheGeneration != gen)
+            {
+                _groupedCache = GetGroupedToolingTypes();
+                _groupedCacheGeneration = gen;
+            }
+            return _groupedCache;
+        }
+
+        /// <summary>
+        /// Builds the ordered list of tooling-type buttons to show, paired with the DB keys each one covers.
+        /// Avionics types are collapsed by category letter (all tech levels of "Avionics-N*" share one entry).
+        /// Tank types are collapsed by construction (Conventional / Isogrid / Balloon / Fuselage / ServiceModule)
+        /// so a player can search for an existing tooled diameter without picking material first. Non-tank,
+        /// non-avionics types pass through unchanged.
+        /// </summary>
+        private static List<KeyValuePair<string, List<string>>> GetGroupedToolingTypes()
         {
             var groups = new Dictionary<string, List<string>>();
             var order = new List<string>();
             foreach (string type in ToolingDatabase.toolings.Keys)
             {
                 string key = GetGroupingKey(type);
-                if (!groups.TryGetValue(key, out var list))
+                if (!groups.TryGetValue(key, out List<string> list))
                 {
                     list = new List<string>();
                     groups[key] = list;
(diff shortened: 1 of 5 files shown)


siimav:
What is that mysterious SteelAl in your images?

Source/RP0/UI/ToolingGUI.cs
```diff
@@ -181,8 +207,143 @@ public UITab RenderToolingTab()
...
+            if (toolingType == "TankShielded") return TankShieldedKey;
+            if (toolingType == "Cryogenic" || toolingType == "BalloonCryo") return TankCryogenic;
+
+            bool hp = toolingType.EndsWith("-HP");
+            if (toolingType.StartsWith("Tank-Iso-"))                          return hp ? TankIsogridHP  : TankIsogrid;
+            if (toolingType.StartsWith("Tank-Balloon-") || toolingType == "Balloon")
+                                                                              return hp ? TankBalloonHP  : TankBalloon;
+            if (toolingType == "Fuselage")                                    return hp ? TankFuselageHP : TankFuselage;
+            if (toolingType.StartsWith("Tank-Sep-"))                          return hp ? TankConventionalHP : TankConventional;
+            return TankOther;  // legacy RF types we don't know how to place: Default, ElectricPropulsion, etc.
```
siimav, line 347: IDK why your clanker thought this is good formatting.

Source/RP0/UI/ToolingGUI.cs
```diff
@@ -181,8 +207,143 @@ public UITab RenderToolingTab()
...
+            { TankIsogrid,        "Isogrid Tanks"         },
+            { TankIsogridHP,      "HP Isogrid Tanks"      },
+            { TankBalloon,        "Balloon Tanks"         },
+            { TankBalloonHP,      "HP Balloon Tanks"      },
+            { TankFuselage,       "Fuselage"              },
+            { TankFuselageHP,     "HP Fuselage"           },
+            { TankServiceModule,  "Service Modules"       },
+            { TankShieldedKey,    "Shielded Tanks"        },
+            { TankCryogenic,      "Cryogenic Tanks"       },
+            { TankOther,          "Other Tanks"           },
```
siimav: HP and non-HP bucket pairs sort non-adjacently GetGroupedToolingTypes() sorts by GetGroupTitle(). "Conventional Tanks" and "HP Conventional Tanks" sort with several other buckets between them (alphabetically H follows C, I, etc.). Users looking for HP variants would expect them next to their base family. Consider naming them "Conventional Tanks (HP)" so they sort adjacent to the base.

Source/RP0/UI/ToolingGUI.cs
```diff
@@ -181,8 +207,143 @@ public UITab RenderToolingTab()
...
+                    return false;
+            }
+        }
+
         public void RenderTypeTab()
         {
+            // Same auto-switch behaviour as RenderToolingTab: opening a PAW on a different tank
+            // jumps into that tank's bucket. Needed here because the user is already inside a
+            // bucket page (RenderToolingTab doesn't run), and otherwise the page is stuck.
+            TryAutoSwitchBucket(GetGroupedToolingTypes().ToList());
```
siimav: GetGroupedToolingTypes().ToList() called every frame in RenderTypeTab
TryAutoSwitchBucket(GetGroupedToolingTypes().ToList());
RenderToolingTab correctly caches this into grouped_list for its frame, but RenderTypeTab rebuilds it from scratch every frame. Since GetGroupedToolingTypes() iterates ToolingDatabase.toolings.Keys and allocates a Dictionary + List, this is unnecessary churn. Consider caching with Generation the same way GetEntriesForCurrentBucket() does, or reusing the cached list.

Source/RP0/UI/ToolingGUI.cs
```diff
@@ -82,27 +101,34 @@ public UITab RenderToolingTab()
...
 
             _currentToolingType = _currentToolingTitle = null;
+            _currentUnderlyingTypes = null;
             GUILayout.BeginHorizontal();
             GUILayout.FlexibleSpace();
             GUILayout.Label("Tooling Types", HighLogic.Skin.label);
             GUILayout.FlexibleSpace();
             GUILayout.EndHorizontal();
 
+            var grouped_list = GetGroupedToolingTypes().ToList();
```
siimav: Naming inconsistency: grouped_list uses underscores while the rest of the file uses camelCase (pawTarget, bucketKey, etc.). Should be groupedList.

Source/RP0/UI/ToolingGUI.cs
```diff
@@ -181,8 +207,143 @@ public UITab RenderToolingTab()
...
+            _currentToolingTitle = GetGroupTitle(bucketKey);
+            _currentUnderlyingTypes = match.Value;
+        }
+
+        // Returns the ordered list of tooling-type buttons to show, paired with the DB keys each one covers.
+        // Avionics types are collapsed by category letter (all tech levels of "Avionics-N*" share one entry).
+        // Tank types are collapsed by construction (Stringer / Isogrid / Balloon / Fuselage / ServiceModule)
+        // so a player can search for an existing tooled diameter without picking material first. Non-tank,
+        // non-avionics types pass through unchanged.
+        private static IEnumerable<KeyValuePair<string, List<string>>> GetGroupedToolingTypes()
```
siimav: GetGroupedToolingTypes return type: The method returns IEnumerable but all callers immediately call .ToList().
Returning List<> directly would remove the allocation at each call site and make the intent explicit.

Source/RP0/UI/ToolingPartResizer.cs
```diff
@@ -0,0 +1,217 @@
...
+                "Truss"                  => "ProceduralShapeHollowTruss",
+                _                        => "ProceduralShapeCylinder",
+            };
+            return p.Modules.Cast<PartModule>().FirstOrDefault(m => m.moduleName == moduleName);
+        }
+
+        // BaseField.SetValue writes the underlying field, but some KSP versions don't fire the
+        // UI_Control.onFieldChanged callback that shape modules listen on; fire it explicitly with
+        // the prior value so the shape rebuilds even when there's no PAW interaction.
+        static void SetField(PartModule m, string name, float value)
```
siimav: SetField silently swallows all exceptions: The outer try { ... } catch { } with no logging means a field that fails
to set produces no observable feedback during development. At least a Debug.LogWarning on failure would help diagnose shape issues.

Source/RP0/UI/ToolingPartResizer.cs
```diff
@@ -0,0 +1,217 @@
...
+            {
+                var f = m.Fields[name];
+                if (f == null) return 0f;
+                var v = f.GetValue(m);
+                return v is float fv ? fv : Convert.ToSingle(v);
+            }
+            catch { return 0f; }
+        }
+
+        static bool TryInvoke(object o, string method, object[] args)
```
siimav: TryInvoke with bare catch { return false; }: Same concern — silently eating exceptions from UpdateTankType/ChangeTankType. A Debug.LogWarning would surface RF API changes without breaking the player.

Source/RP0/UI/ToolingPartResizer.cs
```diff
@@ -0,0 +1,217 @@
...
+                }
+                Apply(p, diameter, length);
+                Msg(typeChanged
+                    ? $"Refit {p.partInfo?.title} to {targetRfType} at d={diameter:F3}m, L={length:F3}m"
+                    : $"Resized {p.partInfo?.title} to d={diameter:F3}m, L={length:F3}m");
+            }
+            catch (Exception ex) { Debug.LogError("[RP0 Tooling] resize failed: " + ex); Msg("Resize failed: " + ex.Message); }
+        }
+
+        static void Apply(Part p, float diameter, float length)
```
siimav: No access modifiers on all the methods below

siimav (changes requested):
General notes:
* Use proper /// method comments
* Do not use var unless the type is apparent from the right side of the assignment.
* Do not mix fields between methods

IDK what to think of the "Refit" feature. Sounds like there's so many corner cases that can force the part into an invalid state. For instance it doesn't appear to handle avionics parameters at all so the outcome will be a weird mix. Same for PP shapes. PickRfType will silently swallow a correct type not being found and will happily use the first one available.

Source/RP0/UI/ToolingGUI.cs
chambm: That's intentional. I think users first decide "Do I need HP or non-HP?" and then they want to see if they have any tanks already tooled for the dimensions they need. They're likely to be flexible about whether it's balloon/conventional/isogrid and even more flexible about the material of the tank.

If tank type (balloon/conventional/isogrid) was a slider instead of part of the type, I'd probably just have two buckets: HP and non-HP (perhaps as a user option for users who aren't so flexible about the tank type, e.g. to keep that isogrid integration bonus). The Refit button would then change tank type as well as material. That would be a considerably more convenient than the current situation where you have to take the rocket apart and put it back together to change the tank type. Outside this PR's scope, of course.

Source/RP0/UI/ToolingGUI.cs
chambm: I agree it could be improved, but I'm not sure which part you're objecting to. That (only) half of the return statements are manually indented? That all but one are single line ifs? Looking at it again, I will make predicates for each tank type (`IsBalloonTank`, `IsShieldedTank`, `IsCryoTank`) so each line can be a single line. At that point it could change to a switch statement or expression.

Source/RP0/UI/ToolingGUI.cs
chambm: The clanker was way better than me at optimizing for performance, but it missed this case. Thanks for catching it.

Source/RP0/UI/ToolingGUI.cs
chambm: Oh, but it is weird that the HP types sort together but the non-HP types do not. Add explicit "Non-HP" prefix so they sort together? That would look clunky but it would maintain my intention and alphabetical listing.

chambm:
> IDK what to think of the "Refit" feature. Sounds like there's so many corner cases that can force the part into an invalid state. For instance it doesn't appear to handle avionics parameters at all so the outcome will be a weird mix. Same for PP shapes. PickRfType will silently swallow a correct type not being found and will happily use the first one available.

I'm not sure what you mean about avionics parameters. The Refit button is only for fuel tanks right now. It would be nice to apply it to avionics parts too but like you say, the controllable mass and EC parameters would have to be handled additionally. Those don't apply for fuel tanks though, right?

Different PP shapes do present a bit of an issue when they have more than the basic "diameter/length" or "diameter/vscale" sliders. But since RP-1 tooling doesn't distinguish between them (AFAIK?), it seems reasonable to just have Refit handle those diameter and length parameters and leave the other parameters alone. Certainly this handles the bulk of cases with cylindrical/cone/cyl-deriv/cone-deriv. I do need to test that it's working as I expect with other shapes though, I think I only tested switching between cylinders of each tank type and material (tank types being different parts, as mentioned above).

Source/RP0/UI/ToolingGUI.cs
chambm: How about `Tank (HP Conventional)` `Tank (HP Fuselage)` `Tank (HP Isogrid)` `Tank (HP Shielded)` `Tank (Non-HP Balloon)` `Tank (Non-HP Conventional)` `Tank (Non-HP Fuselage)` `Tank (Non-HP Isogrid)` `Tank (Non-HP Shielded)`

Better than a clunky Non-HP prefix at the beginning, with the added advantage that all the Tanks sort together.

chambm:
> What is that mysterious SteelAl in your images?

 - Tank-Balloon-SteelAl → title "Balloon Tank"
 - Tank-Balloon-SteelAlCu → "Refined Balloon Tank"
 - Tank-Balloon-SteelAlLi → "Modern Balloon Tank"

Shall I change it to "Balloon", "Refined", and "Modern"? Or "Al", "AlCu", "AlLi"?

Source/RP0/UI/ToolingGUI.cs
chambm: image

chambm:
> > IDK what to think of the "Refit" feature. Sounds like there's so many corner cases that can force the part into an invalid state. For instance it doesn't appear to handle avionics parameters at all so the outcome will be a weird mix. Same for PP shapes. PickRfType will silently swallow a correct type not being found and will happily use the first one available.
> 
> I'm not sure what you mean about avionics parameters. The Refit button is only for fuel tanks right now. It would be nice to apply it to avionics parts too but like you say, the controllable mass and EC parameters would have to be handled additionally. Those don't apply for fuel tanks though, right?
> 
> Different PP shapes do present a bit of an issue when they have more than the basic "diameter/length" or "diameter/vscale" sliders. But since RP-1 tooling doesn't distinguish between them (AFAIK?), it seems reasonable to just have Refit handle those diameter and length parameters and leave the other parameters alone. Certainly this handles the bulk of cases with cylindrical/cone/cyl-deriv/cone-deriv. I do need to test that it's working as I expect with other shapes though, I think I only tested switching between cylinders of each tank type and material (tank types being different parts, as mentioned above).

Definitely found some issues here with different shapes and symmetry, will address them soon. Thanks for the review.

chambm:
OK I've tested pretty thoroughly with the main tank types (modular, another modular, procedural), and with changing shapes with procedural tanks. Refit works well with all of them. It's easier to test with changing shapes with the PR I just put in to preserve diameter/length when changing procedural shape. That way you can see that Refit even applies to cones (treated as cylinders), hollow/filleted variants, and polygon, Refit on truss is currently disabled although I could revisit that and just have it handle top/bottom diameter and length the same way it does for other shapes with both top and bottom diameter.

siimav:
> Shall I change it to "Balloon", "Refined", and "Modern"? Or "Al", "AlCu", "AlLi"?

Well we have to follow what is shown on the tank part itself or it's gonna get hella confusing.

Source/RP0/UI/ToolingGUI.cs
siimav: Not clue what it's even aligned on. Just half of the code dangling pretty much nowhere.

chambm: Are you looking at the latest commit that uses predicates for each if statement?

siimav: No, I haven't looked at the new commits. Too bad GH makes the process a pain.

chambm: GH usually marks the commented snippet as outdated (like the below comment) if it detects a later commit has changed it. I suppose I changed this section so much it couldn't even recognize it as being related.

Source/RP0/UI/ToolingGUI.cs
```diff
@@ -22,6 +23,66 @@ private struct UntooledPart
...
+        // qualifier sits first inside the parens so all HP variants cluster and all Non-HP variants
+        // cluster within that block (siimav review: HP and base families sorted non-adjacently).
+        private static readonly Dictionary<string, string> _groupTitles = new Dictionary<string, string>
+        {
+            { TankConventional,   "Tank (Non-HP Conventional)" },
+            { TankConventionalHP, "Tank (HP Conventional)"     },
+            { TankIsogrid,        "Tank (Non-HP Isogrid)"      },
+            { TankIsogridHP,      "Tank (HP Isogrid)"          },
+            { TankBalloon,        "Tank (Non-HP Balloon)"      },
+            { TankBalloonHP,      "Tank (HP Balloon)"          },
```
siimav: HP balloons aren't a thing

Source/RP0/UI/ToolingPartResizer.cs
```diff
@@ -0,0 +1,364 @@
...
+                "Cone"                   => "ProceduralShapeCone",
+                "Fillet Cylinder"        => "ProceduralShapePill",
+                "Polygon"                => "ProceduralShapePolygon",
+                "Hollow Cylinder"        => "ProceduralShapeHollowCylinder",
+                "Hollow Cone"            => "ProceduralShapeHollowCone",
+                "Hollow Fillet Cylinder" => "ProceduralShapeHollowPill",
+                "Truss"                  => "ProceduralShapeHollowTruss",
+                _                        => "ProceduralShapeCylinder",
+            };
+            return p.Modules.Cast<PartModule>().FirstOrDefault(m => m.moduleName == moduleName);
```
siimav: `GetModule` exists

chambm: Fixed.

Source/RP0/UI/ToolingGUI.cs
chambm: Fixed.

chambm:
One issue I found is Shielded tanks are shown as "Structural". That's probably something that should be fixed in the cfgs, yeah?

siimav:
Shielded tanks are using legacy tank definitions. So Structural and Fuselage are expected there.

PR #2843, opened 2026-06-08
Update and harmonize tags on SSPX parts

Glloyd:
This updates all of the human-rated SSPX parts to include the NoResourceCostMult to bring them in line with other non-SSPX station parts. I also added the Instruments tag to the science labs, again to bring them in line with existing parts.

That said, there's a number of SSPX parts that have habitat volume, but don't have the Human Rated tag. I'd also be willing to add that tag to all of those if that's something that the maintainers believe should be done.


siimav:
Yes, I believe all parts with habitable volume should come with HumanRated tag.

Glloyd:
Done 👍 

All parts that have habitable volume now have the Human-Rated tag.

PR #2845, opened 2026-06-10
Crewed Mercury exploration program

gabo463:
this add a crewed Mercury exploration program after  the up and coming Venus reactivations and programs 

(this is my first pr please tell me if there is any issue)


billy54:
I am not sure why anyone would ever want to do this in real life, but I guess we can have it for completeness?

gabo463:
i believe in real life a mission to mercury would be primarily for its resources and for scientific purposes.  yes i will admit that it act as a completion of the inner planets  but mercury remain an interesting place to visit whit a crew

this program is also quite short only the flyby and the orbit a required and the Mercurian landing is optional

hayghostin:
> I am not sure why anyone would ever want to do this in real life, but I guess we can have it for completeness?

For fun!!!

gabo463:
frogot to add contract groups

PR #2852, opened 2026-06-15
Show warnings in Editor GUI when the vessel cannot be integrated

dlesbre:
The editor KCT GUI had a single warning for human-rated vessels in non human-rated LC. This PR expands that to all vessel checks (mass limits, size limits, resources, clamp in hangar), and makes the label yellow. To avoid issues with long resource lists, the warnings list is cut after 5 elements:
imageimage

I've spliced into the VesselProject.MeetsFacilityRequirements to avoid recoding the checks. However, since the GUI is quite narrow, I wanted shorter error messages to avoid line breaks. So I added an extra shortReasons parameter, which returns alternate strings.

Full disclaimer: while I have fairly extensive programming experience, this is my first time with C#. Let me know if what I'm missing some good coding practices.

### DIFF
--- a/Source/RP0/SpaceCenter/Projects/VesselProject.cs
+++ b/Source/RP0/SpaceCenter/Projects/VesselProject.cs
@@ -804,6 +804,8 @@ namespace RP0
             if (ShipSize.sqrMagnitude > 0)
                 return ShipSize;
 
+            if (ShipNodeCompressed.Node is null) // true when first loading VAB
+                return Vector3.zero;
             ShipTemplate template = new ShipTemplate();
             template.LoadShip(ShipNodeCompressed.Node);
             ShipSize = template.GetShipSize();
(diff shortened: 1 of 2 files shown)


siimav (changes requested):
* How does this behave with empty editor? I.e when you load the scene and there's no vessel loaded or when you reset the editor. No warnings should be shown then.
* Many cases where the brace style is wrong.
* When editorFacility and ActiveLC.LCType mismatch (e.g. editing a plane in the SPH
while a Pad LC is still active), the warning evaluates plane mass/size against the
wrong LC type and can show misleading messages instead of the real "switch to Hangar" guidance
* Perhaps "…and {count - 4} more" reads better
* Lots of unrelated whitespace changes. Best not to autoformat the entire file or at least double check what you stage.

dlesbre:
I've pushed a new version with should fix those issues:
- Added a check to only enable warnings if the ship part count is above 0
- Added a warning for wrong LC type (Hangar vs VAB), which supersedes all others. I'm not quite sure how that is triggered though, If I use the switch editor button, I see the warning temporarily but the scene quickly switches back to the correct LC
- I've corrected the brace style, I'm not sure if you also have a specific style in mind for the three line ternary operators I added, so I've left them as is
- Sorry about the whitespace changes, my editor autoformats and hides them in the diffs

PR #2853, opened 2026-06-16
[Staff GUI] Fix time estimates

dlesbre:
As reported by [Krex on discord](https://discord.com/channels/319857228905447436/620690446540341261/1516287211106996315), the staffing GUI and space center management GUI show different time estimates to next project completion. This occurs both for engineers (the time to next vessel completion for each LC disagree), and to a lesser extent, for researchers. In both case, the SCM GUI has the correct value, and the staffing values change gradually over time.

In my save I saw the discrepancy for engineers but not for researchers:
Screenshot_20260616_210448

The issue is that in both case, the staff GUI uses different function for time estimates:
- For engineers, it uses a flat `remaining_BP / build_rate`, which fails to account for LC efficiency gain
- For scientists, it uses `Formula.GetResearchRate`, while SCM uses `ResearchProject.BuildRate`. The latter calls `Formula.GetResearchRate` but adds three multipliers: `YearBasedRateMult`, `CurrencyUtils.Rate(TransactionReasonsRP0.RateResearch)` and `Leaders.LeaderUtils.GetResearchRateEffect(nodeType, techID)`

This PR modifies the staff GUI so that it reports the same numbers as the space center management, as I doubt the differing displays was intended behavior.

### DIFF
--- a/Source/RP0/UI/KCT/GUI_Personnel.cs
+++ b/Source/RP0/UI/KCT/GUI_Personnel.cs
@@ -122,7 +122,7 @@ namespace RP0
 
             GUILayout.BeginHorizontal();
             GUILayout.Label("Assigned:");
-            
+
             string assignStr = GetAssignText(true, currentLC, out int assignAmt);
             string unassignStr = GetAssignText(false, currentLC, out int unassignAmt);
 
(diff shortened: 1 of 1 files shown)


siimav:
I believe the idea there is that you can use the modifier keys and for instance see how much +100 engineers or scientists will affect the end time.

dlesbre:
I never considered that. Its true that by hovering over the `+1/10/100` button in the engineers window the times and build rate values get updated (with the time no longer updated in my PR). Similarly, in the research tab, hovering over the `hire` or `fire` buttons updates research rates and times.

It a neat feature and I'll try to amend my PR to preserve it. I also have a few ways in which I believe it could be improved:
- Engineer figures don't update if you hover over the hire/fire buttons, only the `+`/`-` button. Which means you only see updated estimates if you already have the spare engineers on stand-by. I believe that limits its usefulness (in that case you can freely move them in and out of the LC as needed). Whats harder is figuring how many more you need to hire.
- Researchers should probably show not only the next tech, but maybe also the last tech in the queue. In my current game (early 1968 with 2.5k researchers), I rarely see changes significant enough for the next tech research date to change

I don't know when I'll have time to work on this, setting to draft in the mean time

PR #2863, opened 2026-06-27
Improve display of leader unlocks on contracts with a minimum count

### LINKED ISSUE #2304: [3.2.1.0] Flight Director - Message by 15km & 20km
Flight Director should be unlocked at 25km or Survivability Tech.
image

It hasn't been activated, but you get these messages at 15km and 20km.
image  
image

Clayell:
Fix https://github.com/KSP-RO/RP-1/issues/2304

The text to show which leaders will be unlocked by a contract will only be shown if you are one before meeting the minCount completions of that contract to unlock the leader

Example (flight directors are unlocked when completing the 25km contract):

image
image
image
image

### DIFF
--- a/Source/RP0/Requirements/Requirements.cs
+++ b/Source/RP0/Requirements/Requirements.cs
@@ -31,62 +31,26 @@ namespace RP0.Requirements
 
         public uint? MinCount { get; set; }
 
-        public override bool IsMet
-        {
-            get
-            {
-                bool isMet;
-                if (MinCount.HasValue && MinCount > 1)
-                {
-                    int c = ConfiguredContract.CompletedContractsByName.TryGetValue(ContractName, out var list) ? list.Count : 0;
-                    isMet = c >= MinCount;
-                }
-                else
-                {
-                    isMet = ConfiguredContract.CompletedContractsByName.TryGetValue(ContractName, out var list) && list.Count > 0;
-                }
+        private bool HasMinCount => MinCount.HasValue && MinCount > 1;
 
-                return IsInverted ? !isMet : isMet;
-            }
-        }
+        private int CompletedCount => ConfiguredContract.CompletedContractsByName.TryGetValue(ContractName, out var list) ? list.Count : 0;
 
-        public bool IsStrictlyMet
-        {
-            get
-            {
-                bool isStrictlyMet;
-                if (MinCount.HasValue && MinCount > 1)
-                {
-                    int c = ConfiguredContract.CompletedContractsByName.TryGetValue(ContractName, out var list) ? list.Count : 0;
-                    isStrictlyMet = c == MinCount;
-                }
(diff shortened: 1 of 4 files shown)


Clayell:
Need to confirm that this doesn't mess up the leader notification after completing the contract

Clayell:
(the leader unlocks are intentionally not shown here and instead reserved for the special leader notification, which matches with past behavior)
image
image
image

Here's the bug I mentioned in the TODO comment, where the unlocks will show up in the archives for the 15km and 20km contracts. This only happens after you have completed the 25km contract, and goes away once you complete the 30km contract (along with the leader unlock on the 25km contract archive):
image
image

I'm fine with leaving these issues as is, since I am sure that most people do not look in the contract archives window, nor would they notice this.

Source/RP0/Requirements/Requirements.cs
```diff
@@ -50,6 +50,44 @@ public override bool IsMet
...
+            {
+                bool isAlmostMet;
+                if (MinCount.HasValue && MinCount > 1)
+                {
+                    int c = ConfiguredContract.CompletedContractsByName.TryGetValue(ContractName, out var list) ? list.Count : 0;
+                    isAlmostMet = c == MinCount - 1;
+                }
+                else
+                {
+                    isAlmostMet = false;
```
siimav: This hardcodes false for the non-MinCount path. So single-completion leader contracts lose their preview text.

siimav (changes requested):
Not a fan of the duplication. Something along those lines might make more sense.
public uint? MinCount { get; set; }

private bool HasMinCount => MinCount.HasValue && MinCount > 1;

private int CompletedCount =>
    ConfiguredContract.CompletedContractsByName.TryGetValue(ContractName, out var list) ? list.Count : 0;

private bool ApplyInversion(bool met) => IsInverted ? !met : met;

// completed enough times
public override bool IsMet =>
    ApplyInversion(HasMinCount ? CompletedCount >= MinCount : CompletedCount > 0);

// exactly at the threshold (just crossed it on this completion)
public bool IsStrictlyMet =>
    ApplyInversion(HasMinCount ? CompletedCount == MinCount : CompletedCount > 0);

// one completion away from the threshold
public bool IsAlmostMet =>
    ApplyInversion(HasMinCount ? CompletedCount == MinCount - 1 : CompletedCount == 0);

Clayell:
Sounds good, I implemented those suggestions.

Although, for IsStrictlyMet, wouldn't it make more sense to have this:
`public bool IsStrictlyMet => ApplyInversion(HasMinCount ? CompletedCount == MinCount : CompletedCount == 1);`

instead of this:
`public bool IsStrictlyMet => ApplyInversion(HasMinCount ? CompletedCount == MinCount : CompletedCount > 0);`

PR #2866, opened 2026-06-28
Clean up contract files

siimav:


GameData/RP-1/Contracts/Outer Gas Giant Survey/Ariel Flyby.cfg
```diff
@@ -50,7 +50,6 @@ CONTRACT_TYPE
 		name = VesselGroup
 		type = VesselParameterGroup
 		title = Flyby Ariel
-		define = FlybyAriel
```
Clayell, line 53: Were lines like these intended to be removed?

Clayell: Nevermind, just saw the commit that shows it is intended.

PR #2867, opened 2026-06-29
Add Crewed Venus Program

gabo463:
this adds a crewed Venus program  as a first  crewed venture into interplanetary space i set it after the EarlyEarthSpaceStation program and the EarlyLunarHabitation program 

this program / pr contains 

crewed fly-by of venus contract
crewed orbit of venus contract 
crewed venus orbital  stays of 4-8-16 months 

(im still new at tthe pr making so please give feedback if anyting)


WithCrow:
I don't see the relevance of a robotic rover on Venus in this program, especially since I don't think that is possible with the available parts

gabo463:
the decision of reactivating the rover contract was probably made by me in the pursuit of some sort of surface activity since a crewed landing is impossible  and the lander/ landings find themselves in another program   I definitely do think rovers are not the best option tho

Clayell:
> crewed landing is impossible

TDChannel [has done a crewed venus landing in RP-1](https://www.youtube.com/watch?v=CXBk0xWtwK8), although he made use of non-RO blimps. Perhaps in the future these could be given RO/RP-1 configs and then a crewed landing would actually be feasible. (it also matches NASA proposals for a crewed landing, which involve balloons)

siimav:
That mod isn't configurable with MM.

gabo463:
> > crewed landing is impossible
> 
> TDChannel [has done a crewed venus landing in RP-1](https://www.youtube.com/watch?v=CXBk0xWtwK8), although he made use of non-RO blimps. Perhaps in the future these could be given RO/RP-1 configs and then a crewed landing would actually be feasible. (it also matches NASA proposals for a crewed landing, which involve balloons)

 I do want to investigate the kerbaloon mod when I’m in better shape cause I think it be beneficial for future solar exploration programs

gabo463:
Using #2892 completions of dives contract is made easier now giving 2 methods of completion worked so well that a contract of an atmospheric stay can be envisaged early testing showed a stay of up to 40 days in Venus atmosphere. As for the landings I haven’t tested anything other than potential samples return with the balloon’s that gave a variation of results. Crewed landing has not been considered yet as I believe balloons wouldn’t be enough. And why would you land on hell ???

WithCrow:
How does the 40 day stay work? Can you time warp? How do you enter & leave?

gabo463:
after reaching the desired altitude in the atmosphere stay you can indeed time wrap trough the stay no problem  although the game itself is inconsistent if you leave the craft sometimes its stays there, sometimes you lose altitude which can result in crew loss if to grave and other time everything is fine

WithCrow:
Sounds far too risky IMO. Please experiment & show the altitude variance, etc. here

gabo463:
Will do im also envisaging testing whit airships to give the player a 3 method of atmospheric exploration and to a potential landing

gabo463:
i cant verify this yet as I'm working on other stuff  but it seams from another persons test  that the balloon does and can stay in a stable  state  and you can leave the craft to do other mission or stuff during the stay

gabo463:
Lost sleep for this but the airship does work adds an additional 3 methods of atmospheric activity the latter being the best as i extended the stay to 80 day( not the limit) I also believe these can be useful for potential crewed Titan activities.  

On a side note i really don’t know where to put the air ships in the tree for balancing reasons since the first crewed airship where in 1852 and if I go the helium route it’s in 1960

gabo463:
increased the funding from 15M to 25m  to take account of the extra activities brought by the program.

WithCrow:
Why? They are just optional contracts. A 10 million increase for optionals is not fit IMO

gabo463:
after review i think your right this is to much of an increase im thinking 17 million or 20 million would be the sweat spot

gabo463:
settled for 18 million

gabo463:
Balloon stay test :
53 km 40 days 
60 km 60 days

gabo463:
now using  #2900  43 km for 70 days

gabo463:
After test and added method of doing an atmospheric dive and upcoming stay contract i found that blimp/ airships are the best way of doing them allowing much deeper atmospheric dive coming to a penultimate 30 km(before i got scared and pulled up)  balloons are also great for upper atmosphere exploration allowing for stays in high Venus atmosphere (50 to 60 km)plane are all dependent on how the player uses them and the purpose they going. now is a Venus landing possible ? i think with all those added methods  and a small care for kerbals life  yes it could be i haven't tested it myself but i assume it could.

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -1849,36 +1849,49 @@ RP0_PROGRAM
...
+	  VenusStationStay15months = true 
+	  VenusAtmosphericDive5m = true
+	  VenusAtmosphericDive10m = true 
+	  VenusAtmosphericDive20m = true
+	  roverVenus = true 
+
 	}
+	CONFIDENCECOSTS
+	{
+		Normal = 45000
```
WithCrow: Holy moly I did not even see this until now. You are off by an order of magnitude. This should be 4.5k not 45k

GameData/RP-1/Contracts/Crewed Venus/VenusSpaceStation.cfg
```diff
@@ -0,0 +1,203 @@
...
+    {
+        name = CompleteContract
+        type = CompleteContract
+        contractType = EarlyEarthSpaceStation
+    }
+    REQUIREMENT
+    {
+        name = CompleteContract
+        type = CompleteContract
+        contractType = CrewedLunarOrbitRepeat
```
WithCrow: Senseless requirement, you have ELH as a program requirement

GameData/RP-1/Contracts/Crewed Venus/VenusSpaceStation.cfg
```diff
@@ -0,0 +1,203 @@
...
+    {
+        name = ProgramActive
+        type = ProgramActive
+        program = CrewedVenusExp
+    }
+    REQUIREMENT
+    {
+        name = CompleteContract
+        type = CompleteContract
+        contractType = EarlyEarthSpaceStation
```
WithCrow: Senseless requirement, you have ESS as a program requirement

GameData/RP-1/Contracts/Crewed Venus/VenusSpaceStation.cfg
```diff
@@ -0,0 +1,203 @@
+CCONTRACT_TYPE
+{
+    name = VenusSpaceStation
+    title = Venus Space Station
+    group = CrewedVenusExp
+    agent = Stations
+    description = <b>Program: Crewed Venus Exploration <br>Type: <color=green>Required</color></b><br><br>Design, build, and launch a space station into orbit of Venus, crew it for 4 months with at least four crew, and bring the crew safely home. Once you're done, youll get missions to perform further crew rotations and resupply missions, as well as expansion or replacement missions.<br><br>Don't forget that the transfer window to Venus is often around 2 years. Make sure you bring enough Life Support to last long enough for a resupply mission to get there.
```
WithCrow: Why is this required? I am not aware of the AAP planning on sending a station there. Especially since this means you need to do two trips.

GameData/RP-1/Contracts/Crewed Venus/VenusStationStay15m.cfg
```diff
@@ -0,0 +1,83 @@
...
+        type = CompleteContract
+        contractType = VenusStationStay8months
+    }
+
+ // ************ PARAMETERS ************
+        PARAMETER
+    {
+        name = stayOnStation
+        type = VesselParameterGroup
+        title = Keep at least 4 crew aboard the station for 8 months.
```
WithCrow: Incorrect title

GameData/RP-1/Contracts/Crewed Venus/VenusStationStay15m.cfg
```diff
@@ -0,0 +1,83 @@
+CONTRACT_TYPE
+{
+    name = VenusStationStay15months
+    title =  venus station 15 month stay
+    group = CrewedVenusExp
+    agent = Stations
+    description = <b>Program: Crewed Venus Exploration <br>Type: <color=blue>Optional</color></b><br><br> Maintain a continuous presence in orbit for 8 months (240 days). The crew must endure deep-space radiation, strict life support limitations, and complex thermal management. Bring them home safe.
```
WithCrow: Incorrect description

GameData/RP-1/Contracts/Crewed Venus/VenusFlybyCrew.cfg
```diff
@@ -0,0 +1,106 @@
...
+		name = ProgramActive
+		type = ProgramActive
+		program = CrewedVenusExp
+	}
+
+	REQUIREMENT
+	{
+		name = CompleteContract
+		type = CompleteContract
+		contractType = RepeatMoonLandingCrew
```
WithCrow: ELH as a program requirement makes this obsolete

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -1849,36 +1849,49 @@ RP0_PROGRAM
...
+	objectivesPrettyText = Send a crewed expedition to Venus orbit and eventually establish a permanent, long-duration research infrastructure around it
+	nominalDurationYears = 15
+	baseFunding = 18000000
+	repDeltaOnCompletePerYearEarly = 1500
+    repPenaltyPerYearLate = 1200
+    repToConfidence = 3
+    slots = 4
 
 	REQUIREMENTS
 	{
```
WithCrow, line 1869: Why is EIPP not a requirement? Why is Venus Surface Exploration not a requirement?

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -1849,36 +1849,49 @@ RP0_PROGRAM
...
-	nominalDurationYears = 10
-	baseFunding = 15000000
-	repDeltaOnCompletePerYearEarly = 1130
-	repPenaltyPerYearLate = 1130
-	repToConfidence = 3
-	slots = 4
+	title = Crewed Venus Exploration 
+	description = We have shown we can survive around and on the Moon. Now, we take the next monumental leap into the deep solar system: Venus. Searing heat, dense crushing atmosphere, and deep space radiation stand between our crew and the morning star.
+	requirementsPrettyText = Complete crewed operations around Earth's Moon and master early long-duration space stations.
+	objectivesPrettyText = Send a crewed expedition to Venus orbit and eventually establish a permanent, long-duration research infrastructure around it
```
WithCrow: This program is not about "establishing a permanent, long duration research infrastructure"

GameData/RP-1/Programs/Programs.cfg
periodically-makes-puns: 4.5k is probably too low for a program longer than Stations/Boots, but 45k is definitely way too high. I'd expect somewhere around 150-200 confidence/slot-year for Fast, which would put this at around 9000-12000 confidence for Fast and double that for Breakneck (for 60 slot-years).

gabo463:
i dont understand some of the discrepencies in contract have been fixed by me a long time ago how are they still present !!!

GameData/RP-1/Programs/Programs.cfg
WithCrow: At least according to my own balancing excel, with 4500, it is 267 funds per year per normal conf. cost, compared to the 250 of Crewed Lunar and 267 of ELH

GameData/RP-1/Programs/Programs.cfg
gabo463: this is my fault original vision of the program planned for some kind of venus hbitation program

GameData/RP-1/Contracts/Crewed Venus/VenusSpaceStation.cfg
gabo463: i put space station as required just for the reason the the program would had only 2 required contracts  and since its only a 4 month stay i thought and found that it fitted well

GameData/RP-1/Programs/Programs.cfg
gabo463: it was idk what happened

GameData/RP-1/Contracts/Crewed Venus/VenusSpaceStation.cfg
gabo463: and with the Atmospheric stay contract arriving i was playing with the idea of making at least one of the dives required

WithCrow: I cannot understate how much I dislike that. There is no serious proposal I am aware of that features multi-minute atmospheric dives of Venus.

GameData/RP-1/Programs/Programs.cfg
gabo463: i got 
8000 normal and 1200 fast

GameData/RP-1/Contracts/Crewed Venus/VenusSpaceStation.cfg
gabo463: consider it noted !

GameData/RP-1/Programs/Programs.cfg
WithCrow: image
Is what I am basing this on

gabo463: could i get a hold of this if it doesn't bother ?

gabo463:
Added the atmospheric stay now possible using #2900  and #2892  which now offer  3 different branch of contract for the player to complete   the player can after the 2 required contract are completed( fly-by and orbit) go the station route, the atmospheric dives route  or the atmospheric stay route

gabo463:
Hmm I’m starting to think thats way to much contract for no reason 

maybe removing the dives since the atmospheric stays are a longer version of them

gabo463:
After review and reflection i decided to keep them for multiple reasons one of them being that  I like the idea of the player choosing something quick to do rather than a full stay

gabo463:
Onto Venus!! space station are in my opinion to much of a constraint on the player this is easily fixable by making them into orbital stay  so there for 
removal of the station contract and its requirements 

Addition of orbital stay of 4-8-16 months repeatable twice

gabo463:
Change of structure for the program to be completed by a fly-by of Venus wich was the mission objective irl

gabo463:
removed all atmospheric dives as there was no true place for them with the arrival of the atmospheric stay

gabo463:
360 atmospheric stay is way to long  so it is removed

gabo463:
also sort keys where doing shenanigains aparantly that has been fixed

gabo463:
removal of the atmospheric stays to repetitive for a series of contract  so instead  1 repeatable atmospheric stay of 30 days up to 6 times

karbolenius:
Hi, @gabo463!

Just some thoughts here from me, as a player. 

1. The "Venus surface exploration" program can only be taken after "Early inner planet probes", so the requirements for the Crewed Venus Program can only include Venus surface exploration.

2. Currently, a player can complete the Crewed Venus Program by completing the "Flyby Venus (Crewed)" contract. However, the "Mars Crewed" program has an optional "MarsVenusFlybyCrew" contract. If a player takes both programs, he can potentially complete the Crewed Venus Program within the Mars program, spending virtually no resources, and earning 20 million for free. I don't know if is it good or not.

3. I think it would be better to retain the contracts from the "VenusAtmosphericDive" list, but implement logic similar to what we see in the "Earth Crewed" or "Spaceplanes" programs. In these programs, we have suborbital flight contracts that become unavailable once the player has accepted or comleted full orbital contracts. So, in the case of Venus, make it so that VenusAtmosphericDive contracts are available _until_ the VenusAtmosphericStay30d contract is accepted or completed. Then the player would have the choice of diving for a long time or just touching the atmosphere. This would allow for additional flexibility in the approach to this program, as "AtmosphericDive" can be achieved in a completely different way from "AtmosphericStay".

Also, I just completed the Mars Crewed program in my career save, so I'll try this Crewed Venus program and write if I find any bugs.

Sincerely, Daniil.

WithCrow:
I am of extreme opposition to any atmospheric dive/stay regardless. No agency would approve something so incredibly reckless for zero gain.

gabo463:
> Hi, @gabo463!
> 
> Just some thoughts here from me, as a player.
> 
> 1. The "Venus surface exploration" program can only be taken after "Early inner planet probes", so the requirements for the Crewed Venus Program can only include Venus surface exploration.
> 2. Currently, a player can complete the Crewed Venus Program by completing the "Flyby Venus (Crewed)" contract. However, the "Mars Crewed" program has an optional "MarsVenusFlybyCrew" contract. If a player takes both programs, he can potentially complete the Crewed Venus Program within the Mars program, spending virtually no resources, and earning 20 million for free. I don't know if is it good or not.

good catch on both of those issues thanks for that! as what concerns dives i decided to remove them because they were to similar to the atmospheric stays and it was getting repetitive to do them. it was also kinda of a riskier contract for noting since it required the player to fly in the atmosphere and the stays where that but longer. i also went whit the stays because of the  High Altitude Venus Operational Concept (HAVOC) wich Phase 3 involves astronauts descending into the atmosphere, for 30 Earth days and Phase 4 of the concept is to send humans into the atmosphere of Venus for 1 Earth year  (phase 4 that im in the process of adding).  structuring the program that way also opens the door to a potential Venus habitation program

gabo463:
> I am of extreme opposition to any atmospheric dive/stay regardless. No agency would approve something so incredibly reckless for zero gain.

I gotta disagree with this HAVOC (https://www.nasa.gov/general/havoc/ & https://ntrs.nasa.gov/citations/20160006329)  shows how it can be possible and achieved.
image


 id also argue that every crewed venture in the solar system is risky to some degree

WithCrow:
> I gotta disagree with this HAVOC (https://www.nasa.gov/general/havoc/ & https://ntrs.nasa.gov/citations/20160006329)  shows how it can be possible and achieved.
> image

No. It shows that someone can dream of a mission profile. As far as I can tell, it is mostly just a design exercise. It is not even a real proposal.

As for "possible" - certainly not with 1970s era technology. As for "achievable", since it isn't even a proposal, I think that is answered by my first message. 

I too can draw up a mission plan for crewed venus, doesn't mean it is realistic.

>  id also argue that every crewed venture in the solar system is risky to some degree    

Sure, but there is a wild difference between "We are landing crew on Mars to do geological surveys" to "we are telling the astronauts to do the single highest risk phase in a spaceflight mission repeatedly for no gain" or "we are telling them to spend time in a blimp for no fain".

WithCrow:
Like I have said from the beginning of the original PR: this program should have crew enter orbit around Venus & drop atmospheric probes. There is zero reason for the crew to be an atmospheric probe. Zero. No gain except incredible danger.

WithCrow:
Furthermore: HAVOC is an illsuited example anyway, since it is from 2015. AAP is far better, due to being from the same time period. The avionics, material science, etc. just did not exist.

gabo463:
the goal of this program always has been to appeal to all tech level not just the 70s in my opinion doing that when it comes to crewed interplanetary contract restrain the content there for making less fun programs. now yes HAVOC functioned purely as an internal training and systems analysis exercise but i dont see why there wouldnt be a chance of if crewed venture to venus was ever done someting like Havoc  wouldn't be pursued. in term of gain such mission id argue that it  Aligns with VEXAG (Venus Exploration Analysis Group) goals to study atmospheric chemistry, weather patterns, cloud dynamics, and solar wind interactions safely from an altitude platform and that The upper cloud layers could potentially harbor microbial life adapted to acidic environments. now that all something that could be done with a probe but so is geological surveys on mars. on that note i should require the deployment of atmospheric probes

WithCrow:
Since your "extra fun" content currently requires the installation of two other PRs & a non-express install mod, I am already not a huge fan of it.

It is _not_ suitable for all tech levels. HAVOC-level tech should not be available prior to the 2020s.

You misunderstand the impact of humans in exploration. The advantage of a human over a probe: the ability to improvise. A human can dynamically evaluate what rocks are interesting, where to drill, etc.

A human stuck in a glorified can in the venusian atmosphere provides no such ability.

gabo463:
alright point taken im gonna remove the optional contract

Glloyd:
I'm with Crow here. I think if you really want to have an atmospheric entry included, the best way to do this would be to have some kind of atmospheric sample return experiment for Venus that you can run with uncrewed probes, but have to return to complete the contract/program. But I'm fairly certain siimav has already said that they won't accept any Venus program that relies on adding new mods, so any atmospheric stay is out of the question.

gabo463:
removed the crewed atmospheric contrac this will be replaced by a  a  third required contract of atmospheric probe deployment

gabo463:
> I'm with Crow here. I think if you really want to have an atmospheric entry included, the best way to do this would be to have some kind of atmospheric sample return experiment for Venus that you can run with uncrewed probes, but have to return to complete the contract/program. But I'm fairly certain siimav has already said that they won't accept any Venus program that relies on adding new mods, so any atmospheric stay is out of the question.

i did not know about this development warning i wouldn't have included it if i had known. but all is good anyway i removed this contract

gabo463:
added the atmospheric probe - for it not to be completed without the crew i made it so it must be deployed by a carrier craft that is crewed

gabo463:
Question for this program : should the mandatory atmospheric probe deployment be numbered (a set number of probes to be deployed) or it should be left to the player to decide

connorjak:
There already is an atmospheric probe contract in a past program, so there needs to be something different about one if it's in this program. https://github.com/KSP-RO/RP-1/blob/master/GameData/RP-1/Contracts/Venus/Venus%20Probe.cfg

If you're doing an atmospheric sample return contract, then it should be a certain experiment required with realistic restrictions on what situations result in collecting the experiment. For example, a specific altitude range and max velocity is probably enough.

For gameplay reasons, typically one sample experiment that requires the player to engineer a vehicle to collect is way better than an arbitrary "deploy x probes".

As for why this would fit in a crewed Venus program, maybe the operations and astrodynamics smarts to do atmospheric relaunch and sample recovery practically require human teleoperation from a mothership, before year ~2000. The RP-1 modpack doesn't simulate communication time delay of teleoperation or imperfect astrodynamics, so this is just a for-lore-reasons kind of situation.

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -1849,36 +1849,47 @@ RP0_PROGRAM
...
-	objectivesPrettyText = Do this and that
-	nominalDurationYears = 10
-	baseFunding = 15000000
-	repDeltaOnCompletePerYearEarly = 1130
-	repPenaltyPerYearLate = 1130
-	repToConfidence = 3
-	slots = 4
+	title = Crewed Venus Exploration 
+	description = We have shown we can survive around and on the Moon. Now, we take the next monumental leap into the deep solar system: Venus. Searing heat, dense crushing atmosphere, and deep space radiation stand between our crew and the morning star.
+	requirementsPrettyText = Complete crewed operations around Earth's Moon and master early long-duration space stations.
```
connorjak: Not yet edited

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -1849,36 +1849,47 @@ RP0_PROGRAM
...
+	   complete_program = EarlyEarthSpaceStation
+	   complete_program = CrewedLunar
+       complete_program = VenusSurfaceExp
 	}
 
 	OBJECTIVES
 	{
-		complete_contract = TODO
+
+		omplete_contract = FirstCrewedVenusOrbit
```
connorjak: typo

GameData/RP-1/Programs/Programs.cfg
```diff
@@ -1849,36 +1849,47 @@ RP0_PROGRAM
...
+		complete_contract = firstVenusFlybyCrewed
+		complete_contract = atmosphericProbes
```
connorjak: I disagree with this being required

connorjak (changes requested):

GameData/RP-1/Programs/Programs.cfg
gabo463: Wich one ?

gabo463:
> There already is an atmospheric probe contract in a past program, so there needs to be something different about one if it's in this program. https://github.com/KSP-RO/RP-1/blob/master/GameData/RP-1/Contracts/Venus/Venus%20Probe.cfg
> 
> If you're doing an atmospheric sample return contract, then it should be a certain experiment required with realistic restrictions on what situations result in collecting the experiment. For example, a specific altitude range and max velocity is probably enough.
> 
> For gameplay reasons, typically one sample experiment that requires the player to engineer a vehicle to collect is way better than an arbitrary "deploy x probes".
> 
> As for why this would fit in a crewed Venus program, maybe the operations and astrodynamics smarts to do atmospheric relaunch and sample recovery practically require human teleoperation from a mothership, before year ~2000. The RP-1 modpack doesn't simulate communication time delay of teleoperation or imperfect astrodynamics, so this is just a for-lore-reasons kind of situation.

I do agree with the atmospheric-sample return contract however I do not think something like runing a specific a specific experiment is the way to go this is in my opinion to restrictive for the player especially since probing Venus is required for this program this would put some sort of (save that science experiment for later)  effect wich is not wanted. Also the target tech level for this program is the 70s to 80s

gabo463:
added requested change and some of the suggestion

karbolenius:
Hi, @gabo463!

I've started diving into planning a Venus mission for this program and have noticed a number of design issues.
These mostly concern the VenusOrbitalStay contracts.
First, the VenusOrbitalStay4months contract depends on VenusOrbitalStay8months, and the latter depends on the former. This makes it impossible to accept this chain at all.

Second, if I fly from Earth to Venus during the window and wait there for 4 months, then to immediately return to Earth I'll need about ~33 km/s of delta due to the unfavorable mutual orbital position of Venus and Earth. Of course, I'll wait longer. In fact, to wait for the window from Venus to Earth, I'll need to spend about 16 months in Venus orbit. This makes the 4- and 8-month contracts pointless.

Third, I don't understand why the "atmosphericSampleReturnVenus" contract requires the "crewedVenusFlyby" contract to be executed. As it requires building a new vessel, I'll have to send a new craft to Venus during the next window from Earth, and it'll be easier for me to make it unmanned. But the program is called the "Crewed Venus Program." Look, in the "Crewed Mars Exploration" program, we can complete all mandatory and optional contracts within a single expedition. It would be logical for the crew on the mothership to take and dock a craft containing an atmospheric sample (after it leaves the atmosphere) and than return home with it.

gabo463:
I would respond to this by saying the Venus program is not designed in the mind of being completed in a single expedition at least 2 is required  one for the fly-by and another for the first crewed orbit that’s said the atmospheric sample return is designed to be completed with the first crewed orbit of Venus tho I can understand some would like to attempt to do it during the flyby due of its repeatable nature as what concerns the habitable stays I will have a look into it during testing it wasn’t bothering that much in fact  I tought it would be welcomed to complete contract like that for free while waiting

WithCrow:
The fact it isn't clashes with Crewed Mars & expectations built up there. If the user _can_, they should be able to do this immediately.

gabo463:
alright removed 4 to 8 month contracts  and made it so the program could be completed in 1 expedition

gabo463:
contract felt a bit short in the end so i added a 32 month stay  that can be completed in tandem with the 16 month stay or if the player desire alone. following feedback all of this is designed to if the player can to be able to complete it straight away

PR #2868, opened 2026-06-30
Add SustainedCruise contract parameter (range-based endurance)

chambm:
Adds a `SustainedCruise` Contract Configurator parameter (in CC_RP0) for
sustained-cruise / endurance objectives, so endurance contracts don't have to
require flying a long real-time `Duration`.

The active vessel must hold ALL of these **simultaneously and continuously** for
`holdSeconds`:

- surface speed within `[minSpeed, maxSpeed]`,
- vertical speed within `[minVerticalSpeed, maxVerticalSpeed]`,
- projected **range** >= `requiredRange`.

Any violation resets the continuous-hold timer, so the conditions can't be
satisfied separately ("cheesed").

Endurance is measured as **range** rather than time:
`range = speed * fuel / burnRate`, computed with the Breguet range equation so
it credits the craft getting lighter as fuel burns. This rewards flying faster
and more efficiently (e.g. higher, where jets burn less). Burn rate is averaged
over a rolling `rateWindowSeconds` and only sampled while at cruise speed, so the
rate is representative.

Config fields: `resource`, `requiredRange` (m), `holdSeconds` (def 300),
`minSpeed`/`maxSpeed`, `minVerticalSpeed`/`maxVerticalSpeed`, `rateWindowSeconds`
(def 30), `updateFrequency` (def 0.5), `debug` (def false). Follows the existing
ReachMach parameter/factory pattern.

Intended for the Advanced X-Planes supersonic contracts (those contract changes
are kept separate as they depend on KSP-RO/RP-1#2828).

### DIFF
--- a/Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
+++ b/Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
@@ -5,31 +5,31 @@ using UnityEngine;
 namespace ContractConfigurator.RP0
 {
     /// <summary>
-    /// A single sustained-cruise test. The active vessel must hold ALL of these SIMULTANEOUSLY and
-    /// continuously for holdSeconds:
+    /// Instantaneous cruise-endurance test: met while the active vessel simultaneously satisfies ALL of:
     ///   * surface speed within [minSpeed, maxSpeed]
     ///   * vertical speed within [minVerticalSpeed, maxVerticalSpeed]
     ///   * projected range >= requiredRange
     ///
     /// Endurance is measured as RANGE rather than time, so flying faster and/or more efficiently (e.g.
-    /// higher, where jets burn less for the same speed) is rewarded. Any violation resets the continuous
-    /// hold timer, so the conditions can't be satisfied separately ("cheesed") -- they must all be true at
-    /// once for the whole period.
+    /// higher, where jets burn less for the same speed) is rewarded. Range is computed for EVERY mass-bearing
+    /// propellant the vessel is actively burning (jets run on all sorts of resources here -- kerosene,
+    /// hydrogen, methane, nitrous, water injection, enriched uranium), and the reported range is the MINIMUM
+    /// across them, i.e. whichever fuel runs out first. Each fuel's burn rate is averaged over a rolling
+    /// rateWindowSeconds, and samples are only taken while at cruise speed so the rate is representative.
+    /// Massless resources (ElectricCharge, IntakeAir) are ignored.
     ///
-    /// Range is computed for EVERY mass-bearing propellant the vessel is actively burning (jets run on all
-    /// sorts of resources here -- kerosene, hydrogen, methane, nitrous, water injection, enriched uranium),
-    /// and the reported range is the MINIMUM across them, i.e. whichever fuel runs out first. Each fuel's
-    /// burn rate is averaged over a rolling rateWindowSeconds, and samples are only taken while at cruise
-    /// speed so the rate is representative. Massless resources (ElectricCharge, IntakeAir) are ignored.
+    /// This parameter is a plain instantaneous check. To require the condition be held for a time -- and to
+    /// get a countdown display -- pair it with a stock Duration parameter placed as a LATER sibling, and set
+    /// disableOnStateChange = false here so its state toggles as the vessel enters/leaves the band (Duration
+    /// times only while its preceding siblings are Complete, and resets when one stops).
     ///
-    /// Config: requiredRange m (req), holdSeconds (def 300), minSpeed/maxSpeed m/s,
-    /// minVerticalSpeed/maxVerticalSpeed m/s, rateWindowSeconds (def 30), updateFrequency (def 0.5).
-    /// Enable CC verbose logging to see the per-tick speed/VS/range/held values.
+    /// Config: requiredRange m (req), minSpeed/maxSpeed m/s, minVerticalSpeed/maxVerticalSpeed m/s,
+    /// rateWindowSeconds (def 30), updateFrequency (def 0.5). Enable CC verbose logging to see the per-tick
+    /// speed/VS/range/met values.
     /// </summary>
     public class SustainedCruise : VesselParameter
     {
         protected double requiredRange { get; set; }
-        protected double holdSeconds { get; set; }
         protected double minSpeed { get; set; }
         protected double maxSpeed { get; set; }
         protected double minVerticalSpeed { get; set; }
@@ -37,7 +37,6 @@ namespace ContractConfigurator.RP0
         protected double rateWindowSeconds { get; set; }
         protected float updateFrequency { get; set; }
 
-        internal const double DEFAULT_HOLD = 300.0;
         internal const double DEFAULT_RATE_WINDOW = 30.0;
         internal const float DEFAULT_UPDATE_FREQUENCY = 0.5f;
         private const double MIN_RATE = 1e-4;   // below this the craft isn't really burning this resource
@@ -46,7 +45,7 @@ namespace ContractConfigurator.RP0
         private double lastCheckTime = double.NegativeInfinity;
         private int lastPartCount = -1;
         private uint lastVesselPersistentId;
-        private double heldTime = 0.0;   // continuous time all conditions have held
+        private bool met = false;   // whether all conditions were satisfied as of the last sample
         // Rolling per-tick samples for the burn-rate average: sampleTimes[k] holds the time of tick k, and
         // sampleAmounts[k] maps resource id -> total amount on the vessel at that tick.
         private readonly List<double> sampleTimes = new List<double>();
@@ -55,13 +54,12 @@ namespace ContractConfigurator.RP0
 
         public SustainedCruise() : base(null) { }
 
-        public SustainedCruise(string title, double requiredRange, double holdSeconds,
+        public SustainedCruise(string title, double requiredRange,
                                double minSpeed, double maxSpeed, double minVerticalSpeed, double maxVerticalSpeed,
                                double rateWindowSeconds, float updateFrequency)
             : base(title)
         {
             this.requiredRange = requiredRange;
-            this.holdSeconds = holdSeconds;
             this.minSpeed = minSpeed;
             this.maxSpeed = maxSpeed;
             this.minVerticalSpeed = minVerticalSpeed;
@@ -75,7 +73,6 @@ namespace ContractConfigurator.RP0
         {
             base.OnParameterSave(node);
             node.AddValue("requiredRange", requiredRange);
-            node.AddValue("holdSeconds", holdSeconds);
             node.AddValue("minSpeed", minSpeed);
             node.AddValue("maxSpeed", maxSpeed);
             node.AddValue("minVerticalSpeed", minVerticalSpeed);
@@ -88,7 +85,6 @@ namespace ContractConfigurator.RP0
         {
             base.OnParameterLoad(node);
             requiredRange = ConfigNodeUtil.ParseValue<double>(node, "requiredRange");
-            holdSeconds = ConfigNodeUtil.ParseValue<double>(node, "holdSeconds", DEFAULT_HOLD);
             minSpeed = ConfigNodeUtil.ParseValue<double>(node, "minSpeed", 0.0);
             maxSpeed = ConfigNodeUtil.ParseValue<double>(node, "maxSpeed", double.MaxValue);
             minVerticalSpeed = ConfigNodeUtil.ParseValue<double>(node, "minVerticalSpeed", double.MinValue);
@@ -100,16 +96,14 @@ namespace ContractConfigurator.RP0
         protected override string GetParameterTitle()
         {
             double km = requiredRange / 1000.0;
-            double hold = holdSeconds / 60.0;
             string band = maxSpeed >= double.MaxValue * 0.5 ? $">= {minSpeed:0} m/s" : $"{minSpeed:0}-{maxSpeed:0} m/s";
-            return $"Sustain cruise for {hold:0} min: range >= {km:0} km, speed {band}";
+            return $"Cruise: range >= {km:0} km, speed {band}";
         }
 
         private void ResetAll()
         {
             sampleTimes.Clear();
             sampleAmounts.Clear();
-            heldTime = 0.0;
         }
 
         // Smallest projected range (m) across every mass-bearing resource the craft is actively burning,
@@ -206,7 +200,6 @@ namespace ContractConfigurator.RP0
                 lastVesselPersistentId = v.persistentId;
                 lastPartCount = v.parts.Count;
                 ResetAll();
-                dt = 0.0;
             }
             lastCheckTime = now;
 
@@ -214,11 +207,12 @@ namespace ContractConfigurator.RP0
             double vs = v.verticalSpeed;
             bool speedOk = speed >= minSpeed && speed <= maxSpeed;
 
-            // Out of the cruise speed band: rate isn't representative and the hold is broken. Drop the
-            // window + timer so endurance can't be "proven" by slowing down.
+            // Out of the cruise speed band: rate isn't representative. Drop the window so the burn rate has
+            // to rebuild before range can be proven again, and mark the condition unmet.
             if (!speedOk)
             {
                 ResetAll();
+                met = false;
                 LogState(speed, vs, false, false, false, 0.0, null);
                 CheckVessel(v);
                 return;
@@ -235,9 +229,7 @@ namespace ContractConfigurator.RP0
             bool rateReady = TryMinRange(v, speed, out double range, out PartResourceDefinition limiter);
             bool vsOk = vs >= minVerticalSpeed && vs <= maxVerticalSpeed;
             bool rangeOk = rateReady && range >= requiredRange;
-
-            if (vsOk && rangeOk) heldTime += dt;   // speedOk already true here
-            else heldTime = 0.0;
+            met = vsOk && rangeOk;   // speedOk already true here
 
             LogState(speed, vs, speedOk, vsOk, rangeOk, range, limiter);
             CheckVessel(v);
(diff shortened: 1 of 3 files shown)


siimav:
> [Add contract-earned permanent training-rate bonus](https://github.com/KSP-RO/RP-1/pull/2868/commits/08865ce1a03ec2f904f8c6786ede99e31d9b7f58)

Uhm. WHAT.

chambm:
> > [Add contract-earned permanent training-rate bonus](https://github.com/KSP-RO/RP-1/pull/2868/commits/08865ce1a03ec2f904f8c6786ede99e31d9b7f58)
> 
> Uhm. WHAT.

I mentioned it in Discord back when I first came up with the idea. It seems like a cool and appropriate incentive to do the contract. People are always complaining about X-Planes being a dead end and not being relevant to further goals in RP-1's spaceflight programs, but the Vomit Comet has been used for decades to train astronauts. I am, of course, completely open to whatever balance suggestions you have. I think the current bonus is probably too powerful. And if you really think such a bonus is inappropriate, it's at least a reason to design a regular subsonic airliner style plane: what else are you going to use those high bypass turbofan engines for. :)

siimav:
In case you haven't noticed, I do not like PRs that do too many things at once.

chambm:
> In case you haven't noticed, I do not like PRs that do too many things at once.

OK, do you want it split into 2, or 3? I figure the contract criteria parameters can go together or not.

siimav:
Ideally 3

Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
```diff
@@ -0,0 +1,227 @@
...
+                CheckVessel(v);
+                return;
+            }
+
+            double amount = 0.0;
+            for (int i = 0; i < v.parts.Count; i++)
+            {
+                PartResource pr = v.parts[i].Resources.Get(resource.id);
+                if (pr != null) amount += pr.amount;
+            }
```
siimav: Vessel.GetConnectedResourceTotals

Source/CC_RP0/ParameterFactory/SustainedCruiseFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+        protected double maxVerticalSpeed;
+        protected double rateWindowSeconds;
+        protected float updateFrequency;
+        protected bool debug;
+
+        public override bool Load(ConfigNode configNode)
+        {
+            bool valid = base.Load(configNode);
+
+            valid &= ConfigNodeUtil.ParseValue<Resource>(configNode, "resource", x => resource = x.res, this);
```
siimav: Pretty sure engines run on more propellants than one. Especially so for rocketplanes.

Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
```diff
@@ -0,0 +1,227 @@
...
+            node.AddValue("maxVerticalSpeed", maxVerticalSpeed);
+            node.AddValue("rateWindowSeconds", rateWindowSeconds);
+            node.AddValue("updateFrequency", updateFrequency);
+            node.AddValue("debug", debug);
+        }
+
+        protected override void OnParameterLoad(ConfigNode node)
+        {
+            base.OnParameterLoad(node);
+            resource = PartResourceLibrary.Instance.GetDefinition(ConfigNodeUtil.ParseValue<string>(node, "resource"));
```
siimav: This blowing up due to resource being missing would be real bad.

Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
```diff
@@ -0,0 +1,227 @@
...
+            holdSeconds = ConfigNodeUtil.ParseValue<double>(node, "holdSeconds", DEFAULT_HOLD);
+            minSpeed = ConfigNodeUtil.ParseValue<double>(node, "minSpeed", 0.0);
+            maxSpeed = ConfigNodeUtil.ParseValue<double>(node, "maxSpeed", double.MaxValue);
+            minVerticalSpeed = ConfigNodeUtil.ParseValue<double>(node, "minVerticalSpeed", double.MinValue);
+            maxVerticalSpeed = ConfigNodeUtil.ParseValue<double>(node, "maxVerticalSpeed", double.MaxValue);
+            rateWindowSeconds = ConfigNodeUtil.ParseValue<double>(node, "rateWindowSeconds", DEFAULT_RATE_WINDOW);
+            updateFrequency = ConfigNodeUtil.ParseValue<float>(node, "updateFrequency", DEFAULT_UPDATE_FREQUENCY);
+            // bool collides with CC's (node, key, allowExpression) overload, so read it manually.
+            debug = false;
+            if (node.HasValue("debug") && bool.TryParse(node.GetValue("debug"), out bool db)) debug = db;
```
siimav: manual bool.TryParse workaround is avoidable. DownrangeDistanceVesselParam reads a bool with the 2-arg ConfigNodeUtil.ParseValue(node, "triggered"). debug = node.HasValue("debug") && ConfigNodeUtil.ParseValue(node, "debug"); matches sibling style without hand-parsing.

Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
```diff
@@ -0,0 +1,227 @@
...
+            // window + timer so endurance can't be "proven" by slowing down.
+            if (!speedOk)
+            {
+                sampleTimes.Clear();
+                sampleAmounts.Clear();
+                heldTime = 0.0;
+                DebugLog(speed, vs, false, false, false, 0.0, 0.0, 0.0);
+                CheckVessel(v);
+                return;
+            }
```
siimav: duplicates ResetAll()

Source/CC_RP0/ParameterFactory/SustainedCruiseFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+            valid &= ConfigNodeUtil.ParseValue<double>(configNode, "holdSeconds", x => holdSeconds = x, this, SustainedCruise.DEFAULT_HOLD, x => Validation.GT(x, 0.0));
+            valid &= ConfigNodeUtil.ParseValue<double>(configNode, "minSpeed", x => minSpeed = x, this, 0.0, x => Validation.GE(x, 0.0));
+            valid &= ConfigNodeUtil.ParseValue<double>(configNode, "maxSpeed", x => maxSpeed = x, this, double.MaxValue);
+            valid &= ConfigNodeUtil.ParseValue<double>(configNode, "minVerticalSpeed", x => minVerticalSpeed = x, this, double.MinValue);
+            valid &= ConfigNodeUtil.ParseValue<double>(configNode, "maxVerticalSpeed", x => maxVerticalSpeed = x, this, double.MaxValue);
+            valid &= ConfigNodeUtil.ParseValue<double>(configNode, "rateWindowSeconds", x => rateWindowSeconds = x, this, SustainedCruise.DEFAULT_RATE_WINDOW, x => Validation.GT(x, 0.0));
+            valid &= ConfigNodeUtil.ParseValue<float>(configNode, "updateFrequency", x => updateFrequency = x, this, SustainedCruise.DEFAULT_UPDATE_FREQUENCY, x => Validation.GT(x, 0.0f));
+            valid &= ConfigNodeUtil.ParseValue<bool>(configNode, "debug", x => debug = x, this, false);
```
siimav: Honestly not the best idea to make some kind of specific logging for just this one contract parameter. CC itself has verbose logger and so does RP-1.

siimav (changes requested):
How do the range numbers compare to what FAR reports?

chambm:
> How do the range numbers compare to what FAR reports?

Very close, but more stable because it's averaged over a longer window.

Source/CC_RP0/ParameterFactory/SustainedCruiseFactory.cs
chambm: So, no dedicated debug mode: instead send those messages to the debug log channel for CC?

Source/CC_RP0/ParameterFactory/SustainedCruiseFactory.cs
chambm: You think it's worth tracking burn rate for all the fuels? I figured this was mostly for jets.

siimav: I'm not sure we're ever going to use this for rockets in practice BUT even jets can run on various fuels or combinations. For instance there are ones with water injection. Then there are a few than run on either hydrogen or methane. J58-P-4B-N2O on nitrous. And nuclear engines on EnrichedUranium. We even have ducted fans.

chambm: Tricky. So, track all fuels present on the vessel, and final endurance is the minimum endurance for each fuel type? It's a best guess that assumes the player will keep the flying the way they have been; i.e. it's not going to behave great, at least right away, if player switches fuels in midflight, e.g. injection toggle or RAPIER.

chambm:
I'm still testing the parity with FAR endurance. I think FAR is being overly optimistic and my code is more correct (pessimistic/realistic), but will verify.

chambm:
After extensive testing I think it's the original FAR range numbers that range from slightly too optimistic to wildly optimistic.
image
I ported the sustained cruise calculation into FAR itself (with 5 second window instead of 30) and in my testing it's really close to what the vehicle actually can achieve. Apparently the FAR calculation is excluding ALL resources in its "dry mass": things that the jet engines don't actually burn like payload/sounding resources, oxidizer/rocket fuel, etc. I'll see about fixing that in my FAR PR though.

Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
```diff
@@ -0,0 +1,252 @@
...
+                return speed * (M / massRate) * System.Math.Log(M / mEmpty);
+            return speed * (currentUnits / rateUnits);       // linear fallback
+        }
+
+        // Snapshot of every mass-bearing resource's vessel-wide total, keyed by resource id. Massless
+        // resources (ElectricCharge, IntakeAir) are skipped so they can't masquerade as a fuel.
+        private Dictionary<int, double> SampleResources(Vessel v)
+        {
+            massBearingIds.Clear();
+            for (int i = 0; i < v.parts.Count; i++)
```
siimav, line 192: Better to gather the interesting resources once and only update the list on vessel part count change.
Perhaps this would also help with getting rid of snapshot dictionary allocations on every call?

Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
```diff
@@ -0,0 +1,252 @@
...
+            updateFrequency = ConfigNodeUtil.ParseValue<float>(node, "updateFrequency", DEFAULT_UPDATE_FREQUENCY);
+        }
+
+        protected override string GetParameterTitle()
+        {
+            double reqKm = requiredRange / 1000.0;
+            string band = maxSpeed >= double.MaxValue * 0.5 ? $">= {minSpeed:0} m/s" : $"{minSpeed:0}-{maxSpeed:0} m/s";
+            if (FlightGlobals.ActiveVessel == null)
+                return $"Cruise: range >= {reqKm:0} km, speed {band}";
+            return $"Cruise: range {curRange / 1000.0:0} / {reqKm:0} km, speed {band}";
```
siimav: The other parameters use N0 as the format.

Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
```diff
@@ -0,0 +1,252 @@
...
+            minVerticalSpeed = ConfigNodeUtil.ParseValue<double>(node, "minVerticalSpeed", double.MinValue);
+            maxVerticalSpeed = ConfigNodeUtil.ParseValue<double>(node, "maxVerticalSpeed", double.MaxValue);
+            rateWindowSeconds = ConfigNodeUtil.ParseValue<double>(node, "rateWindowSeconds", DEFAULT_RATE_WINDOW);
+            updateFrequency = ConfigNodeUtil.ParseValue<float>(node, "updateFrequency", DEFAULT_UPDATE_FREQUENCY);
+        }
+
+        protected override string GetParameterTitle()
+        {
+            double reqKm = requiredRange / 1000.0;
+            string band = maxSpeed >= double.MaxValue * 0.5 ? $">= {minSpeed:0} m/s" : $"{minSpeed:0}-{maxSpeed:0} m/s";
```
siimav: Should make maxSpeed nullable.

Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
```diff
@@ -0,0 +1,252 @@
...
+            double mEmpty = M - mFuel;                       // mass when it runs out (t)
+            double massRate = rateUnits * dens;              // burn rate (t/s)
+            if (dens > 0.0 && massRate > 0.0 && mFuel > 1e-6 && mEmpty > 1e-3)
+                return speed * (M / massRate) * System.Math.Log(M / mEmpty);
+            return speed * (currentUnits / rateUnits);       // linear fallback
+        }
+
+        // Snapshot of every mass-bearing resource's vessel-wide total, keyed by resource id. Massless
+        // resources (ElectricCharge, IntakeAir) are skipped so they can't masquerade as a fuel.
+        private Dictionary<int, double> SampleResources(Vessel v)
```
siimav: I wonder if it would be cleaner to just find all working engines and gather the propellants from those.

Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        protected override string GetParameterTitle()
+        {
+            double reqKm = requiredRange / 1000.0;
+            string band = maxSpeed >= double.MaxValue * 0.5 ? $">= {minSpeed:0} m/s" : $"{minSpeed:0}-{maxSpeed:0} m/s";
+            if (FlightGlobals.ActiveVessel == null)
+                return $"Cruise: range >= {reqKm:0} km, speed {band}";
+            return $"Cruise: range {curRange / 1000.0:0} / {reqKm:0} km, speed {band}";
+        }
```
siimav, line 116: The title omits the vertical-speed constraint entirely, so a player failing on VS gets no feedback.

siimav (changes requested):

chambm:
I think this addresses your concerns. I tested it with a NO2 afterburner and the range changed drastically after switching on the NO2, although it happened gradually as the 30 seconds of the kerosene only window was flushed out.

Source/CC_RP0/ParameterFactory/SustainedCruiseFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+            valid &= ConfigNodeUtil.ParseValue<double>(configNode, "minSpeed", x => minSpeed = x, this, 0.0, x => Validation.GE(x, 0.0));
+            maxSpeed = null;
+            if (configNode.HasValue("maxSpeed"))
+                valid &= ConfigNodeUtil.ParseValue<double>(configNode, "maxSpeed", x => maxSpeed = x, this, double.MaxValue, x => Validation.GT(x, minSpeed));
+            minVerticalSpeed = null;
+            if (configNode.HasValue("minVerticalSpeed"))
+                valid &= ConfigNodeUtil.ParseValue<double>(configNode, "minVerticalSpeed", x => minVerticalSpeed = x, this);
+            maxVerticalSpeed = null;
+            if (configNode.HasValue("maxVerticalSpeed"))
+                valid &= ConfigNodeUtil.ParseValue<double>(configNode, "maxVerticalSpeed", x => maxVerticalSpeed = x, this);
```
siimav: Check out HorizontalLandingVesselParam.cs. These can be simplified a lot.

Source/CC_RP0/ParameterFactory/SustainedCruiseFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+        protected double? minVerticalSpeed;
+        protected double? maxVerticalSpeed;
+        protected double rateWindowSeconds;
+        protected float updateFrequency;
+
+        public override bool Load(ConfigNode configNode)
+        {
+            bool valid = base.Load(configNode);
+
+            valid &= ConfigNodeUtil.ParseValue<double>(configNode, "requiredRange", x => requiredRange = x, this, 0.0, x => Validation.GT(x, 0.0));
```
siimav: Either default to something valid or drop the default so CC reports it as missing

Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
```diff
@@ -0,0 +1,344 @@
...
+    /// get a countdown display -- pair it with a stock Duration parameter placed as a LATER sibling, and set
+    /// disableOnStateChange = false here so its state toggles as the vessel enters/leaves the band (Duration
+    /// times only while its preceding siblings are Complete, and resets when one stops).
+    ///
+    /// Config: requiredRange m (req), minSpeed/maxSpeed m/s, minVerticalSpeed/maxVerticalSpeed m/s,
+    /// rateWindowSeconds (def 30), updateFrequency (def 0.5). Set the CC log level for the SustainedCruise
+    /// type to DEBUG (or VERBOSE) via an ADD_LOGLEVEL_EXCEPTION to see the per-tick speed/VS/range/met and
+    /// Breguet-input breakdown lines. (LogDebug, not LogVerbose, so it survives the Release build --
+    /// LogVerbose is [Conditional("DEBUG")] and gets compiled out.)
+    /// </summary>
```
siimav, line 18: Good lord is your clanker writing a book here. I doubt anyone wants to read implementation notes here. If you want to keep all that information then try moving parts of it closer to the actual logic.

Source/CC_RP0/Parameter/SustainedCruiseVesselParam.cs
```diff
@@ -0,0 +1,344 @@
...
+                limiter.name, speed, M, mFuel, mEmpty, mdotKgS, lnTerm, range, linear);
+        }
+
+        private void LogState(double speed, double vs, bool speedOk, bool vsOk, bool rangeOk, double range, PartResourceDefinition limiter)
+        {
+            LoggingUtil.LogDebug(this, "SustainedCruise spd={0:0}({1}) vs={2:0.0}({3}) range={4:0}/{5:0}km({6}) limiter={7} -> {8}",
+                speed, speedOk ? "ok" : "X", vs, vsOk ? "ok" : "X",
+                range / 1000.0, requiredRange / 1000.0, rangeOk ? "ok" : "X",
+                limiter != null ? limiter.name : "-", met ? "IN" : "out");
+        }
```
siimav, line 322: Perhaps guard all that with `LoggingUtil.GetLogLevel(GetType()) <= LoggingUtil.LogLevel.DEBUG`. Saves a bunch of code from running; arrays getting allocated; and doubles getting boxed.

siimav (changes requested):
I swear this is the last round...

Source/CC_RP0/ParameterFactory/SustainedCruiseFactory.cs
```diff
@@ -0,0 +1,36 @@
...
+        protected double? minVerticalSpeed;
+        protected double? maxVerticalSpeed;
+        protected double rateWindowSeconds;
+        protected float updateFrequency;
+
+        public override bool Load(ConfigNode configNode)
+        {
+            bool valid = base.Load(configNode);
+
+            valid &= ConfigNodeUtil.ParseValue<double>(configNode, "requiredRange", x => requiredRange = x, this, x => Validation.GT(x, 0.0));
```
siimav, line 19: Either default to something valid or drop the default so CC reports it as missing

siimav: I swear I added it into the last 2 reviews but for some reason GH didn't officially post it.

chambm: Are you referring to the requiredRange line? I don't think it has a default value there. Which parameter is acting as the default?

siimav: Oh, looks like you changed it at one point. I can now see another review comment above with the old state. IDK what GH was up to here.

PR #2870, opened 2026-07-02
Prevent undefined behaviour when LC is dismantled while under construction

periodically-makes-puns:
https://discord.com/channels/319857228905447436/512556137380315139/1512505950463070310 (reproduction)
https://discord.com/channels/319857228905447436/512556137380315139/1521754770765906063 (save and logs)

If an LC is Dismantled while it is actively under construction, `SpaceCenterManagement.ProgressBuildTime` hits an NRE while processing the completion of the dismantled LC construction, which prevents any progress towards research, fund target, and staff target (which all occur after processing constructions).

To remedy this, this PR causes dismantling to fail if the LC or any of its pads are under construction.

### DIFF
--- a/Source/RP0/SpaceCenter/LaunchComplex/LCLaunchPad.cs
+++ b/Source/RP0/SpaceCenter/LaunchComplex/LCLaunchPad.cs
@@ -160,6 +160,15 @@ namespace RP0
                         }
                     }
 
+                    foreach (var cons in currentLC.PadConstructions)
+                    {
+                        if (cons.id == id)
+                        {
+                            failReason = "pad is under construction";
+                            return false;
+                        }
+                    }
+
                     foreach (VesselProject vessel in currentLC.Warehouse)
                     {
                         if (vessel.launchSiteIndex >= idx) vessel.launchSiteIndex--;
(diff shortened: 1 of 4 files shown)


siimav:
* There might be one more case.
With 2 operational pads plus a 3rd under construction, the build list renders a Dismantle button on the under-construction pad's row (GUI_BuildList.cs:1700 gates on LaunchPadCount > 1, which counts only operational pads). Clicking it calls SwitchLaunchPad(j), whose int overload sets ActiveLaunchPadIndex without any isOperational skip (LaunchComplex.cs:315). Confirming "Yes" then deletes the under-construction pad, and LCLaunchPad.Delete() never removes the matching PadConstructionProject.

* Perhaps it might sense to clean up associated constructions in LaunchComplex.Delete() as a safety net?
* LaunchComplex.CanDismantle already kinda does similar things. Maybe it makes sense to fold the checks in there?

periodically-makes-puns:
added:
- don't render the Dismantle button for non-operational Pads
- all relevant constructions are canceled in LaunchComplex.Delete just in case

I considered hooking more logic into CanDismantle, but the current system makes it easier to display the failure reason and I don't really want to rewrite that

PR #2872, opened 2026-07-09
Silence alarms and crew/maintenance processing during sim time jumps

chambm:
An SCM simulation can jump UT forward by years. That fired every KAC alarm the jump crossed and popped crew-retirement notifications for time that isn't really passing.

- CrewHandler.Process and MaintenanceHandler.Update/FixedUpdate bail out when IsSimulatedFlight, so retirement/inactivity/upkeep don't process against the skipped time.
- AlarmHelper.SuppressAllAlarmActions(throughUT) disarms alarms due within the jump window before the UT is set; restored from the pre-sim backup on exit. KAC path sets Enabled=false (its firing loop ignores the legacy AlarmAction field the wrapper bound); gated on AssemblyExists since APIReady is false during the scene load.
- KACWrapper: expose the alarm Enabled field.

### DIFF
--- a/Source/RP0/Maintenance/MaintenanceHandler.cs
+++ b/Source/RP0/Maintenance/MaintenanceHandler.cs
@@ -411,6 +411,11 @@ namespace RP0
             if (HighLogic.CurrentGame == null)
                 return;
 
+            // Skip maintenance/upkeep/reputation processing while simulating; an SCM sim can jump
+            // UT forward by years and this would churn funds/rep against time that isn't passing.
+            if (SpaceCenterManagement.Instance?.IsSimulatedFlight == true)
+                return;
+
             if (_frameCount++ < 2)
             {
                 return;
@@ -453,6 +458,10 @@ namespace RP0
 
         public void FixedUpdate()
         {
+            // See Update(): don't process crew/maintenance during a simulation time jump.
+            if (SpaceCenterManagement.Instance?.IsSimulatedFlight == true)
+                return;
+
             double UT = Planetarium.GetUniversalTime();
             if (_lastUpdateFixed == 0)
                 _lastUpdateFixed = UT;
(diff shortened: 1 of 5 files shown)


chambm:
Finally, you can sim far into the future over and over without having to close those darn KAC alarms and retirement notifications every time. :)

Source/RP0/Maintenance/MaintenanceHandler.cs
chambm: Weird. I did not modify this line so this "null on execution path" is pre-existing. Should I fix with `Instance?`?

Clayell: No harm in doing so (performance is negligible iirc)
However, if there is an occasion where `SpaceCenterManagement.Instance` is actually null, more serious problems will likely emerge...

Also I've noticed that SonarCloud often makes comments about pieces of code that are unchanged, so this is to be expected.

Source/RP0/ModIntegrations/AlarmHelper.cs
```diff
@@ -76,6 +76,49 @@ public static string CreateAlarm(string title, string description, double UT, KA
...
+            //
+            // Gate on AssemblyExists (KAC installed), NOT UseKAC/APIReady: this runs during the sim's
+            // flight-scene load, and KAC clears APIReady across scene transitions, so UseKAC would be
+            // false here and we'd fall through to the (empty) stock branch. The KAC alarm list is
+            // static and stays mutable through the transition.
+            //
+            // We set Enabled=false rather than the AlarmAction: KAC's firing loop ignores the legacy
+            // AlarmAction field and gates triggering on Enabled (KerbalAlarmClock.cs), so disabling
+            // is what actually stops the warp kill / message / sound.
+            if (KACWrapper.AssemblyExists && KACWrapper.KAC != null)
```
Clayell, line 101: ~~This really should have used `if (UseKac)`, missed it~~

Nevermind just read your comment. Would probably be better to have a static `KACExists` bool

PR #2873, opened 2026-07-10
Rework the horizontal landing CC parameter

siimav:
Now with a configurable time window (5s) for evaluating ground track vs descended altitude over. 
Will now also allow smol bounces to not reset the parameter. If the plane has been airborne for over 5s (configurable as well) however the parameter will flip to unmet again.


siimav:
I guess it is time for some live testing.

PR #2874, opened 2026-07-11
Add VerticalAcceleration contract parameter

chambm:
Adds a `VerticalAcceleration` Contract Configurator parameter (CC_RP0) that tests the active vessel's **felt (proper) vertical acceleration** -- what a G-meter reads along local up:

`vertA = dot(vessel.perturbation, vessel.upAxis)` (perturbation = acceleration - gravity)

Level flight reads ~+9.81 m/s^2 (lift), free-fall reads ~0, so a band around zero (e.g. `[-1, 1]`) detects reduced-gravity / parabolic flight -- something the plain scalar acceleration magnitude can't do, since a level turn reads ~1 g just like free-fall.

It's a plain **instantaneous** check: met while `vertA` is in `[minVerticalAccel, maxVerticalAccel]` (signed, + up), optionally while surface speed is within `[minSpeed, maxSpeed]`. To require the condition be *held* for a time (with a countdown), pair it with a stock `Duration` parameter as a later sibling and set `disableOnStateChange = false`.

Config: `minVerticalAccel` and/or `maxVerticalAccel` (m/s^2, at least one), `minSpeed`/`maxSpeed` (m/s, opt), `updateFrequency` (def 0.5), `debug` (def false). Follows the existing parameter/factory pattern.

Intended for the Advanced X-Planes microgravity flight contract.

### DIFF
--- a/Source/CC_RP0/CC_RP0.csproj
+++ b/Source/CC_RP0/CC_RP0.csproj
@@ -106,6 +106,8 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Parameter\RP1ReturnHome.cs" />
     <Compile Include="ParameterFactory\RP1ReturnHomeFactory.cs" />
+    <Compile Include="Parameter\VerticalAccelerationVesselParam.cs" />
+    <Compile Include="ParameterFactory\VerticalAccelerationFactory.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\RP0\RP0.csproj">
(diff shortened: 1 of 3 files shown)


siimav:
Is there really a use for this? Flying a Vomit Comet™️shouldn't be a contract IMO.

chambm:
Don't knock it till you try it! It was even more interesting than making a VTOL craft IMO. The difficulty was not in the making of the craft but in the controlling of it. It's tough to think of other contracts to use it where the existing direction-ambivalent acceleration parameter wouldn't work. Maybe some kind of extreme pilot training where you have to make jet that can sustain high positive and negative vertical Gs.

PR #2877, opened 2026-07-11
Configure hard drives for soviet stations, lander and zond return capsule.

AntonKuzin:


siimav:
There's something to be said about Soviet electronics being 10 years behind the rest of the world but I guess it's fine for gameplay.

PR #2879, opened 2026-07-12
Adds FFT Extra engine/Utilities

gabo463:
this pr aims to adds extra very endgame engine and  utilities to rp1 
more information bellow 

linked pr :https://github.com/KSP-RO/RealismOverhaul/pull/3175 and https://github.com/KSP-RO/ROKerbalism/pull/201


WithCrow:
See my comment on #2880, same applies here

gabo463:
Pretty big update on that one :

Added ECM's
Adjusted the timeline of unlock of engine following the engine adjustments 
increased the price of some engine  due to their performance
set adequate prices on engine ( yes they should be that expensive )
antimatter integration price scales with antimatter (👀)
 
image


 thanks to @WithCrow for overall help

gabo463:
Added the solid core fission engine and moved the mirofission antimatter engine up  because having an engine that shoots what is antimatter bomb should not be anywhere else than the end node

gabo463:
added lithium tanks

PR #2885, opened 2026-07-18
Add Space Telescopes Program

WithCrow:
Continues #2699 - I've fixed the part validation bugs (thanks pmp), as well as adjusted the description inconsistency that Pap pointed out. As it appears that @wyllowthewisp has gone inactive, I've opened this PR, but if they want me to close it or continue working on their version, I will close this one.

I currently do not have the time to playtest this extensively in a natural playthrough, so any help there would be very appreciated!


WithCrow:
Now that EOS2 has been merged, I've enabled the First Space Telescope / OAO contract and added it to EOS2 - so far keeping to the original vision. I have to say that I am not entirely sure how I feel about that - while it makes sense & adds another fun dynamic to EOS2, it also means that a player that wants to do space telescopes has to have at least done that contract in the program.

PR #2891, opened 2026-07-21
Activate rest of the KerbalFoundries wheels/tracks

gabo463:
this reactivation include:
Screw Drive
RBI Inverting Track
Long Track
RBI Mole Track
Medium Track 
Simple Track 
Small Track 
Surface Track
Tiny Track 
Single Wheel Truck 
Dual Wheel Truck


WithCrow:
Do they work?

gabo463:
i don't see why not when looking at the files other than some kraken shenanigans 

and previous testing from my part before i made this showed that they are usable  (on my end)

GameData/RP-1/Tree/TREE-Parts.cfg
```diff
@@ -2964,120 +2964,120 @@
 {
     %TechRequired = lunarRatedHeatshields
     %cost = 100
-    %entryCost = 4000
-    RP0conf = false
+    %entryCost = 0
```
WithCrow: Why did you remove the entry cost for all of them?

gabo463: I did !!!  I must have changed it by mistake I ment to remove the entry cost for some of them not all

WithCrow:
Hm, I could have sworn there was some problem with these, but after flipping them all to configured in my install, they do appear to work, yeah. 

I would advise changing the PR title, you aren't adding anything. You're just flipping RP0conf to true

gabo463:
fixed the Entry cost mistake and changed to the recommended title change

PR #2892, opened 2026-07-21
Add an High-Altitude Scientific Research Balloon from the Kerballoons mod 

gabo463:
this pr adds the universal balloon form the  kerballoons mod and so contains a :

High-Altitude Scientific Research Balloon (Universal)

Capture d’écran 2026-07-21 175817

image from:https://youtu.be/0Oj4jqqJQOA?si=Nh_Pg8Eo0rrs1liJ


gabo463:
moved the balloon from 1950 to 1960 to "prevent wild shenanigans from happening"

gabo463:
the presence of procedural airship makes this redundant

WithCrow:
Not at all, no. Multiple mods that cover similar areas & across different tech levels is not a bad thing

gabo463:
i could keep it  but its realy does not ft rp1  in my opinion

WithCrow:
Why not? Several proposals to use balloons in Venus' atmosphere exist

gabo463:
this is a great point i myself used these proposal i just find it really old and some feature in the mod that are intended for stock cant be changed

gabo463:
I may revisit it later but the functionality of other mods ( proc airship makes ) cast a shadow over it

WithCrow:
That is no argument. Why are you intentionally removing diversity? With the same argument, there would be no need to have configs for more than one Shuttle mod, different engines, etc. - you are comparing appples to oranages here. These are configs for a mod, not must haves.

gabo463:
Yes I agree with this  the main problem that I did not state is kerballoon is really really old on many point the mod is made for stock and has all sort of old stock configs that are either hard to modify or can’t be changed altogether and siimav was also right to warn that this mod is not MM patchable 

also the mod does not permit clicking off the vessel for other missions and is inconsistent in stationary flight  wich can lead to undesirable results but that’s just a grudge from my part

PR #2896, opened 2026-07-23
Fix grouped tooling list collapsing distinct entries

chambm:
Follow-up bugfix to #2840.

`GetMergedEntries` located the destination node with `GetEntryIndex`, which
compares via `EpsilonCompare` (the 4% tooling margin). That is correct for
`GetToolingLevel` -- a slightly smaller tooling still covers a part -- but wrong
for building a display list: an entry landing within 4% of an existing one was
absorbed into it, inherited the survivor's value, and stopped appearing as its
own row.

Reported on a career save with several Avionics-N toolings. Actual DB contents
vs what the Toolings tab showed:

| In save | Displayed |
|---|---|
| 15.000 t x 1.25 m x 0.500 m (N2) | 15.077 t |
| 80.000 t x 3.0 m x 0.122 m (N3) | 83.000 t |
| 85.000 t x 3.0 m x 0.125 m (N3) | absent |
| 90.000 t x 3.0 m x 0.109 m (N4) | 93.000 t |

Merging now matches on exact value (`ExactCompare`, relative 1e-5 to absorb
float round-trip noise). `EpsilonCompare` is untouched, so tooling coverage
checks and `UnlockTooling` behave exactly as before. Entries that genuinely
share a value still merge into one row, which is what the tank material
grouping relies on. `GetEntryIndex` now takes the comparator so the coverage
and display searches can't drift apart.

### DIFF
--- a/Source/RP0/Tooling/ToolingDatabase.cs
+++ b/Source/RP0/Tooling/ToolingDatabase.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 
 namespace RP0
 {
@@ -24,6 +25,10 @@ namespace RP0
         protected const float comparisonLow = 1.00f - toolingMargin;
         protected const float epsilon = 1e-6f;
 
+        // Band within which two entries being merged for display count as the same value. This is
+        // float round-trip noise only -- unlike toolingMargin it carries no physical meaning.
+        protected const float mergeEpsilon = 1e-5f;
+
         public static float GetLowComparison(float value) => value * (comparisonLow - epsilon);
 
         public static float GetHighComparison(float value) => value * (comparisonHigh + epsilon);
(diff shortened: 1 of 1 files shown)


Source/RP0/Tooling/ToolingDatabase.cs
```diff
@@ -24,6 +25,10 @@ public class ToolingDatabase
         protected const float comparisonLow = 1.00f - toolingMargin;
         protected const float epsilon = 1e-6f;
 
+        // Band within which two entries being merged for display count as the same value. This is
+        // float round-trip noise only -- unlike toolingMargin it carries no physical meaning.
+        protected const float mergeEpsilon = 1e-5f;
+
```
siimav, line 31: What's the point of having two only marginally different epsilons now?

chambm: Clanker logic. Consolidated now.

PR #2897, opened 2026-07-24
Clarify some RP1CollectScience parameters

periodically-makes-puns:
Hopefully this alleviates some of the confusion about the Infrared Radiometry contract completion conditions in EEOS. Also changes a Nimbus1 parameter in EOS2 to use the same wording.

Also, ULSE's First Lunar Orbiter and Mapper contract had an incorrect percentage in the parameter title. (Unless it was intended to be 60% originally.)


siimav:
I would assume @WithCrow knows how it was intended to be.

WithCrow:
They're right on the money, since it is just a rewrite of the parameter to be more verbose about what you actually need to do

PR #2901, opened 2026-07-29
Apply leader effects for integration + rollout time in integration info

periodically-makes-puns:
When LC efficiency is not maximum, the displays in the Integration Info panel do not consider the LeaderEffect, This PR makes the net effect of leaders on integration rate visible in the Integration Info panel, and modifies all the values in the panel to account for all leader effects (including conditional ones!) properly.
image

### DIFF
--- a/Source/RP0/UI/KCT/GUI_Editor.cs
+++ b/Source/RP0/UI/KCT/GUI_Editor.cs
@@ -23,6 +23,10 @@ namespace RP0
         private static readonly GUIContent _gcSwitchToHangar = new GUIContent("Switch to Hangar", "A Launch Complex is currently selected; this will switch to the Hangar, needed for planes.");
         private static List<string> vesselFailedChecks = new List<string>();
 
+        // Rollout Rate effects should not change when this window is open.
+        private static double _cachedRolloutRateLeader = double.NegativeInfinity;
+        public static double cachedRolloutRateLeader => _cachedRolloutRateLeader.IsFinite() ? _cachedRolloutRateLeader : (_cachedRolloutRateLeader = CurrencyUtils.Rate(TransactionReasonsRP0.RateRollout));
+
         public static void DrawEditorGUI(int windowID)
         {
             if (EditorLogic.fetch == null)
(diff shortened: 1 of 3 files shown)


Source/RP0/UI/KCT/GUI_Editor.cs
```diff
@@ -73,18 +73,22 @@ private static void RenderBuildMode()
                 GUILayout.Label($"EC: {SpaceCenterManagement.Instance.EditorVessel.effectiveCost:N0}");
             }
 
+            double mult = SpaceCenterManagement.Instance.ActiveSC.ActiveLC.StrategyRateMultiplier * bpLeaderEffect;
+            GUILayout.Label($"Leader Effect on Integration Rate: {LocalizationHandler.FormatRatioAsPercent(mult)}");
```
siimav: Perhaps show this line conditionally?
Also, isn't it double counting here? StrategyRateMultiplier is already baked into the "Integration at N BP/s" field rendered directly above it, while bpLeaderEffect is not.

Source/RP0/UI/KCT/GUI_Editor.cs
```diff
@@ -73,18 +73,22 @@ private static void RenderBuildMode()
...
                         ? buildPoints / (bR * bpLeaderEffect)
-                        : SpaceCenterManagement.Instance.EditorVessel.CalculateTimeLeftForBuildRate(buildPoints, bR / effic, effic, out _))
+                        : SpaceCenterManagement.Instance.EditorVessel.CalculateTimeLeftForBuildRate(buildPoints, bR / effic * bpLeaderEffect, effic, out _))
                     : 0d;
                 GUILayout.Label($"Integration Time: {(bR > 0 ? KSPUtil.PrintDateDeltaCompact(buildTime, true, false) : "infinity")} at {effic:P0}");
 
                 if (SpaceCenterManagement.EditorRolloutBP > 0)
                 {
-                    GUILayout.Label($"Rollout Time: {(bR > 0 ? KSPUtil.PrintDateDeltaCompact(SpaceCenterManagement.EditorRolloutBP / bR, true, false) : "infinity")} at {effic:P0}");
+                    double rolloutBR = bR / SpaceCenterManagement.Instance.ActiveSC.ActiveLC.StrategyRateMultiplier * CurrencyUtils.Rate(TransactionReasonsRP0.RateRollout);
```
siimav: Might not be best to run the rate query twice per frame here. Check if the others are cached and do something similar if so.

siimav (changes requested):

Source/RP0/UI/KCT/GUI_Editor.cs
periodically-makes-puns: Doh, it is double-counting. I may remove StrategyRateMultilier from `rateWithCurEngis` and multiply it into `bpLeaderEffect`, then. That should also help simplify the rollout BP rate calculations.

periodically-makes-puns:
I think that caught everything?

periodically-makes-puns:
More fixes. Also made the personnel window account for efficiency gain and leader effects.
We explicitly display the effect of leaders on integration rate if there is any.
Also, non-human-rated crafts have their BP/s capped properly in the window.

image

(with -1 hovered)
image
image

Source/RP0/UI/KCT/GUI_Personnel.cs
```diff
@@ -185,29 +185,31 @@ private static void RenderEngineersSection(bool isCostCacheInvalid)
             }
             GUILayout.EndHorizontal();
 
-            double rateFull = KCTUtilities.GetBuildRate(0, type, currentLC, currentLC.IsHumanRated, assignDelta) * stratMult;
+            bool hasVessel = currentLC.CanIntegrate && currentLC.BuildList.Count > 0;
+            VesselProject b = hasVessel ? currentLC.BuildList[0] : null;
+            double rateFull = KCTUtilities.GetBuildRate(0, type, currentLC, hasVessel ? b.humanRated : currentLC.IsHumanRated, assignDelta);
             double rate = rateFull * efficiency;
+            double leaderMult = stratMult * b.LeaderEffect;
```
siimav: This nullref needs fixing for sure.

PR #2902, opened 2026-07-29
Change Rollout Time estimate to use estimated post-integration efficiency

periodically-makes-puns:
image
image
There's a slight discrepancy due to efficiency gain not precisely lining up. It still appears to remain an overestimate of the actual rollout time.

### DIFF
--- a/Source/RP0/UI/KCT/GUI_Editor.cs
+++ b/Source/RP0/UI/KCT/GUI_Editor.cs
@@ -92,17 +92,19 @@ namespace RP0
 
             if (double.TryParse(BuildRateForDisplay, out double bR))
             {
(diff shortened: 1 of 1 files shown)


siimav:
Has conflicts now. Also didn't something like this slip into the other PR already?

periodically-makes-puns:
Fixed conflicts. The other PR just applied the rollout-rate leader effects, this one changes the displayed (and used) efficiency

siimav:
Make sure the logic in RenderEditMode() matches

periodically-makes-puns:
RenderEditMode already used similar logic (rollout efficiency after everything in the queue is done), even prior to the other PR

PR #2905, opened 2026-08-02
Remove early solids nodes

periodically-makes-puns:
Derived from https://discord.com/channels/319857228905447436/620690446540341261/1511838949029576794 and the following discussion.

This PR removes the Early, Basic, and 1956-1957 Solids nodes, moving all parts to the corresponding classical rocketry nodes (Early => PWRT, rest match by years). ‘58SR now requires ‘56-57OR.

Currently still needs work. I think I can have SCM handle the migration internally and the upgrade scripts seem to be otherwise unnecessary unless we want to refund science + time somehow, so I may attempt to avoid the KSPAssembly version bump?


periodically-makes-puns:
I can't avoid the KSPAssembly version bump without breaking backward compatibility, since anyone that had the solid nodes unlocked before this update needs the corresponding nodes unlocked after this update. I don't know of any way to detect that except via an UpgradeScript. So we're going with an UpgradeScript and an KSPAssembly version bump. We could hypothetically also deduct the science difference in the UpgradeScript but I don't know how to do that and don't really want to, either. 

After some thought I am going with a cancel + refund approach for any projects in queue. If any of the removed nodes are in queue, we remove it and all its dependents, and refund the science. This sidesteps many inconvenient issues that would arise if I tried to replace the relevant research projects. There should not be substantial effects on the queue, since the only branch that is affected is the solids branch. Instead of trying to implement this via an UpgradeScript I will instead do an SCM version bump from 9 to 10 and update the TechList.

Source/RP0/Properties/AssemblyInfo.cs
```diff
@@ -38,9 +38,9 @@
 [assembly: AssemblyInformationalVersion("@ASSEMBLYINFORMATIONALVERSION@")]
 [assembly: KSPAssembly("RP-0", @MAJOR@, @MINOR@)]
 #else
-[assembly: AssemblyFileVersion("3.0.0.0")]
-[assembly: AssemblyInformationalVersion("3.0.0.0")]
-[assembly: KSPAssembly("RP-0", 3, 0)]
+[assembly: AssemblyFileVersion("3.1.0.0")]
+[assembly: AssemblyInformationalVersion("3.1.0.0")]
+[assembly: KSPAssembly("RP-0", 3, 1)]
```
siimav: Should match next release version. If you look a few lines up then you'll see that for actual releases the version value is 4.5 already.

periodically-makes-puns:
starting: (61 science banked)
image
ending: (71 science banked)
image

Source/RP0/SpaceCenter/SpaceCenterManagement.cs
```diff
@@ -575,6 +575,37 @@ public override void OnLoad(ConfigNode node)
...
+                            {
+                                KCT_GUI.CancelTechNode(index);
+                            }
+                            if (KSPUtils.CurrentGameHasScience())
+                            {
+                                bool valBef = IsRefundingScience;
+                                IsRefundingScience = true;
+                                try
+                                {
+                                    ResearchAndDevelopment.Instance.AddScience(refund, TransactionReasons.RnDTechResearch);
```
siimav: use new CareerEventScope(CareerEventType.Ignore) to prevent double counting in spyware

siimav: Nah, just a `using` above the `AddScience` call

Source/RP0/UpgradeScripts/v4_6_EarlySolids.cs
```diff
@@ -0,0 +1,65 @@
...
+                            techNode.AddValue("id", tech);
+                            techNode.AddValue("state", RDTech.State.Available);
+                            techNode.AddValue("cost", 0); // placeholder
+                            techs[tech] = techNode;
+                        }
+                    }
+                    foreach (var part in sourceNode.GetValues("Part"))
+                    {
+                        techs[target].AddValue("Part", part);
+                    }
```
siimav, line 57: ProtoTechNode.Save writes purchased parts as lowercase part, and ConfigNode.GetValues is case-sensitive. So GetValues("Part") always returns empty.

siimav:
RFTechLevels.cfg:18 still points at the deleted node.
Additionally go over all the scenarios and csv files and remove the old tech nodes.

Source/RP0/SpaceCenter/SpaceCenterManagement.cs
```diff
@@ -575,6 +575,36 @@ public override void OnLoad(ConfigNode node)
 
                 if (LoadedSaveVersion < VERSION)
                 {
+                    if (LoadedSaveVersion < 10)
+                    {
+                        string[] techsToRemove = { "earlySolids", "basicSolids", "solids1956" };
+                        float refund = TechList.Sum(project => techsToRemove.Contains(project.techID) ? project.scienceCost : 0);
+                        TechList.RemoveAll(project => techsToRemove.Contains(project.techID));
+                        // If the patcher gave free techs, remove them from the queue
+                        TechList.RemoveAll(project => ResearchAndDevelopment.GetTechnologyState(project.techID) == RDTech.State.Available);
```
siimav, line 584: GetTechnologyState returns Available when R&D isn't loaded so that line can wipe the entire queue.
SpaceCenterManagement.OnLoad has no ordering guarantee relative to the ResearchAndDevelopment scenario module. If Instance is null at that point, every entry in TechList matches and the whole research queue is deleted, unrefunded.

Source/RP0/UpgradeScripts/v4_6_EarlySolids.cs
```diff
@@ -0,0 +1,65 @@
...
+                if (techs.TryGetValue(source, out ConfigNode sourceNode))
+                {
+                    foreach (string tech in allTargets)
+                    {
+                        if (!techs.ContainsKey(tech))
+                        {
+                            ConfigNode techNode = node.AddNode("Tech");
+                            techNode.AddValue("id", tech);
+                            techNode.AddValue("state", RDTech.State.Available);
+                            techNode.AddValue("cost", 0); // placeholder
```
siimav, line 50: Since the value is present, ProtoTechNode uses it verbatim and the granted nodes are persisted with scienceCost = 0 forever. Just omit the cost value. ProtoTechNode then backfills the real cost.

Source/RP0/SpaceCenter/SpaceCenterManagement.cs
```diff
@@ -575,6 +575,36 @@ public override void OnLoad(ConfigNode node)
...
+                        if (refund > 0)
```
siimav, line 585: RemoveAll already returns the count

Source/RP0/UpgradeScripts/v4_6_EarlySolids.cs
```diff
@@ -0,0 +1,65 @@
...
+    [UpgradeModule(LoadContext.SFS, sfsNodeUrl = "GAME/SCENARIO")]
+    public class v4_6_EarlySolids : UpgradeScript
+    {
+        public override string Name { get => "RP-1 Early Solid Node Removal"; }
+        public override string Description { get => "Removes the Early Solids, Basic Solids, and 1956 Solids nodes from acquired tech, and unlocks the corresponding replacement nodes. "; }
+        public override Version EarliestCompatibleVersion { get => new Version(2, 0, 0); }
+        protected static Version _targetVersion = new Version(4, 6, 0);
+        public override Version TargetVersion => _targetVersion;
+
+        // Unlock all dependencies as well. This is a lot of extra tech for free :/
```
siimav, line 16: Yeah, that's definitely too much.
Removing 7 science worth of nodes (1 + 2 + 4) grants up to ~25 back: rocketryTesting 1, earlyRocketry 3, basicRocketryRP0 4, orbitalRocketry1956 7, materialsScienceSatellite 10.

periodically-makes-puns: I could shift the solids back a node to early and basic respectively, but then they’re placed ahead of time

periodically-makes-puns:
Bah, I don't see a good way to safely delete nodes without also being horribly unbalanced if you try to migrate across a save that has solids researched but not the corresponding liquids.

Perhaps I could duplicate the relevant parts, unlink the old solid nodes & parts and mark them as deprecated. but leave the nodes + parts? Then I'd add the duplicate part into the new correct combined node. That seems better, and means we don't have to do science compensation or deal with the consequences of removing nodes. That does mean we have duplicates of some parts, though...

siimav:
> Perhaps I could duplicate the relevant parts

Plz no.

How about not unlocking full nodes but adding progress to queued up tech instead. So if you have 3 sci worth of solids nodes unlocked, you get 3 sci worth of progress into your materialsScienceSatellite. I.e +30% progression for a node that costs 10 sci.

periodically-makes-puns:
That’s fair balance-wise, but it does mean crafts that could be queued prior to the migration could end up unable to be queued afterwards.

periodically-makes-puns:
Implemented the above. The UpgradeScript is gone since it is no longer necessary. We also refund any removed research progress (from completed nodes, as well as nodes in queue that were removed).

PR #2906, opened 2026-08-03
Fix incorrect Flight Director effects

### LINKED ISSUE #2790: Incorrect description/effect for Abdul Kalam's malus
https://github.com/KSP-RO/RP-1/blob/6ab88793fb5f79c86c5a9ee103810cbe0b1d2bf2/GameData/RP-1/Strategies/Leaders/LeadersFlightDirectors.cfg#L519-L529

Kalam's effect reads `cost when building new launch complexes` but actually applies to all construction, including facility upgrades.

periodically-makes-puns:
fix:
Gagarin and Slayton actually affect training rate
Kranz only affects VAB rollout, as stated
Kalam only affects LC construction not all buildings, as stated

closes #2790


siimav:
The year is 2051 when the last leader effects will be fixed.

PR #2907, opened 2026-08-10
Update contributing.md and readme.md

Clayell:


README.md
```diff
@@ -43,6 +44,8 @@ Discord Link: https://discord.gg/V73jjNd
 - Toolbar Controller
 - Kerbalism (RO Config)
 - KSP Community Fixes
+- Kerbal Changelog
```
siimav: That isn't actually required

Clayell: Hm, it's listed as a dependency on CKAN. Perhaps that should be downgraded to a recommendation then?

PR #2911, opened 2026-08-11
Add RN N1 rocket and Block D

Explorer-GLW:


Clayell:
Not sure why there's 2 jsons with the same name, one of which not even being in a folder. Additionally, you have not run the parts browser on these json files, so they will have no effect in-game. See https://github.com/KSP-RO/RP-1/tree/master/Source/Tech%20Tree/Parts%20Browser#readme for more info on how to do that.

gabo463:
there also 
no tree-parts cfg or ecm-part.cfg

Explorer-GLW:
OK, I updated every file. I'm still wondering if I should adjust the placement and prices of the parts. Does anyone have any ideas? I tried to make the N1 slightly cheaper than the Saturn V to make up for its lower payload.

siimav:
How does the total part cost and effective cost compare to Saturn V?

Explorer-GLW:
> How does the total part cost and effective cost compare to Saturn V?

"The N1 costs roughly 55,000 Funds per launch across its four stages, with an unlock cost of 1.3 million. In comparison, the Saturn V has a total cost of roughly 64,000, a tooling cost of 590,000, and an unlock cost of 1 million.

siimav:
Hmm, this has not entry cost modifiers set up. Someone can correct me if I'm wrong but such part sets should come with a common chonker of an ECM. So pay the entry cost for one part and the rest come basically free.

Explorer-GLW:
> Hmm, this has not entry cost modifiers set up. Someone can correct me if I'm wrong but such part sets should come with a common chonker of an ECM. So pay the entry cost for one part and the rest come basically free.

Ok, I added an ECM to the rocket. The N-1 (fuel tanks and interstages) costs 1.1 million funds to unlock and the fairings cost 200,000.

GameData/RP-1/Tree/EntryCostModifiers.cfg
```diff
@@ -165,6 +165,10 @@ ENTRYCOSTMODS
 	// Universal Rocket/Proton family
 	Early-Soviet-Staged = 40000, StagedHypergolic, MonolithicInjectors
 	Proton-Family = 80000, Early-Soviet-Staged
+        
+        // N1 Rocket System
+	N1-Rocket = 1100000, NK15-SubFam, Isogrids
```
siimav: Wait, isogrids? Which stages used isogrid tanks?

Explorer-GLW: Oh yeah, thanks for reminding me that the N-1 used those spherical tanks with an outer shell design, i switched the isogrid to aluminum.

siimav:
Please go over those space tug configs once again. Looks like some have at least some entry cost while others are basically 0.
I'm unfamiliar with those RN parts. If they come with built-in engines then you could just use the engine ECM and add some cost on top of that.

Explorer-GLW:
> Please go over those space tug configs once again. Looks like some have at least some entry cost while others are basically 0. I'm unfamiliar with those RN parts. If they come with built-in engines then you could just use the engine ECM and add some cost on top of that.

I added the ECM For Block D.

siimav:
Oof, something bad happened and now almost the entire ECM-Parts.cfg‎ is marked as modified.

periodically-makes-puns:
Looks like it was replaced with the contents of ECM-Engines.cfg somehow?

siimav:
Btw, we have more configs for the various engine types used on those stages. Like RD-58 etc. Better to use those than the baseline 1960 config.
Also 10000 as entry cost is too little IMO. That barely covers the avionics tooling alone. I could perhaps like with something like 25k. Ideally the stage variants would get cost reductions from having purchased a previous one but ehhhh...

Explorer-GLW:
> Btw, we have more configs for the various engine types used on those stages. Like RD-58 etc. Better to use those than the baseline 1960 config. Also 10000 as entry cost is too little IMO. That barely covers the avionics tooling alone. I could perhaps like with something like 25k. Ideally the stage variants would get cost reductions from having purchased a previous one but ehhhh...

I changed the ECM to the RD-58 and increased the entry cost to 30000.

siimav:
No, I meant using the ECM of the actual engine used on that stage. Pretty sure not all of them are RD-58.

Explorer-GLW:
> No, I meant using the ECM of the actual engine used on that stage. Pretty sure not all of them are RD-58.

Changed it to the appropriate engine for each Block D version.

Source/Tech Tree/Parts Browser/data/RN_Soviet_Rockets.json
```diff
@@ -1053,20 +1347,20 @@
...
         "era": "08-LONGTERM",
         "ro": true,
         "orphan": false,
-        "rp0_conf": false,
+        "rp0_conf": true,
         "spacecraft": "Block D-2 Space Tug",
         "engine_config": "",
         "upgrade": false,
-        "entry_cost_mods": "",
+        "entry_cost_mods": "10000, RD58-FamSyntin",
```
siimav: I imagine this should be 17D12 instead

GameData/RP-1/Tree/EntryCostModifiers.cfg
```diff
@@ -165,6 +165,10 @@ ENTRYCOSTMODS
 	// Universal Rocket/Proton family
 	Early-Soviet-Staged = 40000, StagedHypergolic, MonolithicInjectors
 	Proton-Family = 80000, Early-Soviet-Staged
+        
+        // N1 Rocket System
```
siimav: Change the whitespace to single tab

siimav (changes requested):
I see you changed the first tug to 30k but the rest are still unchanged from 10k.

Source/Tech Tree/Parts Browser/data/RN_Soviet_Rockets.json
Explorer-GLW: The 17D12 is the one used for the Buran OMS, and the RD58-FamSyntin is for the RD-58S, which is used in the Block D-2.

GameData/RP-1/Tree/EntryCostModifiers.cfg
Explorer-GLW: Done

Explorer-GLW:
> I see you changed the first tug to 30k but the rest are still unchanged from 10k.

Updated it. The first version of the Block D still costs 30000. Each upgrade now costs 15000 and you need to purchase every previous version before you can use the last one.

siimav:
Did you test this ingame as well? I believe using `_` in ECMs is ill advised due to symbols getting replaced with `-`.

Source/Tech Tree/Parts Browser/data/RN_Soviet_Rockets.json
siimav: Oh, in that case sounds like you should use `RD-58S` instead. The `RD58-FamSyntin` looks a bit of a weird ECM which doesn't actually inherit the RD58 side of the tree.

Explorer-GLW:
> Did you test this ingame as well? I believe using `_` in ECMs is ill advised due to symbols getting replaced with `-`.

Yes, I did test it in the game, and I didn't see any symbols missing or replaced.

Source/Tech Tree/Parts Browser/data/RN_Soviet_Rockets.json
Explorer-GLW: Changed it to the RD-58S

siimav:
The entry cost modifier code will do such replacements internally. I believe in configs we have done the edits manually. Better to be consistent about it.
That's the last change and I'll merge the PR.

Explorer-GLW:
> The entry cost modifier code will do such replacements internally. I believe in configs we have done the edits manually. Better to be consistent about it. That's the last change and I'll merge the PR.

OK, I’m honestly not sure what I support to change.  Do you mean that the ECM from the N1 will automatically generate when I add it in the Part Browser? And that I shouldn’t have manually written the Part of the CFG?

siimav:
`"entry_cost_mods": "15000, RD-58M-CCN, rn_proton_blockd_3"`
The change would have to go through there. OTOH I guess it can be useful if the ECM name there matches the part name to make searches easier. Ehhh... let it be then.

PR #2919, opened 2026-08-30
Fix slot number warning in admin overlay

dlesbre:
Selecting a program you cannot take due to insufficient slots in admin prints a warning at the top of the program page: "This program requires X slots but only Y are available". In the case where Y=0, this message used to incorrectly print Y=1 due to faulty logic.

### DIFF
--- a/Source/RP0/Programs/ProgramStrategy.cs
+++ b/Source/RP0/Programs/ProgramStrategy.cs
@@ -118,16 +118,10 @@ namespace RP0.Programs
             int freeSlots = ProgramHandler.Instance.MaxProgramSlots - ProgramHandler.Instance.ActiveProgramSlots;
             if (_program.slots > freeSlots)
             {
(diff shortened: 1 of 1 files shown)


Source/RP0/Programs/ProgramStrategy.cs
```diff
@@ -118,16 +118,10 @@ public override bool CanActivate(ref string reason)
...
-                    reason = $"This program requires {_program.slots} free slots but only {freeSlots} slots are available.";
-                    return false;
-                }
-                else
-                {
-                    reason = $"This program requires {_program.slots} free slots but only 1 slot is available.";
-                    return false;
-                }
+                reason = (freeSlots != 1)
+                    ? $"This program requires {_program.slots} free slots but only {freeSlots} slots are available.";
```
Clayell: ~~Think this semicolon should not be here~~ didn't see the comment mb

PR #2921, opened 2026-09-02
Add a way to see the science to confidence rate ingame

Clayell:
image
image
image
image

Non-english translations were done by Google translate and are open to correction.

(I (via google translate) also noticed that the Russian translation seems to still be using "Slow" instead of "Normal", which should probably be corrected)

(the rate is determined [here](https://github.com/KSP-RO/RP-1/blob/master/GameData/RP-1/ProgramHandlerSettings.cfg#L4-L11))


Source/RP0/UI/ConfidenceWidget.cs
```diff
@@ -75,7 +75,8 @@ public static void CreateConfidenceWidget(GameObject confidenceWidgetObj)
         }
         private static string GetTooltipTextConf()
         {
-            return Localizer.Format("#rp0_Widgets_Confidence_Tooltip", Confidence.AllConfidenceEarned.ToString("N0"));
+            return Localizer.Format("#rp0_Widgets_Confidence_Tooltip", Confidence.AllConfidenceEarned.ToString("N0"),
+                Programs.ProgramHandler.Settings != null ? Programs.ProgramHandler.Settings.scienceToConfidence.Evaluate(System.Math.Max(0d, SpaceCenterManagement.Instance.SciPointsTotal)).ToString("N0") : "2");
```
siimav: The moment a modifier applies (2 × 1.15 = 2.3 → "2"), or the curve is retuned/interpolated, the tooltip silently rounds.

siimav (changes requested):
The displayed rate ignores currency modifiers - it won't match what the player actually gets

Source/RP0/UI/ConfidenceWidget.cs
Clayell: I think it's correct now? The entire currency modifier stuff is a huge black box to me, but I looked at Confidence.OnCurrenciesModified and I think this is right. (also rounding to 2 decimals now)

```
CurrencyUtils.Conf(TransactionReasonsRP0.ScienceTransmission, Programs.ProgramHandler.Settings != null ? Programs.ProgramHandler.Settings.scienceToConfidence.Evaluate(System.Math.Max(0d, SpaceCenterManagement.Instance.SciPointsTotal)) : 2d).ToString("N2")
```

Source/RP0/UI/ConfidenceWidget.cs
```diff
@@ -75,7 +75,8 @@ public static void CreateConfidenceWidget(GameObject confidenceWidgetObj)
         }
         private static string GetTooltipTextConf()
         {
-            return Localizer.Format("#rp0_Widgets_Confidence_Tooltip", Confidence.AllConfidenceEarned.ToString("N0"));
+            return Localizer.Format("#rp0_Widgets_Confidence_Tooltip", Confidence.AllConfidenceEarned.ToString("N0"),
+                CurrencyUtils.Conf(TransactionReasonsRP0.ScienceTransmission, Programs.ProgramHandler.Settings != null ? Programs.ProgramHandler.Settings.scienceToConfidence.Evaluate(System.Math.Max(0d, SpaceCenterManagement.Instance.SciPointsTotal)) : 2d).ToString("N2"));
```
siimav: Extract to a variable. Pretty awful to read currently.

PR #2929, opened 2026-09-17
Add Operational Lunar Habitation

WithCrow:
This successor to the Early Lunar Habitation program moves the focus from the "Tent in the Woods" style short-term/temporary stays to a more long term & refined infrastructure "Cabin in the Woods". 

This means more requirements: a larger crew size (at the moment, 6) for four consecutive expeditions (total of 24 months or two years), three longer distance crewed rover expeditions from the base and ISRU for food & water production.

The optionals are even longer distance rover expeditions, repeatable "tent in the woods" type contracts, as well as contracts for developing propellant production capability. The latter is complimentary to the main goal of general ISRU and so makes for a nice addition, since it also potentially unlocks entirely new architectures for voyages to Mars & beyond. 

This is by no means done & so this is opened as a draft: the ISRU contracts are empty, the tent in the woods contract is not defined and I need to add a healthy suite of additional long term experiments, as well as LunarHabit3 to serve as the benchmark for this program. 

I'm opening this deliberately early to open up discussion & make the development process more transparent.


gabo463:
all that sound  really great!! just two question/concerns on my end  

for rovers expedition contracts  what does it offer to the player? in my experience  when making crewed rover contract it ended up beeing boring not fun to do  especially in a mandatory contract state 

the second one concerns the agriculture contract 
i believe the production of a certain amount of food could be added to this contract 


there other discrepancies but I'm sure its due to its early dev state

WithCrow:
That is the goal for the agriculture & regolith processor contracts, yes.

As for the rovers: with bon voyage, these are trivial. As I note in the comments, the alternative are expanded capability landers, like the 2 week stay in ELH's first contract. If it turns out the rovers are unfun, I'll gear over to them

PR #2930, opened 2026-09-17
Add Reuse Development Program

WithCrow:
This program is akin to the DC-X and Grasshopper flight testing programs & is therefore small in scope. It is a series of rocket hops to a given altitude and for a given duration without staging. It is gated to after the player has researched Spaceplanes Era Material Science, since a reusable rocket is not possible before (or at least an uncrewed one). 

There are two optionals: one is a series of even higher altitude hops, the other is a suborbital hop.

If there is any easier way of techgating a program, please let me know - I gated it through a virtual contract.


WithCrow:
Has not been playtested & is therefore marked as a draft

WithCrow:
Okay, has been playtested & it works very well! I built a Delta Clipper in the 2000 start for this and a) very fun to fly and b) plenty of money to do so!

WithCrow:
It also means you finally have another reason to have a small launch complex around, allowing for some funkier potential applications like a small satellite launcher, etc.

PR #2931, opened 2026-09-18
Add FullThrust to conflicts on CKAN

Clayell:
FullThrust is now [abandoned](https://github.com/KerbalMissile/RP-1-Full-Thrust/commit/23201e0c1e83c1715424ae26b7a3669ff5c51876), yet it continues to cause problems for new users that think they should download it to 'get more playability' out of RP-1. Some parts of it might be worthy of merging into RP-1, but definitely not all of it. (and work would need to be done to get it in a workable state for RP-1)


siimav:
Funny, this would effectively mean that the mod is entirely deleted from CKAN.

Clayell:
Oh no! Anyway.

